PR #32, opened 2016-06-08
facts/link: handle targets with underscores

uggedal:
The implementation feels quite brittle as there
are loads of potential valid characters.


Fizzadar:
@uggedal thanks for the PR! Agreed that it's brittle - I'm not a huge fan of using regex on a lot of the facts. I think I'll reimplement parsing `ls` without when I get a moment.

PR #33, opened 2016-06-08
Allow relative symlinks

uggedal:
Depends on changes in #32.


Fizzadar:
This is much neater - I need to handle relative values (when passed an absolute target) in the `files.link` operation, not here!

PR #34, opened 2016-06-08
facts/files: allow sha1 for files regardless of filename

uggedal:
This allows for files like:

	/etc/systemd/system/getty@tty1.service.d/override.conf


Fizzadar:
@uggedal thanks!

PR #35, opened 2016-06-15
init: support enabled state for generated units

uggedal:
For generated units (minidlna on Debian for instance) we
need to call out to its generator via systemctl is-enabled
to determine actual enablement status.

### DIFF
--- a/pyinfra/facts/init.py
+++ b/pyinfra/facts/init.py
@@ -50,7 +50,15 @@ class SystemdEnabled(FactBase):
     Returns a dict of name -> whether enabled for systemd managed services.
     '''
 
-    command = 'systemctl -alt service list-unit-files'
+    command = '''
+        systemctl --no-legend -alt service list-unit-files | while read -r SERVICE STATUS; do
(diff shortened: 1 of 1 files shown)


pyinfra/facts/init.py
```diff
@@ -50,7 +50,15 @@ class SystemdEnabled(FactBase):
     Returns a dict of name -> whether enabled for systemd managed services.
     '''
 
-    command = 'systemctl -alt service list-unit-files'
+    command = '''
+        systemctl --no-legend -alt service list-unit-files | while read -r SERVICE STATUS; do
+            if [ "$STATUS" = generated ] &&
+                systemctl is-enabled minidlna.service >/dev/null 2>&1; then
```
Fizzadar: This should be `$SERVICE.service`.

Fizzadar:
Just left on note - thanks for the PR!

uggedal:
Doh! Fixed the bug.

uggedal:
Ping!

Fizzadar:
Thanks for fixing, @uggedal!

PR #41, opened 2016-06-29
Duplicate operation messes up operation order

jkimbo:
So I came across this bug when creating a deploy file that was supposed to duplicate a simple bash script. My assumption was that each operation would be run synchronously after each other, however having two operations that had the same args messed up the whole operation order.

e.g.:

```python
apt.update(sudo=True)

# install docker key
apt.key(
    # ...
)

# setup docker source
files.line(
    # ...
)

# update again
apt.update(sudo=True)

# install docker
apt.packages(
    packages=[
        'docker-engine',
    ],
    present=True,
    sudo=True
)
```

The above resulted in the `packages` operation being run before the necessary `key` and `file` operations. I've created a failing test to illustrate this properly.

Not sure what the correct way of fixing this would be since I'm not sure if this implicit de-duping of the same command is desirable. My personal preference would be for each operation hash to always be unique (potentially by taking into account line number) so that the above example would work as expected.

N.B. I solved this issue by changing the second operation to `apt.update(sudo=True, op='second-update')`

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -20,13 +20,13 @@ ssh.RSAKey = FakeRSAKey
 ssh.AgentRequestHandler = FakeAgentRequestHandler
 
 
-from pyinfra.api import Inventory, Config, State
+from pyinfra.api import Inventory, Config, State, operation
 from pyinfra.api.ssh import connect_all, connect
 from pyinfra.api.operation import add_op
 from pyinfra.api.operations import run_ops
 from pyinfra.api.exceptions import PyinfraError
 
-from pyinfra.modules import files, server
+from pyinfra.modules import files, server, apt
 
 from .util import create_host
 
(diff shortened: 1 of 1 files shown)


Fizzadar:
@jkimbo thanks for pointing this out! I've added the line number to the `op_hash` to ensure similar operations are identified as unique here: https://github.com/Fizzadar/pyinfra/commit/c1c3a61af6065a100748af17277c95ebaa224286

Another fix in addition to the `op` kwarg would also be naming the operations with a `set` as the first argument:

```py
apt.update(
    {'Update apt #1'}
)

...

apt.update(
    {'Update apt #2'}
)
```

jkimbo:
Thats great thanks @Fizzadar !

Great tool btw, having a lot of fun using it.

PR #55, opened 2016-09-01
Fix not returning when user not wanted

hoh:
A state removing a user will fail the second time it is launched, as the rest of the operation implicitly assumes that the user in question exists.

To reproduce, run twice the following state:
```
server.user(
    'ubuntu',
    present=False,
)
```

### DIFF
--- a/pyinfra/modules/server.py
+++ b/pyinfra/modules/server.py
@@ -153,9 +153,10 @@ def user(
     if home is None:
         home = '/home/{0}'.format(name)
 
-    # User exists but we don't want them?
-    if not present and user:
-        commands.append('userdel {0}'.format(name))
+    # User not wanted?
+    if not present:
+        if user:
+            commands.append('userdel {0}'.format(name))
         return commands
 
     # User doesn't exist but we want them?


Fizzadar:
@hoh I don't think this will fix the issue - the `users` fact will only be gathered once per deploy, so even with this change both calls will return a `userdel` command.

Fixing this is complicated - ideally a deploy should only reference one "state item" (ie user X) once. So if you were to define two operations affecting the same user/file/package/etc errors like this might occur. What's your use-case for this?

Regarding the fix I'd like to avoid collecting facts at runtime (beyond callbacks/Python code mid-deploy) as this breaks the whole "generate commands before executing deploy" [deploy process](http://pyinfra.readthedocs.io/deploy_process.html). One option might be that operations can modify facts in-place as they run, such that the user operation will populate the user fact if it outputs commands to generate a user. There's potential issues here though that might cause the facts collected to become "polluted" with dodgy data, which might break later operations...

hoh:
@Fizzadar The bug appeared by launching pyinfra a second time, not by having two calls to `server.user(..., present=False)` in the deploy. 

My goal is simply to remove the default 'ubuntu' user. I had pyinfra successfully removing the user on the first execution, but launching pyinfra later on that machine failed. 

```python
    # User exists but we don't want them?
    if not present and user:
        commands.append('userdel {0}'.format(name))
        return commands
```
If `present = False` but the user does not exist, this will not return. Pyinfra will then fail at [line 207](https://github.com/Fizzadar/pyinfra/blob/develop/pyinfra/modules/server.py#L207) as by default `ensure_home=True`.

The patch fixes the issue on my system with minimal changes, but a larger refactoring of the if-else-return structure of the function would be more reliable.

Fizzadar:
@hoh aha! Makes sense now :) Thanks for the PR!

PR #56, opened 2016-09-09
Add iptables facts and modules

hoh:
_This branch is still work in progress, but I would like early feedback to be coherent with the rest of pyinfra._

**Goal**: Access iptables configuration from pyinfra ; configure packet forwarding

Checklist:
- [x] Fact to list existing rules
- [x] Fact to check presence of a packet forwarding rule
- [x] Module to set a packet forwarding rule
- [x] Tests on the facts
- [x] Tests on the module

In the present form, the following deploy works:
```python
from pyinfra.modules import iptables

if not host.fact.iptables_forward(99, '10.10.10.65'):
    iptables.forward(99, '10.10.10.65')
```

How can I make the code from the module check for the fact ? I did not get how that is supposed to work.


pyinfra/facts/iptables.py
```diff
@@ -0,0 +1,74 @@
...
+    Sorted and parsed list of iptables rules from a specific table
+    '''
+    def command(self, name='nat'):
+        return 'iptables-save -t {0}'.format(name)
+
+    def process(self, output):
+        return list(parse_rules(output))
+
+
+class IptablesForward(Iptables):
```
Fizzadar: I don't think this fact is needed - the `Iptables` fact should provide everything that's needed - and the module should do the actual state comparison using `host.fact.iptables`.

pyinfra/modules/iptables.py
```diff
@@ -0,0 +1,41 @@
...
+
+    '''
+    to_port = to_port or dport
+    definition = {
+        'dport': dport,
+        'destination': '{}:{}'.format(to_ip, to_port),
+        'interface': interface,
+        'protocol': protocol,
+    }
+    commands = [
```
Fizzadar: In here you can access `host.fact.iptables(table='nat')`, verify the desired rule exists or doesn't and then generate any commands needed to add/remove the rule.

pyinfra/modules/iptables.py
```diff
@@ -0,0 +1,41 @@
...
+'''
+
+from __future__ import unicode_literals
+
+from pyinfra.api import operation
+
+
+@operation
+def forward(state, host, dport, to_ip, to_port=None,
+            interface='eth0', protocol='tcp'):
```
Fizzadar: Can we add a `present=True` kwarg to make it also possible to remove rules that are undesired?

pyinfra/modules/iptables.py
```diff
@@ -0,0 +1,41 @@
...
+    '''
+    to_port = to_port or dport
+    definition = {
+        'dport': dport,
+        'destination': '{}:{}'.format(to_ip, to_port),
+        'interface': interface,
+        'protocol': protocol,
+    }
+    commands = [
+        ('iptables -t nat -A PREROUTING '
```
Fizzadar: I think the chain should also be a kwarg on the function with no default (I don't think `PREROUTING` is any more or less valid than the other chains...).

pyinfra/facts/iptables.py
```diff
@@ -0,0 +1,74 @@
+# pyinfra
+# File: pyinfra/facts/iptables.py
+# Desc: facts for the Linux iptables firewall
+
+from __future__ import unicode_literals
+from collections import OrderedDict
```
Fizzadar: This unused import can go!

hoh: ... and it dit

pyinfra/facts/iptables.py
hoh: Done

pyinfra/modules/iptables.py
hoh: Yes we can, done in new commit too

pyinfra/modules/iptables.py
hoh: I changed the chain to be a keyword argument. I prefer having a default for this, but I am not an iptables specialist -- would other chains make more sense for this module ?
https://github.com/Fizzadar/pyinfra/pull/56/commits/ddeacc6958976c9200e367ad7f69747ecc59a738#diff-1227b05a0ff4931b53efc65b863dd0b2R16

hoh:
I added a test on `fact.iptables`, but did not understand how to write tests for the module.

I tried adding a file named `tests/operations/iptables.forward/present.json` with as content
```json
{
    "kwargs": {
        "to_ip": "10.10.10.10",
        "dport": 8080
    },
    "facts": {
        "iptables": {"name": "nat"}
    },
    "commands": [
        "iptables -t nat -D PREROUTING -i eth0 -p tcp -m tcp --dport 8080 -j DNAT --to-destination 10.10.10.10:8080"
    ]
}
```

I probably miss something in the format of the JSON test, and the errors I receive are not very explicit.

Fizzadar:
@hoh I think the fact structure is wrong - it'd need to be:

```json
...
"facts": {
    "iptables": {
        "nat": []
    }
}
...
```

Updates look good - I've left a couple of other notes now that this has expanded to manage any nat rules.

pyinfra/modules/iptables.py
```diff
@@ -0,0 +1,65 @@
...
+The iptables modules handles iptables rules
+'''
+
+from __future__ import unicode_literals
+
+from pyinfra.api import operation
+
+
+@operation
+def forward(state, host, dport, to_ip, to_port=None,
```
Fizzadar: Now that this basically handles any rules on the `nat` table I think we can rename this operation to `iptables.nat`.

pyinfra/modules/iptables.py
```diff
@@ -0,0 +1,65 @@
...
+    '''
+    to_port = to_port or dport
+    definition = {
+        'dport': str(dport),
+        'destination': '{}:{}'.format(to_ip, to_port),
+        'interface': interface,
+        'protocol': protocol,
+        'match': protocol,
+        'chain': chain,
+        'jump': 'DNAT',
```
Fizzadar: Can we also make this a kwarg? That should mean handling most NAT rules is possible with this operation :)

Fizzadar:
Also noticed your docs update - these should/will be autogenerated via the docstring on the operation function, so you shouldn't need to manually write any docs :)

pyinfra/modules/iptables.py
hoh: Done

pyinfra/modules/iptables.py
hoh: I have an issue with testing when I rename the operation:
`renamed:    pyinfra/modules/iptables.py -> pyinfra/modules/iptables/nat.py`

This gives me `tests/operations/iptables.nat.forward/add_rule.json`, but the testing framework does not handle two dots in the name of the directory. I therefore get the following error:

```
### pyinfra Tests
--> Testing everything...
...........................E
======================================================================
ERROR: Failure: ValueError (too many values to unpack (expected 2))
----------------------------------------------------------------------
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/Users/okso/Documents/Repos/pyinfra/venv/lib/python3.5/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/Users/okso/Documents/Repos/pyinfra/tests/test_modules.py", line 120, in <module>
    locals()[operation_name] = make_operation_tests(operation_name)
  File "/Users/okso/Documents/Repos/pyinfra/tests/test_modules.py", line 26, in make_operation_tests
    module_name, op_name = arg.split('.')
ValueError: too many values to unpack (expected 2)
```

No other module uses this depth. How would you like to solve this ?

Fizzadar: Sorry, @hoh I meant keeping the same filename but just renaming the function to `nat`, such that the operation used in deploys is `iptables.nat()`...

hoh:
Looks like I made a mistake in the tests, just fixed it: `already_absent.json` and `already_present.json` do not return any command anymore (facts have been updated accordingly).

hoh:
Function renamed, and small oversight fixed too !

Fizzadar:
@hoh thanks for the updates, all looks good now :)

PR #65, opened 2016-09-17
Fix silently including all hosts on typo

hoh:
Making a typo in the name of a group, either in an inventory file
or in an include in deploy.py, results in all hosts being silently
ignored. This is due to the inventory returning None for unknown
group names.

This fixes the issue by not accepting unknown groups anymore.

### DIFF
--- a/pyinfra/api/inventory.py
+++ b/pyinfra/api/inventory.py
@@ -112,7 +112,7 @@ class Inventory(object):
         Get groups (lists of hosts) from the inventory by name.
         '''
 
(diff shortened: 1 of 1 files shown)


Fizzadar:
@hoh this is now fixed as per enabling lower/uppercase inventory groups @ https://github.com/Fizzadar/pyinfra/commit/4ff751dc17561ac94c7d8b16938fa8649c997286 :)

hoh:
@Fizzadar great, thx 👍

PR #80, opened 2016-10-27
Add module virtualenv

hoh:
This branch adds a new module to create Python virtualenvs, and supports the usual options of `virtualenv`: python interpreter, sytem-site-packages and use of symlink or copy.

It could make sense to merge this module with the `pip` module as they are quite related, but I am not sure how we would make this merge. There is also a `python` module, but it is used for unrelated operations.


hoh:
I live too much in the future, Travis failed because Py2.7 doesn't know `yield from`...

pyinfra/modules/virtualenv.py
```diff
@@ -0,0 +1,44 @@
...
+    + python: python interpreter to use
+    + site_packages: give access to the global site-packages
+    + always_copy: always copy files rather than symlinking
+    + present: whether the virtualenv should be installed
+    '''
+
+    if present is False and host.fact.directory(path):
+        # Ensure deletion of unwanted virtualenv
+        # no 'yield from' in python 2.7
+        for cmd in files.directory(state, host, path, present=False):
```
Fizzadar: I came across this when switching to the `yield` everywhere! You can just do `yield files.directory...` as pyinfra will "unroll" the generator itself @ https://github.com/Fizzadar/pyinfra/blob/develop/pyinfra/api/util.py#L25.

pyinfra/modules/virtualenv.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import unicode_literals
+
+from pyinfra.api import operation
+from pyinfra.modules import files
+
+
+@operation
+def virtualenv(
+    state, host,
+    path, python='python3', site_packages=False, always_copy=False,
```
Fizzadar: Could we default python to `None` and when present add it to the args, otherwise defaulting to the system Python version (eg. not having to specify `python='python2'` on Ubuntu 14, for example).

pyinfra/modules/virtualenv.py
```diff
@@ -0,0 +1,44 @@
...
+            yield cmd
+
+    elif present and not host.fact.directory(path):
+        # Create missing virtualenv
+        command = '/usr/bin/virtualenv -p {}'.format(python)
+        if site_packages:
+            command += ' --system-site-packages'
+        if always_copy:
+            command += ' --always-copy'
+        command += ' ' + path
```
Fizzadar: Can these commands be changed into a list and use join (just to be consistent with the other modules); and also not explicitly list the location of the `virtualenv` bin as it may be installed elsewhere on some systems, eg:

```py
args = ['virtualenv']

if python:
    args.append('-p {0}'.format(python))
...

command = ' '.join(args)
```

pyinfra/modules/virtualenv.py
hoh: Sweet

pyinfra/modules/virtualenv.py
hoh: Sure, I'll update that.

pyinfra/modules/virtualenv.py
hoh: Yes

Fizzadar:
@hoh I'm not sure how to handle the merging of this - we could have `python.virtualenv` which makes sense. I'm not sure about putting it in the pip module because pip/virtualenv don't technically always live together (one might `easy_install <package>` inside a virtualenv).

I'm going to have a think about this, but I'm leaning towards `python.virtualenv` and also creating a `python.easy_install` (eg systems w/o pip people might want to `python.easy_install('pip')... pip.packages...`).

The `pip.packages` operation could then be modified to create the virtualenv if not present by calling the `python.virtualenv` operation (with `create_virtualenv=True, **virtualenv_kwargs` additions).

hoh:
Merging it into `python` makes sense, what about moving `pip.packages` as `python.pip` ?

We would then have
* `python.easy_install`
* `python.virtualenv`
* `python.pip`

For inspiration, here is how SaltStack handles [virtualenv](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.virtualenv_mod.html) and [pip](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.pip_state.html).

Fizzadar:
I'm keen to keep `pip` as a module because I want to add more to it such as
`pip.repo` - this is also inline with the other package managers.

The way I see it `virtualenv` and `easy_install` are like extensions of
core python whereas pip is its own thing. Not sure if that makes sense!

However, this also brings up the `python.execute` operation which wouldn't
fit. I'm thinking this should become `local.execute`, I'll have a think
about how to handle that.

Fizzadar:
@hoh I've had a think about this and I reckon it's best if this becomes part of `pip.packages`. I can't really see a use-case for creating a virtualenv without installing any packages, so I think it makes sense to handle the creation of the virtualenv when installing packages (similar to how [ansible_pip](https://docs.ansible.com/ansible/pip_module.html) works). What do you think?

pyinfra/modules/virtualenv.py
```diff
@@ -35,7 +35,9 @@ def virtualenv(
 
     elif present and not host.fact.directory(path):
         # Create missing virtualenv
-        command = '/usr/bin/virtualenv -p {}'.format(python)
+        command = '/usr/bin/virtualenv'
```
Fizzadar: Could we change this to just `virtualenv`? If Python is installed to `/usr/local` pip will install virtualenv to `/usr/local/bin/virtualenv` - I think it's best to just execute the command and have the system locate it.

pyinfra/modules/pip.py
```diff
@@ -51,7 +51,7 @@ def virtualenv(
 def packages(
     state, host,
     packages=None, present=True, latest=False,
-    requirements=None, virtualenv=None
+    requirements=None, virtualenv_root=None
```
Fizzadar: I'm not keen on renaming this arg as it'll break backwards compatibility - I've worked around the conflict in the `apt` and `yum` modules by creating a second reference, eg. `_virtualenv = virtualenv` (https://github.com/Fizzadar/pyinfra/blob/develop/pyinfra/modules/yum.py#L146-L173).

oksome: That's a bit hacky, but it would work.

hoh: That's a bit of a hack, but it would work indeed.

Fizzadar: Yeah it's a bit messy, but I think it's worth it for the user-friendlyness and no breaking changes.

pyinfra/modules/virtualenv.py
hoh: This was hidden by Github as the line had changed, new commit fixes it.

Fizzadar: Ah these outdated things are super annoying at the moment, thanks!

PR #81, opened 2016-10-27
Fix #79 deprecated --force-yes

### LINKED ISSUE #79: apt --force-yes is deprecated
At every use of the `apt` module, we get the following warning:
`W: --force-yes is deprecated, use one of the options starting with --allow instead`

In the man page, we find the following information:

> --force-yes
>       Force yes; this is a dangerous option that will cause apt to continue without prompting if it is doing something potentially harmful. It should not be used except in very special situations. Using force-yes can potentially destroy your system! Configuration Item: APT::Get::force-yes. This is deprecated and replaced by --allow-downgrades, --allow-remove-essential, --allow-change-held-packages in 1.1.

We are probably looking for `--assume-yes` as a safer and non-deprecated option:

> -y, --yes, --assume-yes
> Automatic yes to prompts; assume "yes" as answer to all prompts and run non-interactively. If an undesirable situation, such as changing a held package, trying to install a unauthenticated package or removing an essential package occurs then apt-get will abort. Configuration Item: APT::Get::Assume-Yes.

hoh:
This fixes fix issue #79

### DIFF
--- a/pyinfra/modules/apt.py
+++ b/pyinfra/modules/apt.py
@@ -23,7 +23,7 @@ APT_UPDATE_FILENAME = '/var/lib/apt/periodic/update-success-stamp'
 
 def noninteractive_apt(command):
     return ' '.join((
-        'DEBIAN_FRONTEND=noninteractive apt-get -y --force-yes',
+        'DEBIAN_FRONTEND=noninteractive apt-get -y --assume-yes',
         '-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"',
         command,
     ))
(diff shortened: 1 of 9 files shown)


Fizzadar:
@hoh from what I can tell the man page states that `-y` is the same as `--assume-yes` so this is already present. I think `--force-yes` needs to be switched out with `--allow-downgrades --allow-remove-essential --allow-change-held-packages`. I'm going to confirm/test this using the test VM's in the example deploy.

Fizzadar:
This is now fixed by adding `force` kwarg to `apt.packages`/`apt.deb`.

Issue #86, opened 2016-10-29
Add: `postgres.user`, `postgres.database`


Fizzadar:

hoh:
Early rough draft:
```python
@operation
def postgresql_user(state, host, name, superuser=False):
   # Warning: this fails if the user already exists
   if superuser:
       yield 'su postgres -c "createuser --superuser {}"'.format(name)
   else:
       yield 'su postgres -c "createuser {}"'.format(name)


@operation
def postgresql_database(state, host, name, user=None):
   user = user or 'postgres'
   yield 'su {} -c "createdb {}"'.format(user, name)
```

In order to check if a user already exists, the [recommended method on StackOverflow](http://stackoverflow.com/questions/8546759/how-to-check-if-a-postgres-user-exists) is to run the following, but I have _quote_ issues running it inside `server.shell`:
```bash
psql postgres -tAc "SELECT 1 FROM pg_roles WHERE rolname='USR_NAME'"
```

PR #91, opened 2016-11-15
Add postgresql

hoh:
This branch aims at adding facts and modules to manage PostgreSQL users and databases.
Related to issue #86 

This branch is ready for some early review, but misses some features:

Features:
- [x] Users fact
- [x] User creation
- [x] User creation with superuser privileges
- [ ] User removal
- [ ] User privileges modification
- [x] Database creation
- [ ] Database removal


hoh:
Thanks, I didn't know these options exist. It might still be an issue for users to have to remember to pass these arguments when they are the default on many distributions.

Should I also remove the `sudo` in the PostgresqlUsers fact ? How should I pass the `sudo_user` then ?

hoh:
@Fizzadar what about the use of `sudo` in the fact ?

Fizzadar:
@hoh sorry for vanishing! Any sudo/sudo_user/su arguments passed to operations will be passed onto facts within those, so you shouldn't need to make any changes for the facts :)

Fizzadar:
@hoh just looked through the changes - all looks good except dropping the sudo stuff from the facts as above :)

Fizzadar:
@hoh I've just pushed the [`mysql` module](https://github.com/Fizzadar/pyinfra/blob/develop/pyinfra/modules/mysql.py) I've been working on; are you cool if I merge this and finish it up for `v0.7`?

Fizzadar:
Merging into a new `postgresql` branch where I'll finish this :)

PR #115, opened 2017-09-03
Fix inline documentation typo

tobald:


Fizzadar:
@tobald thank you for the PR!

Issue #40, opened 2016-06-29
stdin support


jkimbo:
Just started using this project and liking it so far! Had a quick question: any recommend way of dealing with needing a sudo password for a user? Is just disabling it the way to go?

As an aside it would also be awesome if the deploy bailed out as soon as a command requires some input (like prompting for a sudo password) since at the moment it just hangs.

Fizzadar:
@jkimbo at the moment disabling password is the only way - but I'd like to support input for things like sudo sudo and any dpkg config changes/etc.

jkimbo:
👍

Fizzadar:
@jkimbo could you give me an example of this? I've done some tests and calls to sudo where a password is needed _should_ fail with "sudo: no tty present and no askpass program specified", for example the following command:

```
pyinfra -i inventories/dev.py --run server.shell "echo 'sudo su' | sudo su test" -v --limit ubuntu14*
```

I would still like to support inputting the passwords at runtime - note to self: paramiko will generate a PTY by adding `get_pty=True` @ https://github.com/Fizzadar/pyinfra/blob/develop/pyinfra/api/ssh.py#L195

jkimbo:
@Fizzadar the command that fails is the `apt.update(sudo=True)` command:

```
--> Starting operation: Apt/Update
[***] >>> sudo -H -S sh -c 'apt-get update'


^CExiting upon user request!
[***] [sudo] password for jkimbo:
    [***] Success
```

Fizzadar:
Just some notes: I've done some thinking about how to handle stdin - particularly how to do it when working on multiple hots in parallel. This is going to involve some fairly major changes in how pyinfra's API and CLI interact.

Some buffering code that doesn't hang on newline: https://github.com/Fizzadar/pyinfra/commit/58852a53c8e7a82ba65adccbd45347e55f9de5e5

Design thoughts:
+ Bottom line reserved for "[hostname awaiting input] Sudo password:" or similar
+ Up/down keys to switch to change between hosts (show how many awaiting input)

Implementation:
+ CLI mode would need to intercept Python logging output, SSH output, etc
+ Lines keept in a list, reprinted above input line as data comes in
+ Hosts would need to flag if they are "awaiting input" - which would need to be detected
  in `pyinfra.api.util:read_buffer` or similar (no newline for X seconds?)
+ Hosts need to keep track of "current" command, such that input can be pushed to them
[...]

PR #116, opened 2017-09-03
Hackish: show "[sudo] password for user:" message when using @local

tobald:
A naïve fix for letting sudo inform the user it waits for his password. Only tested with @local.

The following ticket details a more advanced take on the issue: https://github.com/Fizzadar/pyinfra/issues/40

### DIFF
--- a/pyinfra_cli/util.py
+++ b/pyinfra_cli/util.py
@@ -36,6 +36,8 @@ from pyinfra.local import exec_file
 from .exceptions import CliError
 
 WAIT_CHARS = deque(('-', '/', '|', '\\'))
+CURSOR_UP_ONE = '\x1b[1A'
+ERASE_LINE = '\x1b[2K'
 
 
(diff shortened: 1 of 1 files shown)


Fizzadar:
@tobald sorry for never getting back to you on this one! I'm still not sure how or whether pyinfra should handle input. I'd like to do it as it makes the user experience far better; but the current plan in #40 is pretty involved and would make pyinfra lot more complicated :/

tobald:
> @tobald sorry for never getting back to you on this one! I'm still not
> sure how or whether pyinfra should handle input. I'd like to do it as
> it makes the user experience far better; but the current plan in #40
> is pretty involved and would make pyinfra lot more complicated :/

No worries, I do not have a use for this feature anymore. Proper input
management is the way to go as stated in #40, and seems complicated.
Meanwhile we could detect/display something to the user, that was the
purpose of this commit.

PR #126, opened 2017-12-03
Reading local files using get_file_io function

grantstephens:
Local file put wasn't working, so I used get_file_io function in util


Fizzadar:
@grantstephens nice spot! Thanks for the PR :)

PR #127, opened 2017-12-03
add gid an uid for group and user creation

gdoumenc:


gdoumenc:
Added (don't know if I should have done or not)

PR #128, opened 2017-12-05
Enhance files.sync

gdoumenc:
Added `add_deploy_dir` option as for `files.put`
Added `exclude_dir` to allow efficient sync (try to exclude node_packages and I think you will understand)

### DIFF
--- a/pyinfra/modules/files.py
+++ b/pyinfra/modules/files.py
@@ -183,7 +183,7 @@ def replace(state, host, name, match, replace, flags=None):
 })
 def sync(
     state, host, source, destination,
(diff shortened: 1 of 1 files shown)


pyinfra/modules/files.py
```diff
@@ -217,20 +217,18 @@ def sync(
...
 
             # Should we exclude this file?
-            to_exclude = False
-            if exclude:
-                for match in exclude:
-                    if fnmatch(full_filename, match):
-                        to_exclude = True
-
-            if to_exclude:
+            if exclude and any(fnmatch(full_filename, match) for match in exclude):
```
Fizzadar: Much nicer!

PR #135, opened 2018-03-27
modules/files: fix regex description

tobald:


Fizzadar:
@tobald nice catch, thanks for the PR!

PR #145, opened 2018-07-17
support ssh_config proxyjumping and include directives

tobald:
Hi,

This patch adds a dependency on sshuserconfig which fixes some paramiko shortcomings before their inclusion upstream.

### DIFF
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -22,12 +22,10 @@ from paramiko import (
     ProxyCommand,
     RSAKey,
     SFTPClient,
-    SSHClient,
-    SSHConfig,
     SSHException,
 )
 from paramiko.agent import AgentRequestHandler
-
+from sshuserclient import SSHClient
 from pyinfra import logger
 from pyinfra.api.exceptions import PyinfraError
 from pyinfra.api.util import get_file_io, make_command, read_buffer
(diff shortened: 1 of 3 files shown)


pyinfra/api/connectors/ssh.py
```diff
@@ -112,50 +110,20 @@ def _get_private_key(state, key_filename, key_password):
...
-            ssh_config.parse(f)
-            host_config = ssh_config.lookup(hostname)
-            if 'user' in host_config:
-                cfg['username'] = host_config['user']
-
-            if 'proxycommand' in host_config:
-                cfg['sock'] = ProxyCommand(host_config['proxycommand'])
-
-            if 'identityfile' in host_config:
-                cfg['pkey'] = _get_private_key(host_config['identityfile'])
```
Fizzadar: How will `sshuserclient` work with password-encrypted SSH keys? Currently in CLI mode this will ask the user for any key password but this will be lost by making the switch.

tobald: doesn't everybody uses ssh-agent or similar nowadays?

tobald: thanks for spotting the key policy setting in sshuerclient, I will remove it.

tobald: sshuserclient updated.

tobald: I adapted the test a bit hastily, I am not sure if you intend to override paramiko default behavior of look_for_keys and allow_agent.

Fizzadar: @tobald yes that's correct. The solution would be just to have:

```
kwargs = {
    'allow_agent': False,
    'look_for_keys': False,
}
...
```

This this should correctly behave and respect any `host.data.X` and CLI arguments, while also satisfying the `ssh_config` defaults :)

setup.py
```diff
@@ -30,6 +30,7 @@
     'python-dateutil>2,<3',
     'six>1,<2',
     'setuptools',
+    'sshuserclient'
```
Fizzadar: Could we pin this please to `sshuserclient>0.3,<1`? Aside from setuptools would like to keep dependencies bound to major versions just-in-case.

Fizzadar:
@tobald thanks for the updates! I think we can ignore the hostname as it'll be picked up in `sshuserclient` and/or `api/connectors/ssh.py`.

tobald:
Yay, thanks.

PR #165, opened 2019-04-11
Bug/fix unicode print

gchazot:


Fizzadar:
Awesome, thank you @gchazot!

PR #177, opened 2019-09-13
Couple of fixes: 'su' was not working + missing arg in doc

sfermigier:


Fizzadar:
Thanks for the fix @sfermigier!

I've also added https://github.com/Fizzadar/pyinfra/issues/178 to get some tests setup for this!

PR #192, opened 2019-12-25
add additional args to yum commands

mkinney:
I think we need this so we can install docker-ce using `--nobest` argument.

### DIFF
--- a/pyinfra/modules/yum.py
+++ b/pyinfra/modules/yum.py
@@ -143,7 +143,7 @@ _update = update  # noqa: E305 (for use below where update is a kwarg)
 def packages(
     state, host, packages=None,
     present=True, latest=False, update=False, clean=False,
-    installArgs='', updateArgs='', uninstallArgs='',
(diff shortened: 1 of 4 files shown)


pyinfra/modules/yum.py
```diff
@@ -143,6 +143,7 @@ def update(state, host):
 def packages(
     state, host, packages=None,
     present=True, latest=False, update=False, clean=False,
+    installArgs='', uninstallArgs='',
```
Fizzadar: For consistency could we call these `extra_install_args` and `extra_uninstall_args`?

mkinney: I was actually impressed that you implemented so many options for all of the operations.

Yes, I am lazy and wanted to use a "catch-all".

There is a down-side risk to this approach. If a user wanted to inject some nefarious things in the string, they could.

I only added the "force" because I went looking for a flag that was not yet implemented.

PR #196, opened 2020-01-02
copy yum over to dnf

### LINKED ISSUE #195: Support for dnf
Fedora (thus Centos and RedHat) switched from yum to dnf. See https://en.wikipedia.org/wiki/DNF_(software) for more info.

How should pyinfra support dnf?

Did you want to copy (or move) all of the yum stuff to dnf?

Either way, the docs will need to reflect that yum and dnf are both supported.

mkinney:
Resolves #195


mkinney:
Are there any integration-like tests?

Let me know if there is anything else that needs to be done.

pyinfra/facts/dnf.py
```diff
@@ -0,0 +1,55 @@
+from __future__ import unicode_literals
```
Fizzadar: I don't think we need this file at all - the facts in `facts/yum.py` will provide these (facts are in a global namespace).

(side note for me: I think it'd make sense to rename `facts/yum.py` -> `facts/rpm.py` as it's more accurate (rpm facts vs yum/dnf operations).)

mkinney: Done.

Issue #208, opened 2020-01-09
fix installing already installed rpms in `yum.rpm` and `dnf.rpm`


mkinney:
How do check if a package is installed?

For instance, on Centos8, if you have this snippet:

```
        dnf.rpm(
            {'Enable EPEL'},
            'https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{  host.fact.linux_distribution.major }}.noarch.rpm',
        )
```

It will work the first time just fine, but a 2nd time it errors:

```
--> Starting operation: Enable EPEL
    [192.168.2.219] package epel-release-8-8.el8.noarch (which is newer than epel-release-8-7.el8.noarch) is already installed
    [192.168.2.219] Error
```

I can add the `ignore_errors=True` option, but that feels... wrong.

Fizzadar:
Hmm this should work - we check for the package being installed already (in `rpm_packages` fact) - but the versions are different here; so I think it's trying to install the old version over the newer `8-8` version, and rejecting it.

What's the output if you run with `-v`? It _should_ be simply running `rpm -U <filename>.rpm`, which is correct - but I believe `rpm` will reject it as there's a newer version already.

mkinney:
```
--> Beginning operation run...
--> Starting operation: Install EPEL rpm to enable EPEL repo
[192.168.2.219] >>> sudo -H sh -c 'rpm -U /tmp/b7f95f3f0daf418959443bae7b416941df7e134d.rpm'
[192.168.2.219] package epel-release-8-8.el8.noarch (which is newer than epel-release-8-7.el8.noarch) is already installed
    [192.168.2.219] Error (ignored)
```

Fizzadar:
Hmm yeah this is as expected really - the RPM it's trying to install is older than the one already installed on the system. Not sure what the best course of action here is; I see a few options:

1. If the package exists, no matter the version, don't do the install (and add a `force=False` keyword to force install of the older version)
2. Compare the installed version to the one to-be-installed, which might not always work as expected (poorly versioned packages/etc)
3. Keep as-is, which will fail if the system has a newer version but this seems silly

I'm leaning towards the first option, what do you think @mkinney?

mkinney:
#1 would require code... that I'm not really sure is necessary at this time.

I think option #4 is to add `ignore_errors=True,`. (lol)

But, what would the python code to see if a 'epel-release*' package is installed?

Fizzadar:
[...]

PR #214, opened 2020-01-10
Add examples for most operations

### LINKED ISSUE #200: Add more patterns/examples for deploys
I would like to see more patterns/examples for deploys.

1) Ansible's docs usually contain a useable snippet as a starting point. Could the pyinfra docs be expanded to include examples? For example, the apt.packages could have an example of what a typical code snippet would look like.

2) Would like to see examples on using the "when" or a conditional task, so as to make deployments quicker/idempotent.

For example, the "if CentOS" in deploys.rst could show how to also use the "when" format.

Another example would be to check for the existence of a file then use that as a "when" condition.

Yet another example, would be to run a remote shell command and use the output from that as a "when" condition in a task. (a good example of this would be check if docker apt key is installed before needing to download that key)

mkinney:
Resolves #200 

Notes: 
1) Did not provide an example for every single operation.
2) There are still some TODOs in standalone_examples that I will need help with
3) Please review the standalone_examples and provide feedback for things that are not pythonic or you think should be done differently.


standalone_examples/docker_ce.py
```diff
@@ -0,0 +1,83 @@
+from pyinfra import host
+from pyinfra.modules import apt, server, python, init
+
+# Standalone example to show how to install Docker CE using
+# https://docs.docker.com/install/linux/docker-ce/ubuntu/
```
Fizzadar: I think it'd be worth linking to `pyinfra-docker` here as well for a prebuild version of this (https://github.com/Fizzadar/pyinfra-docker).

standalone_examples/docker_ce.py
```diff
@@ -0,0 +1,83 @@
...
+        running=True,
+        enabled=True,
+    )
+
+    # TODO: is NotImplementedEror the right value?
+    # TODO: show the stdout if there is an error
+    # TODO: when you change the docker to dockerf (to make it not found),
+    #       the output is not in sequence; is that output or does it really
+    #       not run these sequentially?
+    stdout = host.fact.command('docker run hello-world')
```
Fizzadar: Any `host.fact` will be run _before_ anything else in the "fact" phase (https://pyinfra.readthedocs.io/en/latest/deploy_process.html#executing-deploys). A more suitable check could be implemented using `python.call`:

```py
from pyinfra.modules import python
...

def check_docker_works(state, host):
    status, stdout, stderr = host.run_shell_command('docker run hello-world')
    if not status or 'Hello from Docker!' not in stdout:
        raise Exception('`docker run hello-world` did not work as expected')

python.call(check_docker_works, hello='world')
```

Taken from the example here: https://pyinfra.readthedocs.io/en/latest/examples/dynamic_execution_deploy.html

(also used a plain `Exception` which is probably most appropriate here!)

standalone_examples/lxd.py
```diff
@@ -0,0 +1,14 @@
+from pyinfra import host
+from pyinfra.modules import lxd
+
+SUDO = True
+
+# Note: This is an example of running a command on remote
+# init lxd using all defaults
+host.fact.command('lxd init --auto')
```
Fizzadar: This should really be:

```py
server.shell(
    {'Run lxd auto init'},
    'lxd init --auto',
)
```

The fact version will run before the actual deploy - facts should only ever read, not change anything.

pyinfra/modules/yum.py
```diff
@@ -41,6 +50,16 @@ def repo(
...
     + gpgkey: the URL to the gpg key for this repo
+
+    Example:
+
+    .. code:: python
+
+        yum.repo(
+            {'Add the Docker CentOS repo'},
+            name='DockerCE',
+            baseurl='https://download.docker.com/linux/centos/docker-ce.repo',
```
Fizzadar: This should be a proper base URL ie `https://download.docker.com/linux/centos/7/$basearch/stable`.

standalone_examples/dnf.py
```diff
@@ -30,13 +30,18 @@
...
-        '{{  host.fact.linux_distribution.major }}.noarch.rpm',
-        ignore_errors=True,
-    )
+    packages = host.fact.rpm_packages
+    epel_installed = False
+    for p in packages.keys():
+        if p.startswith('epel-release'):
+            epel_installed = True
+
+    if not epel_installed:
```
Fizzadar: I think the real fix here is to fix the behaviour of `dnf.rpm` and `yum.rpm` to properly check, within the operation, if the RPM is installed and handle appropriately (whether upgrade or downgrade). I'll investigate this as part of #208.

standalone_examples/docker_ce.py
mkinney: Ok. Will do.

standalone_examples/docker_ce.py
mkinney: Hmm.. I tried something like that but did not work as expected.

Here's a standalone example:

```
from pyinfra.modules import apt, python, server

def some_command(state, host):
    # Run a command on the server and collect status, stderr and stdout
    status, stdout, stderr = host.run_shell_command('echo hello')
    assert status == 0  # ensure the command executed OK
    returned = stdout[0]
    print("returned:", returned)

python.call(
    {'Run some command'},
    some_command,
)
server.shell(
    {'Execute some shell'},
    'echo "back to other operations!"',
)
```

And the error:
```
$ pyinfra 192.168.2.232 --user vagrant --password vagrant mike.py
--> Loading config...
--> Loading inventory...

--> Connecting to hosts...
    [192.168.2.232] Connected

--> Preparing operations...
    Loading: mike.py
    [192.168.2.232]  Ready: mike.py

--> Proposed changes:
    Ungrouped:
    [192.168.2.232]   Operations: 2   Commands: 2

--> Starting operation: Run some command
    [192.168.2.232] Unexpected error in Python callback: TypeError("run_shell_command() missing 1 required positional argument: 'command'",)
    [192.168.2.232] Error
```

What am I missing?

standalone_examples/lxd.py
mkinney: Thanks. Will fix.

pyinfra/modules/yum.py
mkinney: Thanks. Will fix.

standalone_examples/dnf.py
mkinney: Ok. I'll leave it like this for now.

standalone_examples/docker_ce.py
mkinney: I had to pass "state" to run_shell_command... weird.

mkinney: Also, status returns a boolean so, need to update https://github.com/Fizzadar/pyinfra/blob/master/docs/examples/dynamic_execution_deploy.md

Fizzadar: Sorry @mkinney this is my bad, I've now fixed those bugs in the example.

mkinney:
Thanks.

I'll stop adding to this PR and will open another PR for future examples/doc changes.

Issue #193, opened 2019-12-25
pylint and/or black or other python tooling for code quality?


mkinney:
Which do you prefer "pylint" or "black"?

```
$ pylint pyinfra/ | wc -l
    1926
```

```
$ black --check pyinfra/
<snip>
Oh no! 💥 💔 💥
73 files would be reformatted, 5 files would be left unchanged.
```
Not sure what caused it to die...

mkinney:
Which ever you choose, I can help to get the code formatted.

mkinney:
https://www.pylint.org/ and/or https://github.com/psf/black

Fizzadar:
Currently pyinfra is [linted by `flake8`](https://github.com/Fizzadar/pyinfra/blob/master/.travis.yml#L16) as part of CI, which I'm pretty happy with - I see no major reason to change from this at this time as it keeps code consistent without dramatically changing it's style.

I'm not a fan of Black's double-quote enforcement (and refusal to enforce single quotes. Have never investigated pylint before which does look really intersting - I will have a look into pylint-ing pyinfra.

mkinney:
Wow. I didn't know that flake8 had so many plugins... including black... lols.

see https://github.com/DmytroLitvinov/awesome-flake8-extensions

I'm thinking of implementing the spellcheck plugin.

mkinney:
Going to close this issue. Can always re-open or create a new issue.

PR #217, opened 2020-01-11
Spelling

mkinney:
Related to #193 

Add spellcheck to flake8

It is kinda non-intelligent spell checking (does not strip numbers/quotes), but I think having it might better than not.

I think it caught some bugs (see the deafult fixes below).

Thoughts?


mkinney:
Hmm... python 2.7 is dead. Do we care about it anymore?

setup.py
```diff
@@ -42,6 +42,7 @@
     'flake8',
     'flake8-commas',
     'flake8-quotes',
+    'flake8-spellcheck',
```
Fizzadar: To fix py2 we could just turn this into a list and append this requirement only if py3.

Fizzadar: Actually - better than that - we can just modify this line (discovered via https://stackoverflow.com/questions/34407802/how-can-i-specify-a-requirement-only-for-python-2-ie-not-python-3-in-requireme):

```py
    'flake8-spellcheck ; python_version >= "3"',
```

Fizzadar: Realised I could just do it as such a small change; hope you don't mind! https://github.com/Fizzadar/pyinfra/pull/217/commits/fb96848867e269c36e220a98d8717f444486f71a

mkinney: No worries!

Thanks

Issue #216, opened 2020-01-11
add doc/info for developing pyinfra


mkinney:
I think there should be a section/docs on how to develop pyinfra.

Things like:
- how to enable git commit hook (ex: https://flake8.pycqa.org/en/latest/user/using-hooks.html)
- how to gen doc locally
- how to run the tests (right now it is just pytest, but I'm thinking of some sort of integration-y tests)

mkinney:
And:
- how to create a virtualenv then need to run:
`pip install -e .[dev]`

Fizzadar:
This should be covered in the contributing file (https://github.com/Fizzadar/pyinfra/blob/master/CONTRIBUTING.md).

PR #226, opened 2020-01-12
Add pre commit

mkinney:
Partly resolves: #216 

Experimenting with pre-commit python plugin.
Just used default config.

I *think* it does not use the same flake8 config as the project uses, which is a bummer. But, it make prevent multiple commits (like 'flake8 fixes').

Thoughts?


mkinney:
Grrr... if in a subdir and try to commit, the whitelist.txt from the main directory will not be used. You can experience this by going into examples and running "flake8".

CONTRIBUTING.md
```diff
@@ -14,8 +14,17 @@ git clone git@github.com:Fizzadar/pyinfra.git
 
 # Install the package in editable mode with development requirements
 pip install -e .[dev]
+
+# Configure git to use pre-commit hook
+flake8 --install-hook git
```
Fizzadar: I think this needs to be `pre-commit install`.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.4.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
```
Fizzadar: Also needs flake8 config here I think as per https://github.com/pre-commit/pre-commit/blob/master/.pre-commit-config.yaml#L14.

CONTRIBUTING.md
mkinney: Those two commands produce different hooks. I'm ok with either one. The flake8 one was much simpler.

Fizzadar: Sounds godo to me - we'll need to delete the `.pre-commit-config.yaml` file and `pre-commit` dependency in that case as it won't be required anymore.

mkinney: Do you have a preference? I'm just experimenting.

Fizzadar: I think `flake8` for now.

.pre-commit-config.yaml
mkinney: not applicable since we're just going to try with flake8 for now...

PR #230, opened 2020-01-12
Examples and doc updates

mkinney:
Even more examples

It seems as I add examples, I find bugs or have questions about the operations.

Let me know if you want a break from these.  :-)


examples/apt.py
```diff
@@ -3,10 +3,49 @@
 
 SUDO = True
 
-if host.fact.linux_name in ['Debian', 'Ubuntu']:
+# Note: Using linux_distribution fact so running from docker
+# will show valid name, otherwise could just use host.fact.linux_name
+linux_distribution = host.fact.linux_distribution
```
Fizzadar: This could all be written as:

```py
code_name = host.fact.linux_distribution['release_meta'].get('DISTRIB_CODENAME')

if host.fact.linux_name...
```

The `linux_name` fact actually uses `linux_distribution` under the hood - so this won't cause any double load.

examples/apt.py
```diff
@@ -3,10 +3,49 @@
...
+    )
+
+    apt.key(
+        {'Install VirtualBox key'},
+        'https://www.virtualbox.org/download/oracle_vbox_2016.asc',
+    )
+
+    apt.repo(
+        {'Install VirtualBox repo'},
+        'deb https://download.virtualbox.org/virtualbox/debian {{ code_name }} contrib',
```
Fizzadar: This string should use f-strings or `.format` for `code_name`, rather than a jinja2 template variable. The inline jinja2-style variables are no longer required and will be deprecated/maybe removed at some point!

examples/git.py
```diff
@@ -1,9 +1,31 @@
-from pyinfra.modules import git
+from pyinfra import host
+from pyinfra.modules import apk, apt, files, git
 
 SUDO = True
 
+linux_name = host.fact.linux_distribution.get('name', '')
```
Fizzadar: This can just be `host.fact.linux_name`.

examples/apt.py
mkinney: Thanks

examples/apt.py
mkinney: Thanks

examples/git.py
mkinney: Ok. Thanks.

mkinney:
I'm ok if you merge this one, if you are ok with changes.

pyinfra/modules/yum.py
```diff
@@ -107,10 +109,11 @@ def rpm(state, host, source, present=True):
 
     .. code:: python
 
+        linux_major = host.fact.linux_distribution['release_meta'].get('VERSION_ID')
         yum.rpm(
            {'Install EPEL rpm to enable EPEL repo'},
            'https://dl.fedoraproject.org/pub/epel/epel-release-latest-'
-           '{{  host.fact.linux_distribution.major }}.noarch.rpm',
+           '{}.noarch.rpm'.format(linux_major),
```
Fizzadar: I'm not sure why this change is needed? `host.fact.linux_distribution.major` should return the same value?

mkinney: Sorry. I misunderstood the jinja2 comment above. Reverted.

Issue #235, opened 2020-01-14
Any way to know if service started successfully or errored? from init.d()


mkinney:
Any way to see if service started successfully?

For example:
```
    init.d(
        {'Restart and enable dnsmasq'},
        'dnsmasq',
        restarted=True,
        enabled=True,
    )
```

If there is an error:
```
root@vagrant:/var/log# systemctl status dnsmasq
● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
   Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2020-01-14 22:34:17 UTC; 4min 36s ago
  Process: 684 ExecStart=/etc/init.d/dnsmasq systemd-exec (code=exited, status=2)
  Process: 656 ExecStartPre=/usr/sbin/dnsmasq --test (code=exited, status=0/SUCCESS)

Jan 14 22:34:16 vagrant systemd[1]: Starting dnsmasq - A lightweight DHCP and caching DNS server...
Jan 14 22:34:17 vagrant dnsmasq[656]: dnsmasq: syntax check OK.
Jan 14 22:34:17 vagrant dnsmasq[684]: dnsmasq: unknown interface eth0
Jan 14 22:34:17 vagrant dnsmasq[684]: unknown interface eth0
Jan 14 22:34:17 vagrant dnsmasq[684]: FAILED to start up
Jan 14 22:34:17 vagrant systemd[1]: dnsmasq.service: Control process exited, code=exited status=2
Jan 14 22:34:17 vagrant systemd[1]: dnsmasq.service: Failed with result 'exit-code'.
Jan 14 22:34:17 vagrant systemd[1]: Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
```

Fizzadar:
Unfortunately I think this is an issue with the `dnsmasq` init script - the operation will execute `/etc/init.d/dnsmasq start`; if this exits with 0 then it's assumed the service has started. If this isn't the case / it didn't start properly, then the init script itself is not behaving correctly.

What does the output of the deploy look like with `-v`?

mkinney:
With this code:

```
    init.d(
        {'Restart and enable dnsmasq'},
        'dnsmasq',
        restarted=True,
        enabled=True,
    )
```

if I intentionally break the service, running without the `-v` option I see this:

```
--> Starting operation: Restart and enable dnsmasq
[...]

PR #237, opened 2020-01-15
WIP: adding pxe_infra and other examples

mkinney:
I implemented a pxe server using pyinfra using https://linuxhint.com/pxe_boot_ubuntu_server/ as a guide.

I have opened some issues and have some TODOs in the code.

pxe_infra.py might be too big of an example... maybe? Thoughts?

Please leave this PR open until these things can be improved.


examples/pxe_infra.py
```diff
@@ -0,0 +1,161 @@
+from pyinfra import host
```
Fizzadar: I think this would be improved if we put the PXE example in it's own folder, `examples/pxe/` containing a `Vagrantfile` and this file with a README containing the instructions. Should make it much easier to give it a try w/o editing any files.

examples/pxe_infra.py
```diff
@@ -0,0 +1,161 @@
...
+        '/netboot/tftp/ubuntu1804',
+    ]
+    for dir in dirs:
+        files.directory(
+            {'Ensure the directory `{}` exists'.format(dir)},
+            dir,
+        )
+
+    # TODO: how to see if the service started ok
+    init.d(
```
Fizzadar: You could use `init.systemd` here? It should me much more reliable than using the `/etc/init.d` script.

examples/pxe_infra.py
mkinney: Good idea. Will do.

examples/pxe_infra.py
mkinney: Ok. Should I check the value of init.systemd() if it did not start, or will it complain?

mkinney:
As I add random examples to demo each of the operations, I keep finding things that I need help with. Please review and let me know if there are better ways to do some of these things. (specifically look for TODOs)

mkinney:
1. Do you want an example with a puppet master/agent just to exercise the puppet.agent() operation?

2. Do you want an example with a plex install on ubuntu? (maybe use https://www.linuxbabe.com/ubuntu/install-plex-media-server-ubuntu-18-04 as a guide) This would be interesting because it has stuff like `Let’s Encrypt`.

3. Do you want more (or fewer) examples? I'm just trying to find examples that might exercise some real-world pyinfra use cases.

4. I'm thinking of adding these examples to an integration-like script and report pass/fail. Thoughts?

mkinney:
Looks like https://github.com/MichaelAquilina/flake8-spellcheck/issues/22 fixed punctation issues in whitelist.txt.

examples/pxe_infra.py
mkinney: Going to resolve this comment as https://github.com/Fizzadar/pyinfra/issues/235 was created.

mkinney:
Thanks.

I really don't like the co-mingling of the examples/tests. Will try to improve that. The puppet test does not work, but I think that's a Puppet thing, not a pyinfra thing.

Ok if you want to merge it now as it is getting quite large.

PR #271, opened 2020-01-28
tweaks to support and bug report

### LINKED ISSUE #261: thoughts on things that would assist with support? (add `pyinfra --support`)
Thoughts on adding a "--for-support" option that would show some helpful info:
- print local os info
- print pyinfra version
- print python version
- Visit 'https://github.com/Fizzadar/pyinfra/issues' and be sure to include the info above plus a small snippet that will demo the issue. (a small deploy) with the options '-vv --debug' added.

Or perhaps when you add '--debug' it will print support info like that above?

Also, may want to consider adding "template" to the issues. (I don't know how to do it but I liked how Molecule used to to it.) I just checked and they've taken it up a notch now that RedHat took ownership of it. See https://github.com/ansible/molecule/issues/new/choose 

Here's the bug one:

```
<!--- Verify first that your issue is not already reported on GitHub -->
<!--- Do not report bugs before reproducing them with the code of the master branch! -->
<!--- Please also check https://molecule.readthedocs.io/en/latest/faq.html --->
<!--- Please use https://groups.google.com/forum/#!forum/molecule-users for usage questions -->

# Issue Type

- Bug report

# Molecule and Ansible details

```
ansible --version && molecule --version
```

Molecule installation method (one of):

- source
- pip

Ansible installation method (one of):

- source
- pip
- OS package

Detail any linters or test runners used:

# Desired Behavior

Please give some details of the feature being requested or what
should happen if providing a bug report.

# Actual Behaviour

Please give some details of what is actually happening.
Include a [minimum complete verifiable example](https://stackoverflow.com/help/mcve) with
output of running `molecule --debug`.
```

Thoughts?

mkinney:
Partially resolves #261

### DIFF
--- a/pyinfra_cli/prints.py
+++ b/pyinfra_cli/prints.py
@@ -143,12 +143,15 @@ def print_facts(facts):
 
 def print_support_info():
     click.echo('''
-    If you're having issues with pyinfra or wish to make feature requests, please
-    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues.
+    If you are having issues with pyinfra or wish to make feature requests, please
+    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
     When adding an issue, be sure to include the following:
 ''')
 
     click.echo('    System: {0}'.format(platform.system()))
+    click.echo('      Platform: {0}'.format(platform.platform()))
+    click.echo('      Release: {0}'.format(platform.uname().release))
+    click.echo('      Machine: {0}'.format(platform.uname().machine))
     click.echo('    pyinfra: v{0}'.format(__version__))
     click.echo('    Executable: {0}'.format(sys.argv[0]))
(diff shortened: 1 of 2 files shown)


pyinfra_cli/prints.py
```diff
@@ -143,12 +143,15 @@ def print_facts(facts):
 
 def print_support_info():
     click.echo('''
-    If you're having issues with pyinfra or wish to make feature requests, please
-    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues.
+    If you are having issues with pyinfra or wish to make feature requests, please
+    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
```
Fizzadar: Why the space before the full stop, out of interest?

mkinney: Didn’t want any confusion on url. 

Ok if you want to remove it.

PR #277, opened 2020-02-13
Command completion

mkinney:
Got command completion working in both bash and zsh.

Not as much completion options as I would have hoped for.

### DIFF
--- a/scripts/pyinfra-complete.sh
+++ b/scripts/pyinfra-complete.sh
@@ -1,27 +1,21 @@
-#!/usr/bin/env bash
-#
-# bash completion file for pyinfra
-#
-# This script provides completion of:
-#  - commands and their options
-#
-# To enable the completions ensure that this
-# file (pyinfra-complete.sh) is sourced. For example:
-# run something like this: (where you change the path)
-#    echo "source /full_path_to/pyinfra-complete.sh" >> ~/.bashrc
-#
-_pyinfra() {
-  local cur prev opts base commands
-  COMPREPLY=()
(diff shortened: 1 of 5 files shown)


mkinney:
Hmm... I was doing completion from a user perspective not from a developer perspective.

For instance, after enabling either, you should be able to run "pyinfra <tab>" or "pyinfra -<tab>" to see available options.

scripts/pyinfra-complete.sh
```diff
@@ -1,21 +1,27 @@
-_pyinfra_completion() {
-    local IFS=$'
-'
-    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
-                   COMP_CWORD=$COMP_CWORD \
-                   _PYINFRA_COMPLETE=complete $1 ) )
-    return 0
-}
+#!/usr/bin/env bash
```
Fizzadar: What's the reasoning with changing this script? The old/current one is dynamic so will always be up to date whereas this hardcodes the options?

Fizzadar:
Info on click's autocomplete: https://click.palletsprojects.com/en/7.x/bashcomplete/

scripts/pyinfra-complete.sh
mkinney: It didn't work for me when I tried it earlier. I'll re-test.

mkinney: Ok. It works. Not sure what I did last time.

mkinney: Will close this PR and open a new one.

PR #282, opened 2020-02-17
Remove unnecessary import and variable declaration

VilhelmPrytz:
`pyinfra/progress.py`: "os" is already imported at the top, so the "from os import" is not necessary
`pyinfra_cli/legacy.py`: the "op_bits" variable is never used

### DIFF
--- a/pyinfra/progress.py
+++ b/pyinfra/progress.py
@@ -25,7 +25,7 @@ TERMINAL_WIDTH = 0
 if IS_TTY:
     try:
         TERMINAL_WIDTH = os.get_terminal_size().columns
-    except ImportError:
+    except AttributeError:
         if not IS_WINDOWS:
             terminal_size = os.popen('stty size', 'r').read().split()
             if len(terminal_size) == 2:
--- a/pyinfra/progress.py
+++ b/pyinfra/progress.py
@@ -24,8 +24,7 @@ TERMINAL_WIDTH = 0
 
 if IS_TTY:
     try:
-        from os import get_terminal_size
-        TERMINAL_WIDTH = get_terminal_size().columns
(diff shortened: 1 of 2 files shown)


pyinfra/progress.py
```diff
@@ -24,8 +24,7 @@
 
 if IS_TTY:
     try:
-        from os import get_terminal_size
-        TERMINAL_WIDTH = get_terminal_size().columns
+        TERMINAL_WIDTH = os.get_terminal_size().columns
```
Fizzadar: The import is here because `os.get_terminal_size` doesn't exist in Python2! This should be more explicit though, as there's no explanation on that, will add a comment on that.

VilhelmPrytz: Ah, but how does it affect this? Importing `from os import get_terminal_size` raises an exception in the same way as `import os` and calling the function as `os.get_terminal_size()` would in python2. I still think the global import `import os` should be avoided, but I think it's better than not having imports at the top.

See this:

```python
Python 2.7.17 (default, Nov  7 2019, 10:07:09) 
[GCC 7.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from os import get_terminal_size
  File "<stdin>", line 1, in <module>
ImportError: cannot import name get_terminal_size
>>> import os
>>> os.get_terminal_size()
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'get_terminal_size'
```

Fizzadar: Agreed this is much nicer than the import; just needs the `ImportError` updated on L28 to `AttributeError`.

VilhelmPrytz: Oh, I forgot that! Will fix, thanks! :+1:

Issue #301, opened 2020-06-08
Import hosts in Ansible inventory


ricardbejarano:
**Is your feature request related to a problem? Please describe.**
I use Ansible _alongside_ pyinfra, and it'd be great if they supported each others' inventory formats.
Being pyinfra the "underdog" one, it'd be great if it added support for parsing Ansible inventories.

**Describe the solution you'd like**
Since Ansible has YAML-formatted inventories, it's actually pretty easy to extract a group's hosts into a Python list (pyinfra's "format").

In fact, given that pyinfra is so scriptable, I've written a little snippet to extract an Ansible group's hosts into a list in pyinfra:

```python
def flatten_list(root_list):
  items = []
  for item in root_list:
    if isinstance(item, list):
      for item in flatten_list(item):
        items.append(item)
    else:
      items.append(item)
  return items

def extract_ansible_inventory_hosts(ansible_inventory):
  hosts = []
  for group in ansible_inventory:
    if "hosts" in ansible_inventory[group]:
      for host in ansible_inventory[group]["hosts"]:
        hosts.append(host)
    if "children" in ansible_inventory[group]:
      for host in extract_ansible_inventory_hosts(ansible_inventory[group]["children"]):
        hosts.append(host)
  return hosts

def extract_ansible_inventory_group(ansible_inventory, group):
  if group in ansible_inventory:
    hosts = []
    if "hosts" in ansible_inventory[group]:
      hosts = list(ansible_inventory[group]["hosts"].keys())
    if "children" in ansible_inventory[group]:
      hosts += extract_ansible_inventory_hosts(ansible_inventory[group]["children"])
    return hosts
  return [
    extract_ansible_inventory_group(ansible_inventory[child]["children"], group)
    for child in ansible_inventory
    if "children" in ansible_inventory[child]
  ]

def extract_ansible_inventory(ansible_inventory_file_path, group):
  import yaml
  with open(ansible_inventory_file_path) as ansible_inventory_file:
    ansible_inventory = yaml.safe_load(ansible_inventory_file)
  return list(set(flatten_list(extract_ansible_inventory_group(ansible_inventory, group))))

linux = extract_ansible_inventory("ansible_inventory.yaml", "linux")
```

The above works with nested groups, too.

This can be prepended in pyinfra's inventory, but it'd be great if it came built-in as some sort of helper.
It could also be extended to support other configuration management systems' inventories.

[...]

PR #303, opened 2020-06-09
Added support for YAML/JSON-formatted Ansible inventories in @ansible.

ricardbejarano:
Caused by #301.

I've refactored the `@ansible` connector into supporting "format parser routing", to enroute the inventory file into a supporting parser (INI, JSON or YAML; for now) based on its format, as well as leaving the door open for new formats to come.

I left the INI parsing logic untouched.
**Warning: I've found the INI parser does not support hosts outside any section, you may want to take a look into that. I can create a separate issue if you will.**

I've tested all three formats with the Ansible inventories shown [here](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html) (I manually translated the YAML into JSON since there's no JSON-formatted inventories in that page), as well as one of my own large YAML inventories. **It works just as expected.**

***Note 1:** I assume (haven't checked) that the logic calling `make_names_data` does not manipulate inventory file names.*

***Note 2:** Not sure what `@memoize` does. Google says it's to optimize, but I don't know what constraints it introduces, so I didn't memoize `make_names_data`. Let me know if it's optimum to wrap it based on the routine's tasks.*


ricardbejarano:
***Note 3:** we may want to add tests for JSON/YAML-formatted inventories. I'm not familiar with this project's testing so I'd highly appreciate it if someone else did. I can provide the inventory files.*

ricardbejarano:
@Fizzadar those CI tests have been stuck for 10 hours now, am I supposed to trigger them or something? Never seen tests take these long, so I'm guessing they're not even running?

Fizzadar:
@ricardbejarano thank you for the PR!, re notes:

1. This is correct it will be passed everything after the slash in `@ansible/XXX`
2. We can drop the `@memoize` call - memoize is only really useful here to show the warning (so it only shows once). I think the memoize on loading the inventory file is invalid, I will remove it
3. The current test is very simple (https://github.com/Fizzadar/pyinfra/blob/master/tests/test_connectors/test_ansible.py), it should be just a case of adding the test inventories into the test deploy directory and calling there

And re:CI - I'm investigating this, looks like the Travis integration is broken somehow. The tests _have_ run (failed) - https://travis-ci.org/github/Fizzadar/pyinfra/pull_requests. Need to add `pyyaml` as a requirement, which I think should be an extra requirement rather than major.

If you add me as a collaborator I'm happy to setup the yaml requirement and additional test inventories.

Fizzadar:
@ricardbejarano thank you for sending those over; I've now added the extra and tests in https://github.com/Fizzadar/pyinfra/pull/303/commits/8027039959c1667cb524f0c46ce6fc75cf8a4fdf + https://github.com/Fizzadar/pyinfra/pull/303/commits/4034e2a1aa765167f760c428cf142452b1326e72, will merge this once tests pass :)

ricardbejarano:
@Fizzadar at this point I don't know what's making flake8 err out.
I'll let you fix that since I'm not familiar with it.

Fizzadar:
All fixed, thank you for the PR @ricardbejarano!

Issue #302, opened 2020-06-09
Rename `pyinfra.modules` -> `pyinfra.operations`


Fizzadar:
Operations/facts are no longer bound to each other, despite having roughly similar namespacing. Now that pyinfra is heading towards v1 this is a good time to improve the import (maintaining backwards compatibility).

```py
from pyinfra.operations import files

files.download(...)
```

Fizzadar:
https://github.com/Fizzadar/pyinfra/commit/99eb369f99f674372ffe2cf165d5de72c70948fd + https://github.com/Fizzadar/pyinfra/commit/8aed9047d0344c90d6f3e9784ff62ff6497eb83e + https://github.com/Fizzadar/pyinfra/commit/3df2a363ae936258a4e02527467050908d9c61d1

Prepped for v0.15.

Fizzadar:
Reopening until this is released, as confusing.

Fizzadar:
`0.15.dev0` is now, `0.15` to follow in a few days.

Fizzadar:
Just released `0.15`!

PR #317, opened 2020-06-16
Fix import "opreations" module in examples: oprerations->modules

paul-nameless:
When using examples, I noticed that operations modules was changed but in examples left the old one


Fizzadar:
@paul-nameless thanks for the PR! This change is actually intentional - https://github.com/Fizzadar/pyinfra/issues/302. I'm reopening that issue until released because it's confusing.

PR #324, opened 2020-06-19
Do not assume that a systemd unit is a service

### LINKED ISSUE #325: Cannot manage systemd units other than service
**Describe the bug**
pyinfra assumes that every systemd unit is a `.service`, that's not the case, we may want to manager `.timer` for example.

**To Reproduce**
```python
init.systemd(
    {'Enable rss2email timer'},
    name='rss2email.timer',
    daemon_reload=True,
    enabled=True,
    running=True,
)
```

**Expected behavior**
It should enable and start my systemd timer.

**Meta**
```
    System: Linux
      Platform: Linux-5.7.2-arch1-1-x86_64-with-debian-10.4
      Release: 5.7.2-arch1-1
      Machine: x86_64
    pyinfra: v0.14.5
    Executable: /home/code/.local/bin/pyinfra
    Python: 3.7.3 (CPython, GCC 8.3.0)
```
```
pip
```

nikaro:
For example there can be `.timer`, `.path`, `.socket`, `.target` and
maybe some others. Let the user explicitly say want does he want to
manage.

Fixes #325

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: python
 
-services:
-    - docker
-
 python:
   - '2.7'
   - '3.5'
@@ -14,7 +11,7 @@ install: 'pip install -e .[test]'
 
 script:
   - flake8
-  - pytest --cov -m"not vagrant"
+  - pytest -v --cov
 
 after_success:
   - codecov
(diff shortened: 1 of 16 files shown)


nikaro:
Let me know if there is something that need to be changed. May be i should add backward compatibility? (not sure since this is still a 0.x release)

Fizzadar:
@nikaro my apologies I had not spotted the PR and made some of these changes myself! Have now merged those together and will merge this once passing 👍 Thank you for the PR!

PR #328, opened 2020-06-20
added chroot connector

FooBarQuaxx:
A new connector for chroot execution.
The code is very much based on the existing connectors namely local and docker.


FooBarQuaxx:
Thank you for accepting it!

PR #344, opened 2020-07-01
declare connectors as setuptools entrypoints

FooBarQuaxx:
This PR changes the way builtins connectors are declared, namely changed from being static entries on the `pyinfra.api.connections`  modules to being declared as `setuptools` `entrypoints`.
This will allow third party python modules to add connectors to pyinfra outside of pytnfra code base.


FooBarQuaxx:
Thanks @Fizzadar.

PR #369, opened 2020-07-11
allow files.file and files.directory to accept PathLike objects

### LINKED ISSUE #366: files.{file,directory} should support Pathlib
`pathlib.Path` is a standard library for dealing with path. Currently files operations do not accept Pathlib. It would be great to have the support of Pathlib, as converting `pathlib.Path` via `str(...)` makes the deploy code less concise.

e.g.

```python
files.file(Path("/usr/local/cuda").joinpath("include/cudnn.h"), assume_present=True)
```

FooBarQuaxx:
This an attempt to resolve #366

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -30,6 +30,7 @@ from pyinfra.api import (
 from pyinfra.api.connectors.util import escape_unix_path
 from pyinfra.api.util import get_file_sha1, get_template
 
+from .util.compat import fspath
 from .util.files import chmod, chown, ensure_mode_int, sed_replace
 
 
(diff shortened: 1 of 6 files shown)


FooBarQuaxx:
I still have to do a minimal test for real pathlib Path objects.

FooBarQuaxx:
@Fizzadar Can you take a look at this implementation?

Fizzadar:
@FooBarQuaxx I think we tried to do the same thing at the same time here! I added support for this similarly in https://github.com/Fizzadar/pyinfra/commit/16c80d0a9dc11a6f82c4c9d13551c104ef908d07...

FooBarQuaxx:
That's what I thought too. Only thing is that using fspath is more flexible that relying on the pathlib module existence.
I can point to you to this PEP 519: https://www.python.org/dev/peps/pep-0519/

Fizzadar:
Good point, keen to enable this flexibility. I'm happy to fix the conflicts on your branch if you give me push access.

FooBarQuaxx:
@Fizzadar you can do that.

FooBarQuaxx:
@Fizzadar Let me know if you had a problem pushing to this PR's branch.

Fizzadar:
@FooBarQuaxx got access thank you! Been offline today (!) so will pick this up first thing tomorrow!

PR #371, opened 2020-07-12
reimplement linux distro info parsing using nir0s/distro

FooBarQuaxx:
[nir0s/distro](https://github.com/nir0s/distro) module does a better job at parsing Linux distribution version and information.
I've rewrote the Linux distribution fact parsing part to make use of it.

### DIFF
--- a/pyinfra/facts/util/distro.py
+++ b/pyinfra/facts/util/distro.py
@@ -1,4 +1,4 @@
-from __future__ import unicode_literals
+from __future__ import absolute_import, unicode_literals
 
 import os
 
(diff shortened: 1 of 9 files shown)


pyinfra/facts/util/distro.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import, unicode_literals
+
+import os
+
+import distro
+
+
+def get_distro_info(root_dir):
+    # We point _UNIXCONFDIR to root_dir
+    old_value = distro._UNIXCONFDIR
```
FooBarQuaxx, line 10: This part is bit hacky but it does the trick. Master version of distro has support for passing a root_dir to LinuxDistribution object __init__. Once that's release and this is accepted we can make use of it instead of resorting to this.

Fizzadar: 👍 works for now!

PR #380, opened 2020-07-15
Removing vim on Fedora Workstation doesn't work

alexandrunastase:
So there is no package called **vim**, but there are **vim-enhanced** and **vim-common**. The problem is when trying to remove the vim package this does not work.

This is the code that I am trying to run on my Fedora 32 Workstation:

```python
from pyinfra import host
from pyinfra.operations import dnf, server

dnf.packages(
    packages=["vim"],
    present=False,
    sudo=True,
)
```
Output copied from the rpm_packages fact
[rpm_packages_fact_output.txt](https://github.com/Fizzadar/pyinfra/files/4929463/rpm_packages_fact_output.txt)

### DIFF
--- a/tests/operations/dnf.packages/remove_packages.json
+++ b/tests/operations/dnf.packages/remove_packages.json
@@ -1,0 +1,20 @@
+{
+    "args": [[
+        "git",
+        "vim-enhanced",
+        "vim-common"
+    ]],
+    "kwargs": {
+        "present": false
+    },
+    "facts": {
+        "rpm_packages": {
+            "git": ["2.26.2-1.fc32"],
+            "vim-enhanced": ["8.2.1081-1.fc32"],
+            "vim-common": ["8.2.1081-1.fc32"]
+        }
+    },
+    "commands": [
+        "dnf remove -y git vim-enhanced vim-common"
+    ]
+}
--- a/tests/operations/dnf.packages/remove_packages_with_alias.json
+++ b/tests/operations/dnf.packages/remove_packages_with_alias.json
@@ -1,0 +1,19 @@
+{
+    "args": [[
+        "git",
+        "vim"
+    ]],
+    "kwargs": {
+        "present": false
+    },
+    "facts": {
+        "rpm_packages": {
+            "git": ["2.26.2-1.fc32"],
+            "vim-enhanced": ["8.2.1081-1.fc32"],
+            "vim-common": ["8.2.1081-1.fc32"]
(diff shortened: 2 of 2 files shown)


Fizzadar:
I'm not sure how to handle this! In this case `vim` is an alias for other packages - currently the `rpm_packages` fact will only include actual package names.

For now my only suggestion is to use the explicit names to remove, rather than just `vim`.

FooBarQuaxx:
I am not sure if this is an issue or a pull request. Plus as @Fizzadar said, just use the real name of the package (in your case) on the distro you're doing the provisioning on.

alexandrunastase:
@Fizzadar @FooBarQuaxx  Thanks for the quick response !

I know I could use the real name of the package, but wanted to mention this particular edge case. I tried to find out more about how this dnf/yum aliases work but came up empty handed. I thought that maybe @Fizzadar had a better idea. I used Ansible before and it somehow took the aliases into consideration as well

I created a pull request because I wanted to start with a test, thinking that it could explain the scenario better. I can create an issue and reference this PR if you guys thing that would be useful. @Fizzadar What do you think ?

Fizzadar:
@alexandrunastase agreed I think this should become an issue rather than PR!

Interested to dig into this, would be a great addition for all the package managers that have alias/similar naming.

FooBarQuaxx:
What can be done here is to register the capacities of the installed rpm (using something like this `rpm --queryformat "%{NAME} %{VERSION}-%{RELEASE}\n" -q --provides vim-enhanced`). That command output looks like this on a fedora:32 docker container.
```
vim-enhanced 8.0.1763-13.el8
/usr/bin/mergetool
/usr/bin/vim
config(vim-enhanced) = 2:8.0.1763-13.el8
vim = 8.0.1763-13.el8
vim-enhanced = 2:8.0.1763-13.el8
vim-enhanced(x86-64) = 2:8.0.1763-13.el8
```
Starting from second line downward are the capabilities (what the package providers) which vim is one of them.

alexandrunastase:
@Fizzadar I attempted to specify the real name of the packages when removing and that seems also not to work. There might be an issue there. I suspect it has something to do with the fact that the packages have the dash character. I added another test which is reproducing the behavior.

 I will create a issue for this and another one for the "alias" issue. Thanks again for taking the time, guys

Fizzadar:
Thanks for making those issues @alexandrunastase! Closing this one now :)

Issue #383, opened 2020-07-17
Add ability for Docker connector to execute on remote machine


charles-l:
**Is your feature request related to a problem? Please describe.**

The docker connector is awesome, but only seems to work for the local machine. If I want to start a docker container on a remote machine, I have to use the remote machine connector and execute a "docker run" command through the shell which doesn't leverage the functionality of the existing `@docker` connector.

**Describe the solution you'd like**

Ideally, I'd be able to specify a remote host and a container name to run on that host, something like `@docker/remote.host/container:1.1` (there's probably a better syntax than this).

charles-l:
BTW, if this seems like a feature you'd be willing to include in the project, I'd be interested in implementing it, either by expanding the current docker connector or by adding a new remote-docker connector.

Fizzadar:
This is a really interesting idea! I originally thought it'd just make sense to install `pyinfra` alongside `docker` on the remote machine - but that's less than ideal.

I think a new connector, combining the docker + SSH connectors (`@docker-ssh/remote.host:image:tag` could work?).

charles-l:
OK -- cool, I'll take a stab at implementing this then.

Fizzadar:
Thanks @charles-l! Feel free to ping me or message here if you need a hand w/anything :)

charles-l:
@Fizzadar ok -- finished up a basic draft PR -- had some questions for next steps for you when you get the chance to look it over.

PR #388, opened 2020-07-18
dockerssh: add support for running docker commands on remote machine

charles-l:
Addresses #383

Basically, I've smashed together the functionality of the docker and ssh connectors.
This is just a first attempt, and will be cleaned up, refactored and tested later on.

I wanted to throw out a quick draft PR just to show the direction I'm going. This is very WIP at the moment.


charles-l:
So, this patch currently allows the happy path of running an "echo Hello World" on a remote machine. Next I'll make sure it can get/put files.

charles-l:
OK, so this works with my few manual tests of the happy paths. I still have to clean up the code and add tests. Before going further I want to clarify a few things, @Fizzadar 

1. I saw there is support for a local LXD connector. Will there eventually be a similar need for an sshlxd connector, and if so should we try to generalize to allow for chaining connectors, or allow DI in the container connectors? e.g. the docker connector could be written in a way that it is constructed with a generic shell execute function and we can chose whether it's local.shell or remote.shell. [1] It might be premature to generalize now, so this could be a PR for later.
2. When constructing tempfiles on the remote system, I assume python is installed and call a remote tempfile.mkstemp(). This is a hack and I want to replace it, but for now it guarantees the correct behavior. Do you have any suggestions/preferences on how I should go about implementing this? I was thinking of just grabbing the source for tempfile.mkstemp and implementing it in the remote case, but there's probably a better way.
3. I had to put in a hack in the connection function to do part of the connection initialization that's normally done outside to make remote ssh calls work (https://github.com/Fizzadar/pyinfra/pull/388/files#diff-dc4e47d4814489b294c98099a91abd05R54). I only used ssh/docker connector functions in this implementation, but can/should I use the ssh operators instead?

[1]: Of course, this could add unnecessary complexity to the code. Hacking on this code base has been such a joy so far because the flatness and simplicity makes it easy to figure out what's going on. It's *really* valuable for debugging and outside contribution. This has definitely been one of the easiest Python projects for me to jump into and start adding features to, so kudos for that!

pyinfra/api/connectors/dockerssh.py
```diff
@@ -0,0 +1,235 @@
...
+def remote_remove(state, host, filename, print_output=False, print_input=False):
+    remove_status, _, remove_stderr = run_shell_command(
+        state, host, 'rm -f {0}'.format(filename),
+        print_output=print_output,
+        print_input=print_input)
+
+    if not remove_status:
+        raise IOError('\n'.join(remove_stderr))
+
+def remote_mkstemp(state, host):
```
Fizzadar, line 25: We should avoid using Python on the remote side here, as this increases the remote requirements - I think just uploading to a temp filename works (see: https://github.com/Fizzadar/pyinfra/blob/master/pyinfra/api/connectors/ssh.py#L399).

charles-l:
OK -- cool -- so I just did some minor cleanup of the code. I'll build some tests based on the ssh connector next.

charles-l:
Whoops -- I didn't realize a rebase would pull the commits into this PR like that. Looks like I've just gotta do a regular merge to catch up.

benoitjpnet:
Hello,

I would like to have a `@lxd` connector, it is planned on the roadmap? If not, should I open a feature request ticket?

Thanks!

Fizzadar:
Hi @benpro an lxd connector is a great idea! Yes please add a new ticket to track this or I can do later today :)

PR #396, opened 2020-07-26
Removes line.strip from read_buffer

sysadmin75:
This resolves the issue where command fact strips leading and trailing whitespace.


leahneukirchen:
I would use `.rstrip` here, even if it makes no practical difference.

Fizzadar:
I'm keen to get this merged / support this but need to ensure it doesn't break anything. Absolutely agree this is far superior to just stripping all whitespace from the output 👍!

Issue #295, opened 2020-05-22
Use checksum arguments to validate existing files in `files.download`


Fizzadar:
If the file already exists at the destination and any of the checksum arguments are provided, check the file matches and re-download if not.

sysadmin75:
@Fizzadar  I was looking at this issue. I see there are a couple routes that can be taken.
One route would be to modify the host.fact.file that is already called in the download operation, and add a sha256sum (also md5sum and sha1sum) function to that fact.
Or the other route would be to add these sha and md5 functions to the download operation.
I just wanted to get your thoughts on what the best way you think would be.

Fizzadar:
@sysadmin75 thanks for looking into this! So I put a placeholder in the operation a while back (https://github.com/Fizzadar/pyinfra/blob/master/pyinfra/operations/files.py#L84). I was thinking of just using the existing `sha1` fact (example @ https://github.com/Fizzadar/pyinfra/blob/master/pyinfra/operations/files.py#L579) to check and, if they don't match, set `download = True`.

Then the same for the other checksums - these will also need facts implemented - currently only `sha1_file` is a fact (https://github.com/Fizzadar/pyinfra/blob/master/pyinfra/facts/files.py#L36).

Let me know if that does/doesn't make sense!

sysadmin75:
>  I was thinking of just using the existing sha1 fact
Do you mean you would actually use the `host.fact.sha1_file` or do you mean you want to implement a host fact for any checksum type that is given (e.g. sha256sum, md5sum, etc)?

Fizzadar:
Both! `host.fact.sha1_file` can be used to check against the `sha1sum` argument. For the `sha256sum` and `ms5sum` arguments will need to create new `host.fact.md5_file` and `host.fact.sha256_file` facts.

sysadmin75:
@Fizzadar 
I noticed you use a regex to get the hash from the sha1sum command. Can you go into detail why you chose to handle it this way?

Fizzadar:
@sysadmin75 simplest option to match the hash within the line - splitting or similar would also work but the regexes covered the different output formats (`sha1` vs `shasum`) nicely. But certainly not concerned about changing these things if there's a superior parsing method!

PR #405, opened 2020-08-02
Issue #295 - validate existing files using checksum in files.download

sysadmin75:
Added functionality for sha256sum and md5sum for files.download.

### DIFF
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -157,8 +157,8 @@ class Sha256File(FactBase):
     def command(self, name):
         name = escape_unix_path(name)
         self.name = name
-        return "sha256sum {0} 2> /dev/null || shasum -a 256 {0} 2> /dev/null \
-                || sha256 {0}".format(name)
+        return 'sha256sum {0} 2> /dev/null || shasum -a 256 {0} 2> /dev/null \
+                || sha256 {0}'.format(name)
 
     def process(self, output):
         for regex in self._regexes:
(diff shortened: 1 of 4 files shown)


sysadmin75:
I'm not sure how to resolve the test failure.  I'm open to suggestions.

sysadmin75:
I was able to resolve the test locally by reverting this commit:
https://github.com/Fizzadar/pyinfra/pull/405/commits/f028fddf7363947ba1fa37f62acbfb9ce4a7a23b
However this will cause the flake8 failure.

sysadmin75:
This change allows both the flake8 and pytest to run.  However this seems like an added step that shouldn't be needed.
```
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -157,8 +157,8 @@ class Sha256File(FactBase):
     def command(self, name):
         name = escape_unix_path(name)
         self.name = name
-        return 'sha256sum {0} 2> /dev/null || shasum -a 256 {0} 2> /dev/null \
-                || sha256 {0}'.format(name)
+        ret_str = 'sha256sum {0} 2> /dev/null || shasum -a 256 {0} 2> /dev/null || sha256 {0}'
+        return ret_str.format(name)
 
     def process(self, output):
         for regex in self._regexes:
```

pyinfra/facts/files.py
```diff
@@ -138,7 +138,57 @@ def process(self, output):
...
+    _regexes = [
+        r'^([a-zA-Z0-9]{64})\s+%s$',
+        r'^SHA256\s+\(%s\)\s+=\s+([a-zA-Z0-9]{64})$',
+    ]
+
+    def command(self, name):
+        name = escape_unix_path(name)
+        self.name = name
+        return 'sha256sum {0} 2> /dev/null || shasum -a 256 {0} 2> /dev/null \
+                || sha256 {0}'.format(name)
```
Fizzadar, line 161: This should cover both flake8 + test:

```suggestion
        return (
            'sha256sum {0} 2> /dev/null '
            '|| shasum -a 256 {0} 2> /dev/null '
            '|| sha256 {0}'
         ).format(name)
```

PR #416, opened 2020-08-21
WIP: Kubernetes support

leahneukirchen:
This is a WIP to create a Kubernetes connector and helpers for dynamic inventory.

The Kubernetes connector is modeled after the Docker connector and uses `kubectl exec` (and `kubectl cp`). Pods can be specified as `@kubernetes/namespace/podname`.

The dynamic inventory provides a method `get_pods` that can select pods by Kubernetes labels.

There are no docs or tests yet, first trying to see if this interests you. Else, I would also maintain this as an external project, but adding third-party connectors doesn't seem to be supported.

### DIFF
--- a/pyinfra/api/connectors/kubernetes.py
+++ b/pyinfra/api/connectors/kubernetes.py
@@ -1,0 +1,225 @@
+import os
+
+from tempfile import mkstemp
+
+import click
+import six
+
+from pyinfra import logger
+from pyinfra.api import QuoteString, StringCommand
+from pyinfra.api.exceptions import InventoryError
+from pyinfra.api.util import get_file_io, memoize
+
+from .local import run_shell_command as run_local_shell_command
+from .util import make_unix_command, run_local_process, split_combined_output
+
+
+@memoize
+def show_warning():
+    logger.warning('The @kubernetes connector is in beta!')
+
+
+def make_names_data(pod=None):
+    if not pod:
+        raise InventoryError('No pod provided!')
+
+    namespace = 'default'
+    if '/' in pod:
+        namespace, pod = pod.split('/', 2)
+
+    show_warning()
+
+    # Save the namespace and pod name as the hostname, @kubernetes group
+    yield '@kubernetes/{0}/{1}'.format(namespace, pod), \
+        {'namespace': namespace, 'pod': pod}, ['@kubernetes']
+
+
+def connect(state, host, for_fact=None):
+    return True
+
+
+def disconnect(state, host):
+    return True
+
+
+def run_shell_command(
+    state, host, command,
+    get_pty=False,
+    timeout=None,
+    stdin=None,
+    success_exit_codes=None,
+    print_output=False,
+    print_input=False,
+    return_combined_output=False,
+    **command_kwargs
+):
+    # Don't sudo/su, see docker connector.
+    for key in ('sudo', 'su_user'):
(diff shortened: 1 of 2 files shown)


pyinfra/api/inventories/kubernetes.py
```diff
@@ -0,0 +1,26 @@
...
+
+def get_pods(selector, namespace='default', all_namespaces=False, container=None):
+    if all_namespaces:
+        ns = ['-A']
+    else:
+        ns = ['-n', namespace]
+
+    return_code, combined_output = run_local_process(
+        ['"$@"', '-',
+         'kubectl', 'get', 'pods', *ns, '-l', selector, '--template',
```
Fizzadar, line 13: Unfortunately Python 2 doesn't support the `*ns` expansion here - you'll need to create a list of arguments, append the namespace ones and then the rest before passing to the `run_local_process` function.

pyinfra/api/inventories/kubernetes.py
```diff
@@ -0,0 +1,26 @@
+from pyinfra.api.connectors.util import run_local_process, split_combined_output
```
Fizzadar, line 1: I think this file/function should live within the `pyinfra.api.connectors` module. Could either be inside `api/connectors/kubernetes.py` or make `api/connectors/kubernetes/__init__.py` + `api/connectors/kubernetes/util.py`.

Fizzadar (commented):
@leahneukirchen apologies for taking so long to get back to you on this - keen for this to be merged as an additional connector - it's never occurred to me the idea of targeting a live K8s pod.

Looks great so far, left a couple of comments!

---

Sidenote: it _should_ be possible to use setuptools entrypoints such that external packages can add connectors without being part of pyinfra itself.

leahneukirchen:
Not sure why this is failing now, seems unrelated to my PR.

Fizzadar:
This appears to be a linting issue now;

```sh
./pyinfra/api/connectors/kubernetes.py:64:32: C812 missing trailing comma
./pyinfra/api/connectors/kubernetes.py:69:42: C812 missing trailing comma
./pyinfra/api/connectors/kubernetes.py:75:34: C812 missing trailing comma
./pyinfra/api/connectors/kubernetes.py:216:19: C812 missing trailing comma
./pyinfra/api/connectors/kubernetes.py:219:63: Q000 Remove bad quotes
```

The verbose pytest output has basically buried it way up in the log - I've pushed https://github.com/Fizzadar/pyinfra/commit/f45ea1a1d6871969b359fe29f6f3b55adbc60cde to stop this happening in future!

leahneukirchen:
Argh, I ran a local copy of flake8, not the one in the venv. Thanks!

robsonpeixoto:
Great! A suggestion is allow to pass the `--context` as a parameter. It's very useful when you are managing multiples clusters

Fizzadar:
This is still a really cool idea but should definitely be managed as an external project. v2 & upcoming v3 both support this via setuptools entrypoints, I am going to provide an example project as part of the upcoming v3 release cycle at which point I shall link it here and close this PR.

leahneukirchen:
I'm not using k8s currently, but still think it could be useful to somebody.

Fizzadar:
If I have time one day (or anyone wants to), we could make `pyinfra-dev/pyinfra-kubernetes-connector` because it's a really cool use of connectors. But for now closing.

Issue #417, opened 2020-08-27
use_sudo_password fails when the password ends with "


kvold:
My sudo password ends with a quotation mark: "
When I try to use "use_sudo_password" on a Centos-server it fails with the following error message:
```
[server] bash: -c: line 0: unexpected EOF while looking for matching `"'
[server] bash: -c: line 1: syntax error: unexpected end of file
[server] Error
```
I assume it's the " which causes the error since the code I'm running is relatively simple (it fails on the yum.packages command):
```
from pyinfra.operations import yum, server

server.shell(
    name='test',
    commands='echo "hello world"',
)

yum.packages(
    name='Ensure the following packages are installed',
    packages=['vim', 'git'],
    sudo=True,
    use_sudo_password=True
)
```
I'm running pyinfra from the terminal on a Fedora 32 machine, it connects to a Centos 7 machine through ssh. 

**Meta**
+ Include output of `pyinfra --support`.
```
(pyinfra) [kloerl]$ pyinfra --support
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Linux
      Platform: Linux-5.7.15-200.fc32.x86_64-x86_64-with-glibc2.2.5
      Release: 5.7.15-200.fc32.x86_64
      Machine: x86_64
    pyinfra: v1.1
    Executable: /home/kloerl/.venv-python/pyinfra/bin/pyinfra
    Python: 3.8.5 (CPython, GCC 10.2.1 20200723 (Red Hat 10.2.1-1))
```

+ Consider including output with `-vv` and `--debug`.
```
(pyinfra) [pyinfra]$ pyinfra -v inventory.py deploy.py --debug -vv
    [pyinfra_cli.main] Checking potential directory: 
    [pyinfra_cli.main] Checking potential directory: 
    [pyinfra_cli.main] Checking potential directory: 
    [pyinfra_cli.main] Checking potential directory: 
--> Loading config...
--> Loading inventory...
    [pyinfra_cli.inventory] Creating fake inventory...

--> Connecting to hosts...
    [pyinfra.api.connectors.ssh] Connecting to: nnrite ({'allow_agent': True, 'look_for_keys': True, 'timeout': 10})
    [nnrite] Connected
    [pyinfra.api.state] Activating host: nnrite

--> Preparing operations...
    Loading: deploy.py
[...]

PR #418, opened 2020-08-28
Fixed issue with sudo password having special characters.

sysadmin75:
Fixes issue with sudo password having special characters.
Issue #417


Fizzadar (approved):
Thanks for the PR @sysadmin75!

PR #419, opened 2020-08-31
Adjust systemd unit name regex to documentation

i-do-cpp:
`systemd.init` fails for units containing dots in their name, e.g. `abc.def.service`, even though they match the systemd spec.
See https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Description for reference:
```
Valid unit names consist of a "name prefix" and a dot and a suffix specifying the unit type. The "unit prefix" must consist of one or more valid characters (ASCII letters, digits, ":", "-", "_", ".", and "\"). The total length of the unit name including the suffix must not exceed 256 characters. The type suffix must be one of ".service", ".socket", ".device", ".mount", ".automount", ".swap", ".target", ".path", ".timer", ".slice", or ".scope".

Units names can be parameterized by a single argument called the "instance name". The unit is then constructed based on a "template file" which serves as the definition of multiple services or other units. A template unit must have a single "@" at the end of the name (right before the type suffix). The name of the full unit is formed by inserting the instance name between "@" and the unit type suffix.
```

This PR makes pyinfra use a regex allowing all unit names described in the manual (and more, but we don't want to overcomplicate the regex).

### DIFF
--- a/pyinfra/facts/init.py
+++ b/pyinfra/facts/init.py
@@ -3,6 +3,21 @@ import re
 from pyinfra.api import FactBase
 
 
+# Valid unit names consist of a "name prefix" and a dot and a suffix specifying the unit type.
+# The "unit prefix" must consist of one or more valid characters
+# (ASCII letters, digits, ":", "-", "_", ".", and "\").
+# The total length of the unit name including the suffix must not exceed 256 characters.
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thank you for the PR @i-do-cpp!

PR #420, opened 2020-08-31
Consider waiting units (timers) to be running

i-do-cpp:
Pyinfra explicitly checks for the `running` sub-state of a systemd unit.
E.g. for timers the sub-state equivalent is `waiting`.

This PR fixes PyInfra always starting timers (even though they are already waiting) if you specify `running=True` in the operation `init.systemd`.


Fizzadar (approved):
Makes sense, thanks for the PR @i-do-cpp!

PR #424, opened 2020-09-01
Fix a typo in state host arg warning.

weakish:


Fizzadar (approved):
Nice spot, thanks for the PR @weakish!

PR #434, opened 2020-09-12
docs: Fix simple typo, envrionment -> environment

timgates42:
There is a small typo in pyinfra/api/connectors/local.py, pyinfra/api/connectors/ssh.py, pyinfra/api/connectors/winrm.py.

Should read `environment` rather than `envrionment`.


Fizzadar (approved):
Nice spot, thanks for the PR @timgates42!

PR #444, opened 2020-10-04
Add typing information for operation decorator

iwoloschin:
This is a quick PR to simply add typing information for the `@operation` decorator. With this typing information added it allows pylance/pyright (and presumably mypy & pyre, though I haven't tested those) to properly check decorated operations. It does *not* check for global arguments (such as `name`), those require additional typing definitions to be added, that would be handled in a subsequent PR.

I have not done any testing beyond on my own system (python3.8 on macOS) so consider this a bit of an exploratory PR. Happy to add/change things as needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -119,7 +119,6 @@ if __name__ == '__main__':
             'ansible': ANSIBLE_REQUIRES,
         },
         include_package_data=True,
-        package_data={'pyinfra': ['py.typed', 'api/*.pyi']},
         classifiers=[
             'Development Status :: 5 - Production/Stable',
             'Environment :: Console',
--- a/setup.py
+++ b/setup.py
@@ -119,6 +119,7 @@ if __name__ == '__main__':
             'ansible': ANSIBLE_REQUIRES,
         },
         include_package_data=True,
+        package_data={'pyinfra': ['py.typed', 'api/*.pyi']},
         classifiers=[
             'Development Status :: 5 - Production/Stable',
             'Environment :: Console',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -119,6 +119,7 @@
             'ansible': ANSIBLE_REQUIRES,
         },
         include_package_data=True,
+        package_data={'pyinfra': ['py.typed', 'api/*.pyi']},
```
Fizzadar, line 122: I believe this should be in `MANIFEST.in` rather than here, they'll override the manifest file currently!

Fizzadar (commented):
@iwoloschin thanks for the PR! This looks great, awesome to see type annotations coming into the project! One change w/packaging stuff.

I'm going to create a branch specifically to build out type annotations & setup mypy requirement/checking as part of CI/etc, and will change the PR base branch to that.

setup.py
iwoloschin: Ah, I've been using Poetry myself lately, so I wasn't really sure how to handle that. Easy enough to change.

blarghmatey:
One thing that might be helpful in retroactively adding type information is https://google.github.io/pytype/. It will scan the existing code and infer types based on static analysis which might speed up some of the initial work.

Fizzadar (approved):
Finally got the base branch + testing setup here! Thank  you for the PR @iwoloschin :)

iwoloschin:
Awesome! I'll see if I can start adding some more typing things in. I'll probably focus on things I'm using myself first, but if there are any specific things that you think would be helpful to focus on let me know.

PR #457, opened 2020-10-20
add pip3

trigremm:
pip3 is a copy of pip with pip='pip3' on line 80


trigremm:
due to existing fix : 
Add support for pip3 or custom pip tool #100

PR #465, opened 2020-10-27
Add git.bare_repo operation

stchris:
First of all: thanks for keeping `pyinfra` so easy to contribute to. It's very pleasant to work with as well, of course! 👏 

I was looking to do `git init --bare` with it and here's my take at it. 

Cheers!


Fizzadar (approved):
Hi @stchris! Thank you for the PR and kind words :)

This looks great!

PR #467, opened 2020-10-29
Add git.worktree operation

remybar:
Thanks for your great tool!

For our server deployment, we need to create several working tree from several GIT repository. For the moment, we are using a bash script but it would be nicer to use a dedicated operation :-)

I have added a git.worktree operation to add working trees either from a new branch, an existing one or for a throwaway working tree which is just a detached HEAD (based on the GIT documentation : https://git-scm.com/docs/git-worktree).

I was thinking about adding new facts to know if a folder is a valid GIT repository / a valid working tree, based on the fact that:
- a valid GIT repository contains a `.git` **folder** and the following diff returns no diff: `diff <(git rev-parse --git-dir) <(git rev-parse --git-common-dir)`
- a valid worktree contains a `.git` **file** and the following diff returns a diff : `diff <(git rev-parse --git-dir) <(git rev-parse --git-common-dir)`
But I'm not sure it would be really useful as GIT commands already output clear information in case of errors.


Fizzadar (approved):
@remybar thank you for the PR, this looks great! I've never seen or heard of git worktress before; spend a while reading up about them :)

PR #475, opened 2020-11-03
Describe the case when we append a line to a file in the files.line operation

remybar:
As I discovered this feature by change while implementing a files.append operation, I've added a short comment about that secret feature ;-)


Fizzadar (approved):
Thank you for the PR @remybar!

PR #485, opened 2020-11-19
Add systemd operation a `--user` operation mode

jprltsnz:
This adds a new switch to `pyinfra.operations.systemd.service` which would allow it to run on user mode `systemctl --user` which will allow to do user specific unit activation.

I am testing pyinfra to setup my personal machine on a "semi-reproducable" way and hit a limitation with my systemd user units.

### DIFF
--- a/pyinfra/operations/systemd.py
+++ b/pyinfra/operations/systemd.py
@@ -25,7 +25,7 @@ def service(
     service,
     running=True, restarted=False, reloaded=False,
     command=None, enabled=None, daemon_reload=False,
-    state=None, host=None,
+    user_mode=False, state=None, host=None,
 ):
     '''
     Manage the state of systemd managed units.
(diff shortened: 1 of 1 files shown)


pyinfra/operations/systemd.py
```diff
@@ -25,7 +25,7 @@ def service(
     service,
     running=True, restarted=False, reloaded=False,
     command=None, enabled=None, daemon_reload=False,
-    state=None, host=None,
+    state=None, host=None, user_mode=False,
```
Fizzadar, line 28: The `user_mode` argument here must live before `state` & `host`.

Sidenote: the documentation on this is _not_ explicit, I've added https://github.com/Fizzadar/pyinfra/issues/486 to track fixing that.

Fizzadar (changes requested):
Hi @jprltsnz, thank you for the PR! One little thing to fix then this is good to go :)

pyinfra/operations/systemd.py
jprltsnz: Done!

jprltsnz:
@Fizzadar should be good to go now.

Fizzadar (approved):
Thank you @jprltsnz!

PR #487, opened 2020-11-19
Set lower=False on `pacman.packages` operation

jprltsnz:
`pacman` capitals on package names, and they are significant see for example `hunspell-en_US` https://www.archlinux.org/packages/extra/any/hunspell-en_US/. Currently that packages fails to install since it will search for `hunspell-en_us` instead.


Fizzadar (approved):
Nice catch, @jprltsnz! Thank you for the PR :)

Issue #488, opened 2020-11-20
Documentation for git.bare_repo operation is missing


stchris:
**Describe the bug**
I provided a PR for creating bare git repos in #465, which was merged and I can see the code in master, but it's not in the release notes and the operation doesn't show up in the docs. I believe I missed to add a changelog entry as well as something to bring in the documentation.

**To Reproduce**

* go to https://docs.pyinfra.com/en/1.x/operations/git.html
* instructions for the operation `git.bare_repo` should be there, but aren't

**Expected behavior**
Documentation should be present.

**Meta**

Fizzadar:
Hi @stchris! The major branch docs only update on actual release so it matches `pip install...`; the master docs are under latest and showing, ie: https://docs.pyinfra.com/en/latest/operations/git.html#git-bare-repo.

On my todo list is to (re)add a version switcher to make this more obvious.

stchris:
Ah I see, cheers. I did seem to break the formatting there unfortunately. Maybe this bug should track that instead?

Fizzadar:
This is now released in v1.3!

PR #489, opened 2020-11-20
Fix documentation for git.bare_repo op

stchris:
Fixes related to #465, as discussed in #488


Fizzadar (approved):
Thank you for fixing this @stchris!

PR #490, opened 2020-11-20
Fix link to contributing page

stchris:
(I'm assuming you want the latest version of the docs here).


Fizzadar (approved):
Missed this one! The `/page/X` are normally redirects to 1.x but for contributing I think latest actually makes more sense 👍. Thank you!

PR #500, opened 2020-12-22
rework of git.worktree + new git.tracking_branch fact

remybar:
I have splitted this PR in 2 commits:
1) add a new `git.tracking_branch` fact to get the actual tracking branch configured on the repo/worktree.
2) reworking of `git.worktree` as the previous interface was not fully compatible with the `git worktree add`
operation. Now, it is and if the worktree exists, a `git pull` is done.

(See commit messages for more details).

### DIFF
--- a/pyinfra/facts/git.py
+++ b/pyinfra/facts/git.py
@@ -41,14 +41,14 @@ class GitTrackingBranch(FactBase):
 
     @staticmethod
     def command(repo):
-        return r'cd {0} && git status -sb'.format(repo)
+        return r'cd {0} && git status --branch --porcelain'.format(repo)
 
     @staticmethod
     def process(output):
         if not output:
             return None
 
-        m = re.search('\\.{3}(\\S+)\\b', output[0])
+        m = re.search(r'\.{3}(\S+)\b', output[0])
         if m:
             return m.group(1)
         return None
--- a/pyinfra/facts/git.py
+++ b/pyinfra/facts/git.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import re
+
 from pyinfra.api.facts import FactBase
 
 
@@ -32,3 +34,21 @@ class GitConfig(FactBase):
             items[key] = value
 
         return items
+
+
+class GitTrackingBranch(FactBase):
+    requires_command = 'git'
+
+    @staticmethod
+    def command(repo):
+        return r'cd {0} && git status -sb'.format(repo)
+
+    @staticmethod
+    def process(output):
+        if not output:
+            return None
+
+        m = re.search('\\.{3}(\\S+)\\b', output[0])
+        if m:
+            return m.group(1)
+        return None
--- a/pyinfra/operations/git.py
+++ b/pyinfra/operations/git.py
@@ -6,6 +6,7 @@ from __future__ import unicode_literals
 
 import re
 
+from pyinfra import logger
 from pyinfra.api import operation, OperationError
 
 from . import files, ssh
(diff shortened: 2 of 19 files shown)


pyinfra/facts/git.py
```diff
@@ -32,3 +34,21 @@ def process(output):
...
 
         return items
+
+
+class GitTrackingBranch(FactBase):
+    requires_command = 'git'
+
+    @staticmethod
+    def command(repo):
+        return r'cd {0} && git status -sb'.format(repo)
```
KangOl, line 44: When parsing git output, it's always better to use the porcelain format to have a well-formatted output.
Even if the Porcelain Format Version 2 is easier to parse, it may be too recent (git 2.11.4, Sept 2017). So we keep using version 1.
Also, short options are great when you are in your shell, but in code, always use the long version. It's more descriptive.

```suggestion
        return 'cd {0} && git status --branch --porcelain'.format(repo)
```

pyinfra/facts/git.py
```diff
@@ -32,3 +34,21 @@ def process(output):
...
+
+    @staticmethod
+    def process(output):
+        if not output:
+            return None
+
+        m = re.search('\\.{3}(\\S+)\\b', output[0])
```
KangOl, line 51:
```suggestion
        m = re.search(r'\.{3}(\S+)\b', output[0])
```

pyinfra/operations/git.py
```diff
@@ -245,6 +298,27 @@ def worktree(
...
+
+            if rebase:
+                command += ' --rebase'
+
+            if from_remote_branch:
+                if '/' not in from_remote_branch:
+                    raise OperationError(
+                        'The remote branch must follow the following format REMOTE/BRANCH_NAME '
+                        '(for example: origin/master)',
+                    )
```
KangOl, line 317: It's fragile. Branches can contain `/`. You may even have a local branch named "origin/master".

To avoid ambiguities, I would change the API and force the `from_remote_branch` argument to be a 2-tuple `(remote, branch)`

KangOl (changes requested):

pyinfra/operations/git.py
```diff
@@ -149,18 +149,23 @@ def repo(
...
     '''
     Manage git worktrees.
 
-    + repo: git main repository directory
     + worktree: git working tree directory
-    + branch: branch to use for the working tree
-    + create_branch: the branch already exist or should be created
+    + repo: git main repository directory
     + detached: create a working tree with a detached HEAD
+    + new_branch: local branch name created at the same time than the worktree
```
Fizzadar, line 163: To avoid breaking changes here we need to keep the old arguments valid (deprecated is fine) - I think `branch` can be left as is (not rename to `new_branch`).

For `create_branch` - is it possible to create a worktree using an existing branch? Ie calling with `branch=x, create_branch=False`)?

Fizzadar (changes requested):
@remybar thank you for the PR; this is a great improvement 👍! Left one comment on the arguments and also agree with @KangOl's suggestions (thank you for the review!).

pyinfra/operations/git.py
remybar: > To avoid breaking changes here we need to keep the old arguments valid (deprecated is fine) - I think `branch` can be left as is (not rename to `new_branch`).

Yes, true! To avoid confusion and to be sure to handle properly old arguments usage vs new arguments usage, I prefer keeping the old arguments (`branch` and `create_branch) and the new ones (`new_branch` and `commitish`). I will handle the conversion/compatibility check at the beginning of the operation.

My first API version was due to a bad understanding of the full `git worktree add` API (I used a small part of this API). Now, I think that the new API is more coherent with the `git worktree add` API :-)

> For `create_branch` - is it possible to create a worktree using an existing branch? Ie calling with `branch=x, create_branch=False`)?

Yes, it's possible. In this case, we have to use the `commititsh` argument which represents a git element from where to create the worktree (a commit, a branch, etc ...). So, for example:
```
        git.worktree(
            name='Create a worktree from the existing local branch `v1.0`',
            repo='/usr/local/src/pyinfra/master',
            worktree='/usr/local/src/pyinfra/hotfix',
            commitish='v1.0'
        )
```

Fizzadar (approved):
Thanks for updating this @remybar, looks good to me!

Edit: failing test is unrelated (fixed in master), so merging!

PR #502, opened 2020-12-26
Fix docs link to Connectors in README.md

mrshu:
Fix docs link to Connectors (used `readthedocs.io` before, now points to `docs.pyinfra.com`)


Fizzadar (approved):
Thank you for spotting & fixing this @mrshu!

PR #503, opened 2021-01-03
Add user_mode to systemd_reload operation

smdnv:
Commit 9b09852b8564eed33a2e1e2c296c533882a5e667 added `user_mode` option for `systemd.service` operation, but it's doesn't work for `systemd daemon-reload`.

I'm trying to use `systemd.service` for manage user unit. If I change unit config fiile (`*.service`) and run the next code, systemd daemon won't be reloaded with `--user` flag, and service will be run with the old unit config.
Another problem with this operation is it's asking for a password when executing `systemctl daemon-reload`.
```python
op.systemd.service(
    service="user_service",
    user_mode=True,
    enabled=True,
    restarted=True,
    running=True,
    daemon_reload=True,
)
```

This is a warning message that I got if I change the unit config file and run `systemd status` command.
```
systemctl --user status user_service.service
Warning: The unit file, source configuration file or drop-ins of user_service.service changed on disk. 
Run 'systemctl --user daemon-reload' to reload units.
```

I added optional `user_mode` parameter into `systemd.daemon_reload` operation. This allows to reload systemd with `--user` flag.

### DIFF
--- a/pyinfra/operations/systemd.py
+++ b/pyinfra/operations/systemd.py
@@ -10,12 +10,15 @@ from .util.service import handle_service_control
 
 
 @operation
-def daemon_reload(state=None, host=None):
+def daemon_reload(state=None, host=None, user_mode=False):
     '''
     Reload the systemd daemon to read unit file changes.
+
+    + user_mode: whether to use per-user systemd (systemctl --user) or not
     '''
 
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thank you for the PR @smdnv, this looks good to me!

PR #505, opened 2021-01-03
Unpack pacman package groups into a list of packages.

unai-ndz:
Arch linux [package groups](https://wiki.archlinux.org/index.php/Meta_package_and_package_group) are not real packages and because of that are not displayed by 'pacman -Q' as installed. This causes pyinfra to reinstall them every time.

To fix that this PR expands groups into the corresponding packages before calling `ensure_packages`.

### DIFF
--- a/tests/operations/pacman.packages/add_packages.json
+++ b/tests/operations/pacman.packages/add_packages.json
@@ -1,9 +1,16 @@
 {
-    "args": ["curl"],
+    "args": [["curl", "xorg-fonts"]],
     "facts": {
+        "pacman_unpack_groups": {
+            "curl": "curl",
+            "xorg-fonts": [
+                "xorg-font-util",
+                "xorg-fonts-encodings"
+            ]
+        },
         "pacman_packages": {}
     },
     "commands": [
-        "pacman --noconfirm -S curl"
+        "pacman --noconfirm -S curl xorg-font-util xorg-fonts-encodings"
     ]
 }
--- a/tests/operations/pacman.packages/add_packages.json
+++ b/tests/operations/pacman.packages/add_packages.json
@@ -2,8 +2,8 @@
     "args": [["curl", "xorg-fonts"]],
     "facts": {
         "pacman_unpack_groups": {
-            "curl": "curl",
-            "xorg-fonts": [
+            "['curl', 'xorg-fonts']": [
+                "curl",
                 "xorg-font-util",
                 "xorg-fonts-encodings"
             ]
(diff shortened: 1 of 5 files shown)


unai-ndz:
After taking a look I can't figure out why the tests are failing. Any pointers?

Fizzadar (commented):
Hi @unai-ndz, thank you for this PR! Looks good so far!

For the tests you need to update the specific JSON tests for the `pacman.packages` operation here: https://github.com/Fizzadar/pyinfra/tree/master/tests/operations/pacman.packages. In each JSON file there's a dictionary of fake fact data which now needs the new `pacman_unpack_groups` fact data.

unai-ndz:
This is better, but I'm unable to fix an issue with the tests.
They fail when supplying a list of arguments. The list gets wrapped around a tuple and the tuple iterated, using the list inside as arguments.

tests/operations/pacman.packages/add_packages.json
```diff
@@ -1,9 +1,16 @@
 {
-    "args": ["curl"],
+    "args": [["curl", "xorg-fonts"]],
     "facts": {
+        "pacman_unpack_groups": {
+            "curl": "curl",
+            "xorg-fonts": [
+                "xorg-font-util",
+                "xorg-fonts-encodings"
+            ]
```
Fizzadar, line 9:
```suggestion
            "['curl', 'xorg-fonts']": [
                "curl",
                "xorg-font-util",
                "xorg-fonts-encodings"
            ]
```

tests/operations/pacman.packages/remove_packages.json
```diff
@@ -4,6 +4,10 @@
         "present": false
     },
     "facts": {
+        "pacman_unpack_groups": {
+            "curl": "curl",
+            "i-dont-exist": ""
```
Fizzadar, line 9:
```suggestion
            "['curl', 'i-dont-exist']": ["curl"]
```

Fizzadar (commented):
Thank you for looking at this @unai-ndz - turns out the test utils can't handle any kind of complex arguments; I've now pushed https://github.com/Fizzadar/pyinfra/commit/57aac6ca7585f1c1aabe5e7398b507060d8f3221 to address this, and these two changes will fix the tests now, then good to merge!

Fizzadar (approved):
Thank you very much for the PR @unai-ndz! Realised GitHub lets me add those changes so will now merge once tests pass :)

PR #508, opened 2021-01-06
doc fix for deploy example

makom:
Changed permissions to mode for files.file example.


Fizzadar (approved):
Nice catch, thank you @makom!

PR #513, opened 2021-01-08
Document ssh_hostname option

### LINKED ISSUE #512: Support multiple servers on one IP
**Is your feature request related to a problem? Please describe.**

I'm testing out a setup in a series of Vagrant boxes locally before doing it on some real servers, which has the slight issue that they've all got the same hostname (but different ports), so listing them as SSH servers in inventory doesn't work!

**Describe the solution you'd like**
`ssh_hostname` option allowing to specify the "connection" hostname as opposed to the "display" hostname, so I can then set hostname to something unique and nice, and `ssh_hostname` to just `127.0.0.1` every time. Default is to just use the standard hostname, so no change if you don't use it.

palfrey:
Fixes #512

### DIFF
--- a/docs/connectors.md
+++ b/docs/connectors.md
@@ -22,6 +22,7 @@ pyinfra @ssh/my-host.net ...
 **Available Data**:
 
 ```py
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Nice spot; thank you for the PR @palfrey!

PR #534, opened 2021-02-23
Fixed apk uninstall command

Gaming4LifeDE:
APK uses "apk del" instead of "apk remove" to uninstall packages. I fixed it.


Fizzadar (approved):
Thanks for spotting & fixing this @Gaming4LifeDE!

PR #535, opened 2021-02-24
server fact Users: add uid & gid

tsnoam:
the "Users" fact of "server" was missing the uid & gid.

Since it is something which I needed for my project, I went ahead and added it + adapted the unitest.


Fizzadar (approved):
Nice addition, thanks for adding & testing @tsnoam!

PR #536, opened 2021-02-24
Update README.md

tsnoam:
Properly quote shell commands


Fizzadar (approved):
Thank you @tsnoam! I pushed a little extension to remove the indent and wrap the other commands.

PR #537, opened 2021-02-25
pyinfra_cli: support reading user operations from `user_ops` package

tsnoam:
This pull request allows a user to create custom operations and simply use them directly from the CLI.

Usage is very simple:
* Create the following directory hierarchy:
```
some_path/
    my_ops.py
```
* The contents of `my_ops.py`:
```python
from pyinfra.api import operation


@operation
def hello(arg1, state=None, host=None):
    yield 'echo Hello {0}'.format(arg1)
```
* Make sure `some_path` is in PYTHONPATH (or simply, just the current working directory) and run:
```shell
pyinfra INVENTORY my_ops.hello arg1=tsnoam
```


pyinfra_cli/util.py
```diff
@@ -125,10 +125,15 @@ def get_operation_and_args(commands):
...
     op_module, op_name = operation_name.split('.')
 
+    # Try to load the requested operation from the main operations package.
+    # If that fails, try to load from the user's operations package.
     try:
         op_module = import_module('pyinfra.operations.{0}'.format(op_module))
     except ImportError:
-        raise CliError('No such module: {0}'.format(op_module))
+        try:
+            op_module = import_module('user_ops.{0}'.format(op_module))
```
Fizzadar, line 134:
```suggestion
            op_module = import_module(op_module)
```

I think removing the `user_ops` requirement makes this feature even more powerful - for example you could import operations installed via external packages/etc and execute them via the CLI.

Fizzadar (commented):
@tsnoam I absolutely love this idea, one suggestion to make it even more flexible/powerful 👍

tsnoam:
> one suggestion to make it even more flexible/powerful +1

@Fizzadar for some reason I feared it wouldn't stand with pyinfra's philosophy. I'm happy you've suggested it that way.
I pushed a commit removing the `user_ops` requirement and also updated the PR description with the proper usage instructions.

Fizzadar (approved):
This is awesome, thank you for the PR @tsnoam!

PR #539, opened 2021-02-25
operation.server.user: allow adding a user with duplicate UID

tsnoam:
Sometimes, one would like to add a user with a duplicate UID. This change would allow this functionality, yet maintain backward compatibility.

I didn't see if and where `operation.server.user` is being properly tested, thus I haven't added a test case for this one.

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -682,6 +682,7 @@ def user(
     present=True, home=None, shell=None, group=None, groups=None,
     public_keys=None, delete_keys=False, ensure_home=True,
     system=False, uid=None, comment=None, add_deploy_dir=True,
+    unique=True,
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thanks for adding this @tsnoam!

For context, tests for operations are here: https://github.com/Fizzadar/pyinfra/tree/master/tests/operations/server.user; I'm going to commit the following test after merge:

```json
{
    "args": ["someuser"],
    "kwargs": {
        "unique": false
    },
    "facts": {
        "users": {}
    },
    "commands": [
        "grep 'someuser:' /etc/passwd || useradd -o someuser",
        "mkdir -p homedir",
        "chown someuser:someuser homedir"
    ]
}
```

PR #545, opened 2021-03-05
Adding Debian to the list of prettified names in LinuxDistribution

blarghmatey:
The LinuxDistribution fact didn't have an entry for Debian in the mapping of `name_to_pretty_name` meaning that the `name` attribute of the returned map requires additional parsing by the end user if running on Debian. This adds that mapping.


Fizzadar (approved):
Thank you for adding this @blarghmatey!

Issue #550, opened 2021-03-13
pyinfra crashes trying to log an operation with a non-string parameter


morrison12:
**Describe the bug**
PyInfra crashes in `_log_operation_start` if an operation is called with a non-string parameter.

**To Reproduce**
with the following in `test.py`:
```
from pyinfra.api import operation

@operation
def example(name, value, host=None, state=None):
    yield "echo test"

example("one", "a")
example("one", 1)
```
try `pyinfra @local test.py`

**Expected behaviour**
One would expect this to run successfully but  instead result is:
```
--> Loading config...
--> Loading inventory...

--> Connecting to hosts...
    [@local] Connected

--> Preparing operations...
    Loading: test.py
    [@local] Ready: test.py

--> Proposed changes:
    Groups: @local
    [@local]   Operations: 2   Commands: 2   

--> Beginning operation run...
--> Starting operation: example (one, a)
    [@local] Success

--> An unexpected internal exception occurred:

  File "/Users/james/.virtualenvs/macos-setup/lib/python3.9/site-packages/pyinfra/api/operations.py", line 216, in _log_operation_start
    args = '({0})'.format(', '.join(arg for arg in op_meta['args']))
TypeError: sequence item 1: expected str instance, int found
```

**Meta**
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Darwin
      Platform: macOS-10.14.6-x86_64-i386-64bit
      Release: 18.7.0
      Machine: x86_64
    pyinfra: v1.3.6
    Executable: /Users/james/.virtualenvs/macos-setup/bin/pyinfra
    Python: 3.9.2 (CPython, Clang 11.0.0 (clang-1100.0.33.17))

Fizzadar:
Thank you for spotting and fixing this @morrison12!

PR #551, opened 2021-03-13
fix _log_operation_start to work with non-string operation arguments

morrison12:
Fix for #550 - crash logging an operation with a non-string parameter.


Fizzadar (approved):
Thanks for fixing this @morrison12!

PR #554, opened 2021-03-17
Improved template documentation

Gaming4LifeDE:


pyinfra/operations/files.py
```diff
@@ -778,15 +778,37 @@ def template(
             group='root',
         )
 
-        # Example showing how to pass python variable to template file.
+        # Example showing how to pass python variable to template file. You can also use dicts and lists
         # The .j2 file can use `{{ foo_variable }}` to be interpolated.
```
Fizzadar, line 782:
```suggestion
        # Example showing how to pass python variable to template file. You can also
        # use dicts and lists. The .j2 file can use `{{ foo_variable }}` to be interpolated.
```

This should fix the linting error (line too long).

Fizzadar (approved):
Great improvement, thank you @Gaming4LifeDE!

PR #557, opened 2021-03-22
Add support for additional winrm transport types and options

mfrg:
Some additional winrm option support, mainly transport usage and timeouts (only transport on the cli though).  More to do here but this is a start.


mfrg:
The printf logging change in here seemed like the right thing to do.  I can revert it if that is problematic, but it seems like a positive change especially when many of those log events are at the debug level.

Fizzadar (approved):
This is fantastic, thank  you for the pr @mfrg!

Totally agree on the logging / printf changes, much more efficient :)

PR #562, opened 2021-03-30
Additional windows ops

mfrg:
There is still much to do here, but the addition of the put/download/link/etc ops allows enough functionality that real work can be done on the windows side now, so I figured it was good to get this PR going and to loop back to continue fleshing things out.

The put op is going to be slow but it does allow transferring content over winrm directly without another service getting involved, which is nice for small files like configs and such.

### DIFF
--- a/pyinfra/api/connectors/util.py
+++ b/pyinfra/api/connectors/util.py
@@ -301,7 +301,7 @@ def make_win_command(command, env=None):
     Builds a windows command with various kwargs.
     '''
 
-    # Use env & build our actual command
+    # TODO: This is not correct, needs to be fixed.
     if env:
         env_string = ' '.join([
             '{0}={1}'.format(key, value)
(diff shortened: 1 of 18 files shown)


Fizzadar:
@mfrg this is *fantastic* - thank you so much for looking into this, it'll be amazing to see pyinfra support Windows properly!

PR wise - would you prefer to keep this PR open and build ontop of it or get this merged / released sooner (1.4) and build any new functionality in future PRs?

mfrg:
@Fizzadar I think merging sooner would be good, this is sort of minimum viable for me to do some xplatform lin/win stuff so I suspect the same could be true for other people as well.  I can open subsequent, likely more targeted, PRs to flesh additional things out.

Fizzadar (approved):
Sounds good to me, @mfrg! Thank you so much for this PR! Merging into the new `next` branch which tracks the upcoming `1.4` release :)

PR #563, opened 2021-03-30
Escape PostgresSQL ROLE and DATABASE

RobWouters:
`postgres.role()` and `postgres.database()` fail if role, database or owner
contains a dash (-) for example.


Fizzadar (approved):
Thanks for spotting & fixing this @RobWouters!

Issue #571, opened 2021-04-19
pyinfra seems to still assume SSH RSA key


Yakulu:
## Describe the bug
pyinfra seems to still only allow SSH RSA key and to raise an exception before other key types can be tried (even after the introduction of other types on 0770560d954db43b20d9ad83d7f6ca1760bf2f66)

I guess problem is here : https://github.com/Fizzadar/pyinfra/blob/6347e352a3b6ed0dc038e50c9ded52af75ae8b40/pyinfra/api/connectors/ssh.py#L88
First `SSHException` is interpreted as  incorrect password and raises on the first iteration of [the loop](https://github.com/Fizzadar/pyinfra/blob/6347e352a3b6ed0dc038e50c9ded52af75ae8b40/pyinfra/api/connectors/ssh.py#L57). 

## To Reproduce
Setup a server with only, for example ed25519 SSH key accepted
Use pyinfra, you should get `pyinfra error: Incorrect password for private key: ~/.ssh/id_ed25519`
Without the SSHException raised, all goes fine.

## Expected behavior
Do not raise before all supported types had been tried.

## Meta
+ Include output of `pyinfra --support`.
```
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Linux
      Platform: Linux-5.4.108-1-MANJARO-x86_64-with-glibc2.33
      Release: 5.4.108-1-MANJARO
      Machine: x86_64
    pyinfra: v1.3.10
    Executable: /home/fabien/.local/bin/pyinfra
    Python: 3.9.2 (CPython, GCC 10.2.0)
```
+ How was pyinfra installed (source/pip)? -> pip

Fizzadar:
Nice catch @Yakulu - I've fixed/removed this logic in https://github.com/Fizzadar/pyinfra/commit/0ba5ecf43ad3d7b13a6a6d681ee91c219ae18e32, pending `1.3.11` release shortly.

Fizzadar:
I've just released this in `v1.3.11`!

Fizzadar:
I've just merged & released your fix (w/new test) in `v1.3.12` @Yakulu, thank you v.much for fixing this properly!

PR #573, opened 2021-04-22
[FIX]#571 suite

Yakulu:
* Ensures loop will continue for all SSH types ;
* Do not raise invalid private key file when key is valid.

The proposed commit 0ba5ecf sadly has not fixed for the use of other types of key (besides RSA).
This pull request tries to ensure all types of keys are tried before raising. Also, it seems that even in this case, `invalid private key` will always be thrown if the key type is not RSA (and so if it's not OK at the first pass).

The MR had been tested locally with RSA and Ed25519 keys. As I only barely know pyinfra internals I'll let you check if you think this is OK or not.

Thanks for your work, have a nice day.

### DIFF
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -79,10 +79,13 @@ def _load_private_key_file(filename, key_filename, key_password):
                         'use this key'.format(key_filename),
                     )
 
-            return key_cls.from_private_key_file(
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
@Yakulu thank you for spotting and properly fixing this! I'm going to merge this in and fix/expand the tests to cover other key types first thing tomorrow.

PR #579, opened 2021-05-03
winrm env var support

mfrg:
This fixes the TODO with env vars not working in winrm.  This should work fine for both cmd and powershell and won't change the shell open/close behavior that pywinrm currently provides.


Fizzadar (approved):
Apologies @mfrg I missed this PR, thank you for fixing this, looks good to me!

PR #583, opened 2021-05-08
propagate return value from @deploy decorator

karlicoss:
e.g. this allows writing code like:

```
  @deploy
  def ripgrep(**kwargs):
      return apt.packages('ripgrep', **kwargs)

  res = ripgrep()
  if res.changed:
       # do something else
```

hopefully it doesn't break anything else?


Fizzadar (approved):
@karlicoss this is a great addition, thank you for the PR!

Issue #592, opened 2021-05-16
casks not supported with brew 2.6.0 and later 


morrison12:
## Describe the bug
In brew 2.6.0 (https://brew.sh/2020/12/01/homebrew-2.6.0/), the CLI for casks was changed, e.g. instead of `brew cask list` one now has to use `brew list --cask`.  This breaks the related facts and operations. 

## To Reproduce
`pyinfra @local fact brew_casks` gives an error:
```
--> Loading config...
    Named facts are deprecated, please use the explicit import: brew.BrewCasks
--> Loading inventory...

--> Connecting to hosts...
    [@local] Connected

--> Gathering facts...
    [@local] Error: Unknown command: cask
    [@local] Error: could not load fact: brew_casks 
    Loaded fact brew_casks 

--> Fact data for: brew_casks
{
    "@local": {}
}
```

as does `pyinfra @local brew.cask magic plot` :
```
--> Loading config...
--> pyinfra error: No such operation: brew.cask
james-mbp:Projects james$ pyinfra @local brew.casks magicplot
--> Loading config...
--> Loading inventory...

--> Connecting to hosts...
    [@local] Connected

--> Preparing operation...
    [@local] Error: Unknown command: cask
    [@local] Error: could not load fact: brew_casks 
--> pyinfra error: No hosts remaining!

```
## Expected behavior
`pyinfra @local brew_casks` should produce a list of casks and `pyinfra @local brew.cask magicplot` should install the magicplot cask

## Meta
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Darwin
      Platform: macOS-10.14.6-x86_64-i386-64bit
      Release: 18.7.0
      Machine: x86_64
    pyinfra: v1.4.1
    Executable: /Users/james/.local/bin/pyinfra
    Python: 3.9.5 (CPython, Clang 11.0.0 (clang-1100.0.33.17))

Fizzadar:
Thank you for the fix @morrison12; I've just released this in `v1.4.2`!

PR #593, opened 2021-05-16
make brew cask facts and operations work with brew ≥ 2.6.0

morrison12:
Address problem noted in https://github.com/Fizzadar/pyinfra/issues/592


Fizzadar (approved):
Thank you for fixing this @morrison12, looks great to me!

PR #594, opened 2021-05-18
Fix a bug in crontab parsing

ghost:
Parsing crontab entries relies on splitting a line on a single space and taking at most 5 splits ([link](https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/facts/server.py#L393)) : 
```
minute, hour, day_of_month, month, day_of_week, command = line.split(' ', 5)
```
This fails if the separator is a tab. The error I encountered parsing OpenBSD default crontab:
```
    data = fact.process(stdout)
  File "/usr/lib/python3.9/site-packages/pyinfra/facts/server.py", line 393, in process
    minute, hour, day_of_month, month, day_of_week, command = line.split(' ', 5)
ValueError: not enough values to unpack (expected 6, got 1)
```
And here the content of the crontab
```
#
#SHELL=/bin/sh
#PATH=/bin:/sbin:/usr/bin:/usr/sbin
#HOME=/var/log
#
#minute hour    mday    month   wday    [flags] command
#
# rotate log files every hour, if necessary
0       *       *       *       *       /usr/bin/newsyslog
# send log file notifications, if necessary
#1-59   *       *       *       *       /usr/bin/newsyslog -m
#
# do daily/weekly/monthly maintenance
30      1       *       *       *       /bin/sh /etc/daily
30      3       *       *       6       /bin/sh /etc/weekly
30      5       1       *       *       /bin/sh /etc/monthly
#~      *       *       *       *       /usr/libexec/spamd-setup

#~      *       *       *       *       -ns rpki-client -v && bgpctl reload
```
 Using the default `split(maxsplit=5)` which splits on white spaces fixes the issue.

PR #595, opened 2021-05-20
Fix second bug in crontab parsing

ghost:
The same bug fixed in #594 is still present in another parsing statement ([link](https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/facts/server.py#L387)) and will cause the same error if the spacing in the cron entries is anything other than a single space. This pull requests fixes it.

```
special_time, command = line.split(' ', 1)
```


Fizzadar (approved):
Thanks for the fix @hazayan!

Issue #596, opened 2021-05-22
User comment with spaces in host.fact.users doesn't parse correctly


sysadmin75:
## Describe the bug
A user with a comment with spaces does not parse correctly in the host.fact.users.

## To Reproduce
Steps to reproduce the behavior (include code & usage example):

Update the test using the following patch:
```
diff --git a/tests/facts/server.Users/mixed.json b/tests/facts/server.Users/mixed.json
index 807e3a95..bb608407 100644
--- a/tests/facts/server.Users/mixed.json
+++ b/tests/facts/server.Users/mixed.json
@@ -3,8 +3,8 @@
         "uid=0(root) gid=0(wheel) groups=0(wheel), 2(kmem), 3(sys), 4(tty), 5(operator), 20(staff), 31(guest) root:/root:/bin/ks",
         "uid=1003(test.testy) gid=1003(test.testy) groups=1003(test.testy) Testy,,,:/home/test.testy:/bin/ksh",
         "uid=1004(_tesTy.test) gid=1004(_tesTy.test) groups=1004(_tesTy.test),1001(pyinfra) :/home/_tesTy.test:/bin/ksh",
-        "uid=1002(noshell) gid=1002(noshell) groups=1002(noshell) noshell:/home/noshell:",
-        "uid=1002(nohome) gid=1002(nohome) groups=1002(nohome) nohome::/bin/bash"
+        "uid=1002(noshell) gid=1002(noshell) groups=1002(noshell) noshell comment with spaces:/home/noshell:",
+        "uid=1002(nohome) gid=1002(nohome) groups=1002(nohome) nohome comment::/bin/bash"
     ],
     "command": "for i in `cat /etc/passwd | cut -d: -f1`; do\n            ID=`id $i`;\n            META=`cat /etc/passwd | grep ^$i: | cut -d: -f5-7`;\n            echo \"$ID $META\";\n        done",
     "fact": {
@@ -46,7 +46,7 @@
         },
         "noshell": {
             "home": "/home/noshell",
-            "comment": "noshell",
+            "comment": "noshell comment with spaces",
             "shell": null,
             "group": "noshell",
             "groups": [],
@@ -55,7 +55,7 @@
         },
         "nohome": {
             "home": null,
-            "comment": "nohome",
+            "comment": "nohome comment",
             "shell": "/bin/bash",
             "group": "nohome",
             "groups": [],
```
The output for the two users with the changes:
```
GOT:
    "noshell": {
[...]

PR #597, opened 2021-05-22
Fix user fact comments wtih spaces parsing by refactoring fact

sysadmin75:
Suggested fix for https://github.com/Fizzadar/pyinfra/issues/596

I understand the amount of changes I propose seem excessive for the fix. I felt using complicated regexes for an output that can be controlled could be simplified.


Fizzadar (approved):
Big fan of this @sysadmin75 - much simpler and cleaner implementation, thank you!

PR #602, opened 2021-05-28
Update files.py

mrshu:
* Add a bit of documentation on `assume_present` in `file.line` operator


Fizzadar (approved):
Nice spot, thank you for the fix @mrshu!

PR #604, opened 2021-05-31
Fix brew casks fact for non-mac platforms

stevenkaras:
This command contains a bashism which causes parsing errors for some shells (for example dash as included in raspbian).
POSIX test doesn't include any pattern matching, so fall back on grep as the next best thing.

tested against both a mac (running bigsur) and an rpi (running raspbian)


Fizzadar (approved):
@stevenkaras thank you for spotting & fixing this!

Issue #581, opened 2021-05-06
Add last login time to `host.fact.users`


kkirsche:
## Is your feature request related to a problem? Please describe
Many corporate environments have requirements related to identifying and removing "stale" user accounts on servers. This is often then delegated to system owners for identification based on the last login date/time of the user. This leads to one-off shell scripts and other ad-hoc methods of achieving this.

## Describe the solution you'd like
I would like for `last_login_at` (or an equivalent key name) to be added to the (populated via `lastlog`?) to be added to the `host.fact.users` facts to simplify the identification of these user accounts.

kkirsche:
Fixed by #606

PR #606, opened 2021-06-12
Add last login time to host.fact.users

sysadmin75:
Adds last login time to host.fact.users info
Addresses request in #581

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -422,6 +422,7 @@ class Users(FactBase):
                 ],
                 'uid': user_id,
                 'gid': main_user_group_id,
+                'lastlog': last_login_time,
             },
         }
     '''
@@ -429,7 +430,8 @@ class Users(FactBase):
     command = '''
         for i in `cat /etc/passwd | cut -d: -f1`; do
             ENTRY=`grep ^$i: /etc/passwd`;
-            echo "$ENTRY|`id -gn $i`|`id -Gn $i`";
+            LASTLOG=`lastlog -u $i | grep ^$i` | tr -s ' ';
+            echo "$ENTRY|`id -gn $i`|`id -Gn $i`|$LASTLOG";
         done
     '''.strip()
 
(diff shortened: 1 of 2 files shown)


pyinfra/facts/server.py
```diff
@@ -452,6 +455,11 @@ def process(self, output):
                     if group_name and group_name != group:
                         groups.append(group_name)
 
+                # Parse lastlog info
+                # lastlog output varies, which is why I use regex to match login time
+                login = re.search(rex, lastlog)
+                login_time = login.group() if login else '**Never logged in**'
```
Fizzadar:
```suggestion
                login_time = login.group() if login else None
```

I think a `None` for never logged in here is more Python-y and easier to compare to.

pyinfra/facts/server.py
```diff
@@ -460,6 +468,7 @@ def process(self, output):
                     'groups': groups,
                     'uid': int(entries[2]),
                     'gid': int(entries[3]),
+                    'lastlog': login_time,
```
Fizzadar, line 471: As well as attaching the raw `lastlog` it might be nice to have parsed version of the date under key `login_time` or similar? ([could use dateutil similar to the `server.Date` fact](https://github.com/Fizzadar/pyinfra/blob/de8d4f3b364ea6ea7ece4841ebe97771bc002ae5/pyinfra/facts/server.py#L127)).

Fizzadar (commented):
@sysadmin75 thank you for implementing this, it looks great! Left a couple of suggestions it'd be good to get your feedback on before merging.

pyinfra/facts/server.py
sysadmin75: I went with '''**Never logged in**''' because that's the output from lastlog.  However, I agree that using `None` in this case is probably better.

pyinfra/facts/server.py
sysadmin75: I'm not sure what you mean by having a parsed version of the date for `login_time`.  The current format of the is straight from `lastlog` and would look something like
```
Sat Jun 19 14:18:07 -0400 2021
```
What change are you wanting specifically?  I would assume if a user wants something more specific, they can parse the output of the fact themselves.  I'm interested in hearing what you have in mind.

Fizzadar: Sorry that wasn't super clear, I think it's nice to provide the user a ready to go `datetime` object from the fact; thinking something like this:

```py
from dateutil.parser import parse as parse_date
...
{
    'lastlog': login_time,
    'login_time': parse_date(login_time),
}
```

Fizzadar (approved):
Thank you for the addition @sysadmin75, will merge this and add the parsed time sometime later, ready to be released in `v1.5`!

PR #607, opened 2021-06-12
WIP: serial console connector

yggdr:
Allow pyinfra to connect to a machine by serial connectors (e.g. RS-232), where one would otherwise use e.g. `minicom`, or `cu`.

So far no login support.


Fizzadar:
Same comment as #416:

This is still a really cool idea but should definitely be managed as an external project. v2 & upcoming v3 both support this via setuptools entrypoints, I am going to provide an example project as part of the upcoming v3 release cycle at which point I shall link it here and close this PR.

Nicoretti:
I guess this PR can/should be closed?

PR #608, opened 2021-06-19
fix: add missing mysql connection parameters

sfermigier:
This fixes a failure to run the `mysql.create` operation when using non-default connection parameters.


Fizzadar (approved):
@sfermigier thank you for spotting & fixing this!

PR #612, opened 2021-06-27
Files explicit fact imports

Fizzadar:
Shifts the files operations to use explicitly imported facts, which migrates by far the most used set of facts over.


Fizzadar:
Made a mess of this one!

Issue #614, opened 2021-06-28
server.script fails when using non-privileged users over SSH


benridley:
## Describe the bug
When using a non-privileged user for SSH, and then executing a `server.script` operation with `sudo=True` and `sudo_user=<unprivileged_user>`, the operation fails with the error: 

`File upload error: mv: cannot move '/tmp/pyinfra-a52ee0536b1bfb3242676133b584ab07879526c4' to '/tmp/pyinfra-a5ab867674785ba103efd18d20f57084bef199fb': Operation not permitted`

## To Reproduce
Ensure you have set `ssh_user` to an unprivileged account. 

Create a simple deploy:
```python
from pyinfra import host
from pyinfra.operations import server

server.script(
    "files/hello.sh",
    sudo=True,
    sudo_user="nobody",
)
```
```bash
#files/hello.sh

#!/bin/bash
echo 'Hello, world!'
```
Run the deploy. 

## Expected behavior
The script executes as the specified user. 

## Meta
I've seen this problem come up in Ansible and also Puppet Bolt, generally its caused because the script is copied over using the unprivileged user's SSH session, which prevents other users from executing it. You then run into some nasty thorns:
- There's no way to give away ownership to the user you want to sudo to in Linux, ownership can only be changed by root. So you can't `chown` the script before changing users. 
- Changing the script to be world readable/executable is a security issue. 

The best solution I know of, which I believe Ansible went with, is by using ACLs to assign execute permissions to the user before executing the `sudo`, which is possible between two unprivileged users. This does mean you rely on filesystems with ACL implementations though.

benridley:
I've created a PR at #616 to demonstrate one method of addressing this by using ACLs and copying the file instead of `mv`ing it. There's some issues with it so happy to discuss further how we might address this best.

Fizzadar:
PR merged & ready for release in `v1.5`!

benridley:
Awesome @Fizzadar, thanks again. I validated my existing code against the `next` branch and its all working as expected, so I think we can close this one out.

PR #616, opened 2021-06-30
Transfer ownership of files over ssh with sudo/su

benridley:
This PR addresses #614 by using the SSH user to assign a read-only ACL to the uploaded file, before switching to the sudo/su user and copying the file to the intended destination rather than attempting to `mv` it.

This allows a file uploaded by an unprivileged user to be 'handed over' to another unprivileged user. 

I've created this PR as a draft because there's a few issues worth discussing:
- This approach relies on the ability to use filesystem ACLs. Most modern filesystems support this but not all of them do, so I wanted to make sure this fits with pyinfra's philosophy. We can always try this and fall back to another approach as well.
- This introduces `cp` instead of `mv` when transferring the file out of the temp directory on upload, so for large files this could be a significant amount of extra storage consumed during the upload process.

I've also tried to update the tests, one of them updated ok but another is failing because it's asserting against the wrong call (it's using the `setfacl` call rather than the `cp` call). Happy to fix this up if I can get some assistance.

### DIFF
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -420,8 +420,25 @@ def put_file(
         temp_file = state.get_temp_filename(remote_filename)
         _put_file(host, filename_or_io, temp_file)
 
+        # Make sure our sudo/su user can access the file
+        if su_user:
+            command = StringCommand(f'setfacl -m u:{su_user}:r', temp_file)
+        elif sudo_user:
+            command = StringCommand(f'setfacl -m u:{sudo_user}:r', temp_file)
+        status, _, stderr = run_shell_command(
+            state, host, command,
+            sudo=False,
+            print_output=print_output,
+            print_input=print_input,
+            **command_kwargs
+        )
+
+        if status is False:
+            logger.error('Error on handover to sudo/su user: {0}'.format('\n'.join(stderr)))
+            return False
+
         # Execute run_shell_command w/sudo and/or su_user
-        command = StringCommand('mv', temp_file, QuoteString(remote_filename))
+        command = StringCommand('cp', temp_file, QuoteString(remote_filename))
 
         # Move it to the su_user if present
         if su_user:
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -422,7 +422,7 @@ def put_file(
 
         # Make sure our sudo/su user can access the file
         if su_user:
-            command = StringCommand('setfacl -m u:{0}:r'.format(su_user), temp_file)
+            command = StringCommand('setfacl', '-m', 'u:{0}:r'.format(su_user), temp_file)
         elif sudo_user:
             command = StringCommand('setfacl -m u:{0}:r'.format(sudo_user), temp_file)
         status, _, stderr = run_shell_command(
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -452,9 +452,9 @@ def put_file(
         if status is False:
             logger.error('File upload error: {0}'.format('\n'.join(stderr)))
             return False
-    
+
         # Delete the temporary file now that we've successfully copied it
-        command = StringCommand('rm', '-f', temp_file) 
+        command = StringCommand('rm', '-f', temp_file)
 
         status, _, stderr = run_shell_command(
             state, host, command,
--- a/pyinfra/api/connectors/ssh.py
+++ b/pyinfra/api/connectors/ssh.py
@@ -441,14 +441,6 @@ def put_file(
         # Execute run_shell_command w/sudo and/or su_user
         command = StringCommand('cp', temp_file, QuoteString(remote_filename))
 
-        # Move it to the su_user if present
-        if su_user:
-            command = StringCommand(command, '&&', 'chown', su_user, QuoteString(remote_filename))
-
-        # Otherwise any sudo_user
-        elif sudo_user:
-            command = StringCommand(command, '&&', 'chown', sudo_user, QuoteString(remote_filename))
-
         status, _, stderr = run_shell_command(
             state, host, command,
             sudo=sudo, sudo_user=sudo_user, su_user=su_user,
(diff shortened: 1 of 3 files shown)


benridley:
Ah, apologies I meant to create this as a draft PR and now don't seem to be able to change it

Fizzadar (commented):
@benridley thank you for looking into this! I like the approach here, seems like a sensible solution to the problem. My only concern is the change to `cp` from `mv`, I'm wondering if there's a way to support using `mv` with acl's here; will do some testing locally over the next few days on that...

benridley:
Cheers @Fizzadar!

I've been using this in my fork of pyinfra for a week or so now for standing up dev environments fairly heavily, and I haven't hit any issues yet. 

I've also done some more investigation into the `mv` vs `cp` semantics that might help you in your investigation: The `setfacl` command demonstrated here actually does allow the other user to `mv` the file under normal circumstances. The problem is actually that the `/tmp` directory on most distributions has the sticky bit set. The `mv` command uses the `rename` syscall under the hood, and `rename` fails on directories with sticky bits set unless the uids match (see EPERM [on this link](https://man7.org/linux/man-pages/man2/rename.2.html)). So, when using /tmp unfortunately I can't see any way around using `cp`. 

The good news is that when the files are not on the same filesystem, `rename` doesn't work (see EXDEV in the link above), so `mv` actually does a copy and delete under the hood. Usually `/tmp` is mounted on its own filesystem, so in cases where a user is moving a file somewhere other than `/tmp` (which I would think would be most cases), and `/tmp` is just used as pyinfra's staging area, this change actually doesn't result in any change in behaviour provided we delete the temporary file after the `cp` operation.

Fizzadar:
@benridley that's incredibly helpful, thank you! I did recall there is actually already [a note on the operation docs](https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/operations/files.py#L670-L672) that it's not suitable for large files due to potential copying.

Given the semantics of `mv`/`cp` I think using `cp` is fine here, if we're on the same filesystem the sticky bit makes it problematic and if separate filesystems `mv` will copy the file anyway. Switching to `cp` only keeps things nice and consistent and makes it possible to solve this problem.

pyinfra/api/connectors/ssh.py
```diff
@@ -420,8 +420,25 @@ def put_file(
         temp_file = state.get_temp_filename(remote_filename)
         _put_file(host, filename_or_io, temp_file)
 
+        # Make sure our sudo/su user can access the file
+        if su_user:
+            command = StringCommand('setfacl -m u:{0}:r'.format(su_user), temp_file)
```
Fizzadar:
```suggestion
            command = StringCommand('setfacl', '-m', 'u:{0}:r'.format(su_user), temp_file)
```

pyinfra/api/connectors/ssh.py
```diff
@@ -420,8 +420,25 @@ def put_file(
...
+        elif sudo_user:
+            command = StringCommand('setfacl -m u:{0}:r'.format(sudo_user), temp_file)
```
Fizzadar, line 427:
```suggestion
            command = StringCommand('setfacl', '-m', 'u:{0}:r'.format(sudo_user), temp_file)
```

Minor nitpick on the style, makes no difference in output terms but consistent with `StringCommand` use!

pyinfra/api/connectors/ssh.py
```diff
@@ -420,8 +420,25 @@ def put_file(
...
+            **command_kwargs
+        )
+
+        if status is False:
+            logger.error('Error on handover to sudo/su user: {0}'.format('\n'.join(stderr)))
+            return False
+
         # Execute run_shell_command w/sudo and/or su_user
-        command = StringCommand('mv', temp_file, QuoteString(remote_filename))
+        command = StringCommand('cp', temp_file, QuoteString(remote_filename))
```
Fizzadar:
```suggestion
        command = StringCommand('cp', temp_file, QuoteString(remote_filename), '&&', 'rm', '-f', temp_file)
```

I think it makes sense to remove the temporary file here once the copy is complete.

tests/test_connectors/test_ssh.py
```diff
@@ -540,7 +540,7 @@ def test_put_file_su_user_fail(self, fake_sftp_client, fake_ssh_client):
         assert status is False
 
         fake_ssh_client().exec_command.assert_called_with((
-            "su centos -c 'sh -c '\"'\"'mv "
+            "su centos -c 'sh -c '\"'\"'cp "
```
Fizzadar, line 578: This command will now be the `acl` one I believe - as the mock above is failing the first shell command executed. Changing this to assert for the acl command should cover the test.

Would also be good to have an addition test for the `cp` command failing vs. first ACL one, but I'm happy to add that in as well.

Fizzadar (commented):
This is awesome @benridley! Left a couple of minor things and the possible test fix :)

pyinfra/api/connectors/ssh.py
benridley: I think for this one we will need to do the removal outside the sudo session, otherwise we'll end up with the same problem as `mv` where we can't delete it due to the sticky bit. I'll give something a try tomorrow and make sure it works!

benridley:
Hi @Fizzadar, I've just pushed some new changes. Now the file is removed after the `cp` operation by running outside `sudo`. I tested this on the environment I've been working on and so far it's working as inteded.

I also fixed up the test cases so they are all passing, but you're correct that now the assertions are against the `setfacl` command, and now I've added the `rm` one of the other tests is asserting against that. 

If it's not too much pain would you mind assisting with adding the additional assertions? I'm happy to do so, but it might take me some time to get my head around it since I've not worked with pytest and mock etc before.

Take a look and let me know what you think, cheers!

pyinfra/api/connectors/ssh.py
benridley: I've added an `rm` outside the `sudo` session to remove the file now.

Fizzadar:
@benridley thank you for updating this; I've just pushed updated tests for failures at each stage in https://github.com/Fizzadar/pyinfra/pull/616/commits/1a2f68640e073bbdd74df795e2d06b486222a04b.

benridley:
Awesome! Thanks @Fizzadar. Anything further you'd like to see on this PR?

Fizzadar (approved):
All good, thank you very much for working on this @benridley!

PR #618, opened 2021-07-03
feat: fetch held deb packages' version

GerardoGR:
It would be useful to fetch the version number of held deb packages to have scripts that manually unhold them to be able to upgrade if the need be.


Fizzadar (approved):
Thank you for the PR @GerardoGR, I think this is an excellent addition 👍.

PR #623, opened 2021-07-12
fix(ssh.upload): fix wrong connection string when using use_remote_sudo and user

GerardoGR:
Fixes the ssh connection string when setting the use_remote_sudo and user options for the ssh.upload operation

Before: sets twice the username
```json
[
  "scp -P 22 local_filename test@remote-host.net:_tempfile_",
  "ssh -p 22 test@test@remote-host.net 'sudo mv _tempfile_ local_filename'"
]
```

After:
```json
[
  "scp -P 22 local_filename test@remote-host.net:_tempfile_",
  "ssh -p 22 test@remote-host.net 'sudo mv _tempfile_ local_filename'"
]
```

### DIFF
--- a/pyinfra/operations/ssh.py
+++ b/pyinfra/operations/ssh.py
@@ -157,7 +157,7 @@ def upload(
 
         # And sudo sudo to move it
         yield command(
-            hostname=connection_target,
+            hostname=hostname,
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice catch, thank you for fixing this @GerardoGR!

PR #634, opened 2021-08-05
files.download: fix cache_time test

stevenkaras:
the condition was flipped. We want to re-download the file only if it hasn't been modified since `cache_time`.


Fizzadar (approved):
Major facepalm moment for me here - thank you for spotting, fixing AND adding a test for @stevenkaras!

PR #637, opened 2021-08-10
WIP: add pipx support

karolzlot:
This is attempt to add pipx support. Pipx is is great tool to install many packages, I want to use it to install for example `docker-compose` and `docker-auto-labels` on hosts.

-------------

I have problem, which I can't understand how to fix. I can't run `pipx` from pyinfra, but I can run it from manual ssh session. It works ok if `pipx` is installed via `apt`, but version in `apt` is very outdated.

Example code below generates error:

```python
from pyinfra.api import Config
from pyinfra.operations import apt, pip, pipx

apt.packages(
    name='Ensure python is installed',
    packages=['python3-pip', 'python3-venv'],
    latest=True,
    sudo=True,  # use sudo when installing the packages
    update=True,
    cache_time=3600,
    upgrade=True,
)

pip.packages(
    name="Ensure pipx is installed",
    packages=["pipx"],
    pip="pip3",
    latest=True,
    # sudo=True,  # If sudo is enabled here, then error vanishes, but sudo is not recommended for pip
)

pipx.packages(
    name="Ensure pycowsay is installed",
    packages=["pycowsay"],
    latest=True,
)
```
Error message:
```
--> Starting operation: Ensure pycowsay is installed
[REDACTED_IP] >>> sudo -H -n sh -c 'pipx install pycowsay'
[REDACTED_IP] sh: 1: pipx: not found
    [REDACTED_IP] Error
No hosts remaining!
--> pyinfra error: 
```

### DIFF
--- a/pyinfra/facts/pipx.py
+++ b/pyinfra/facts/pipx.py
@@ -27,7 +27,7 @@ class PipxPackages(FactBase):
         return '{0} list --json'.format(pipx)
 
     def process(self, output):
-        pipx_list= json.loads(output)
+        pipx_list= json.loads('\n'.join(output))
         versions={}
         for venv in pipx_list['venvs'].values():
             package=venv['metadata']['main_package']['package']
--- a/pyinfra/facts/pipx.py
+++ b/pyinfra/facts/pipx.py
@@ -1,0 +1,36 @@
+from __future__ import unicode_literals
+
+from pyinfra.api import FactBase
+
+import json
+
+
+class PipxPackages(FactBase):
+    '''
+    Returns a dict of installed pipx packages:
+
+    .. code:: python
+
+        {
+            'package_name': ['version'],
+        }
+    '''
+
+    default = dict
+    pipx_command = 'pipx'
+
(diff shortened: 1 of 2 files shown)


karolzlot:
Another issue is that `pipx` needs separate commands, it won't work with:
```
pipx install pycowsay copier docker-compose
```
It needs:
```
pipx install pycowsay 
pipx install copier
pipx install docker-compose
```

I'm not sure if `ensure_packages` is able to work this way.

karolzlot:
Problem is pyinfra uses `sh -c ' command.... '` to execute commands and `$PATH` is different because of it. I would like to have an option to use `bash` instead of `sh`.

Those commands give interesting, different results:

```
echo $PATH
sh -c 'echo $PATH'
sudo sh -c 'echo $PATH'
```

karlicoss:
For multiple packages issue, here's the related pipx issue https://github.com/pypa/pipx/issues/88

Fizzadar:
#1243 has implemented this now.

PR #638, opened 2021-08-10
Add Arch Linux support for the LinuxDistribution fact

TypicalFence:
Arch has an empty files at `/etc/arch-release` that was used back when arch had releases. This lead to straight up errors, I fixed those more broadly by just spiking empty files.

I hope this can get merged as I really wanna use pyinfra for my arch servers.


Fizzadar (approved):
@TypicalFence thank you for locating and fixing this! Should be out in `v1.4.12` shortly :)

PR #639, opened 2021-08-12
git.GitBranch: support checking out tags

stevenkaras:
Checking out a git repo is a last resort, so I always like to pin them to tags when I do have to have one on a host.
`git describe` is considered plumbing and thus format stable. This behavior was introduced in 2005, so there are no
real concerns that a host will be running an existing version of git that doesn't support the command or the `--all`
flag.

### DIFF
--- a/pyinfra/facts/git.py
+++ b/pyinfra/facts/git.py
@@ -10,7 +10,11 @@ class GitBranch(FactBase):
 
     @staticmethod
     def command(repo):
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Thank you for this change, @stevenkaras - much better!

GitHub is refusing to place nice but the tests _do_ pass just fine.

PR #648, opened 2021-08-19
Add some basic SELinux facts & additional fact documentation

benridley:
This PR adds some simple SELinux facts that I've found useful:
 - `selinux.FileContext`: Provides SELinux file contexts for a given path.
 - `selinux.SEBoolean`: Provides the on/off status of an SELinux boolean.

I also added some documentation for the `default` function and `requires_command` variable when writing a custom fact which I found helpful.


Fizzadar (approved):
This is great, thank you @benridley!

PR #649, opened 2021-08-19
Improve dir sync exclude

gchazot:
Currently `files.sync`'s `exclude_dir` parameter doesn't seem to work properly, and it seems to be because it only ignores  matching directories _themselves_, i.e. not the files or directories they contain.

As a consequence, it only seems to be relevant for **empty** directories.

This changes this behaviour by excluding the entire directory tree below matching directories.


Fizzadar (approved):
This is fantastic, thank you @gchazot!

PR #650, opened 2021-08-23
Updates fact finding for links / files / dirs

mattalexanderhill:
These functions take an expression which may include wildcards. As these are now quoted the shell find expression no longer works. Update form from `find '/fo*/b*r' type f`  ->  `find / -path '/fo*/b*r' -type f`

An example of the wildcard pattern is here: https://github.com/Fizzadar/pyinfra/blob/e556d326a923a37ecb55e83726e4df33ff26e1fa/pyinfra/operations/sysvinit.py#L62

And in action: 

```
$ find "/etc/rc*.d/S*ntp" -type l
find: ‘/etc/rc*.d/S*ntp’: No such file or directory

$ find "/etc/rc2.d/S01ntp" -type l
/etc/rc2.d/S01ntp
```

This used to work before the arguments were quoted, see #611 as the wildcards were expanded by the shell. I'm not sure that `find` supports wildcard in the path argument.

### DIFF
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -234,13 +234,13 @@ class FindInFile(FactBase):
 
 class FindFiles(FactBase):
     '''
-    Returns a list of files from a start point, recursively using find.
+    Returns a list of files matching a pattern, recursively from root using find.
     '''
 
     @staticmethod
     def command(path):
         return make_formatted_string_command(
-            'find {0} -type f || true',
+            'find / -path {0} -type f || true',
             QuoteString(path),
         )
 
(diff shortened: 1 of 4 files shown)


mattalexanderhill:
Maybe worth noting that this is not doing the same thing as before the path quoting as this will require `find` traversing the filesystem.

Fizzadar:
Hi @mattalexanderhill! Unfortunately I think this change might break a lot of assumptions with using `find` as the `-path` argument doesn't behave the same as the positional path, ie:

```
find /somepath
/somepath
/somepath/somefile
/somepath/anotherfile

# vs:

find / -path /somepath
/somepath
```

You would have to explicitly pass `/somepath*` to have similar result.

Regarding this issue overall - the `sysvinit.service` issue needs fixing as this is a breaking bug. The easiest option here is to optionally disable path quoting (at increased risk) in the fact, which I've just added in https://github.com/Fizzadar/pyinfra/commit/1477b5d26081b47ade59c44fcac9f8918eebba29 and fixed the op in https://github.com/Fizzadar/pyinfra/commit/d55b141dff32874ea7cb6d29eb8f2667fa616aa6 so this should now be working as of `v1.14.14`!

Issue #533, opened 2021-02-20
sed replace quoting (server.crontab fails to update crontab entry sometimes)


taranlu-houzz:
**Describe the bug**
I have a feeling this is caused by the `cron_name` param. Basically, if I modify the `command` param for `server.crontab` but leave `cron_name` as is, when the deploy is run, it will consider the task to already be present resulting in a no-op.

**To Reproduce**
- Run on host: `crontab -r`
- Run the below deploy once and verify the crontab updated.
- Change the command, and run the below deploy again.
- The crontab will not have been updated.

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-


"""Deploy to illustrate crontab bug.
"""


###############################################################################
# Imports
###############################################################################


# ==Site-Packages==
from pyinfra.operations import server


###############################################################################
# Main
###############################################################################


server.crontab(
    name="Install test cron task.",
    command="echo 'banana'",
    #  command="echo 'yummy'",
    cron_name="My special task",
    minute="*/10",
)

```

**Expected behavior**
The crontab should be updated to reflect the changes to the command.

**Meta**
```
❯ pyinfra --support
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Darwin
      Platform: macOS-10.14.6-x86_64-i386-64bit
      Release: 18.7.0
      Machine: x86_64
    pyinfra: v1.3.6
    Executable: /Users/----/.virtualenvs/05-hz-pyinfra-d0Faa_8y-py3.9/bin/pyinfra
    Python: 3.9.1 (CPython, Clang 10.0.1 (clang-1001.0.46.4))
```

Fizzadar:
Nice spot @taranlu-houzz, little bug in checking the command when using `cron_name`, fixed in https://github.com/Fizzadar/pyinfra/commit/228460d004175ebd539bba641d27773eb955bc40 for release in `1.3.7` shortly!

taranlu-houzz:
I'm still seeing this issue on `1.3.7`. Did the fix not make it into that release?

Fizzadar:
It did! Will reopen this to investigate :grimacing:

Fizzadar:
So it looks like this happens due to a error handling the quotes in the command, for example the following file does change the command correctly:

```py
server.crontab(
    name="Install test cron task.",
[...]

PR #652, opened 2021-08-24
Crontab rewrite

taranlu-houzz:
I made this PR in order to deal with some of the various issues that I have run into when using the crontab operation (recently #533).

Currently, this patch resolves all issues that I have run into personally (I managed to track down the cause of #533 to be an issue with the way `sed_replace` works at least in my specific edge case). I feel like `sed` is pretty tricky to work with when it comes to catching all of the possible edge cases...

@Fizzadar I was hoping that I could get some feedback to see if this is something that you would consider integrating into `pyinfra`.

Advantages:
- `Crontab` fact is more robust, and the entirety of the crontab is gathered, in order, so that any part could be modified (I feel like it could be nice to add a `normalize_whitespace` param that would turn all whitespace chunks into a single blank line, for example).
- `Crontab` operation add/remove/update logic is a bit more straightforward and clear in my opinion.
- `sed` issues with edge cases are avoided.

Remaining issues:
- Need to format code to match `pyinfra` style spec (single quotes mainly).
- Need to get all tests passing (the fact tests were simple to update, but I am a little confused by the generated operation tests).
- I'm not entirely clear on how `interporlate_variables` is supposed to work. I assumed it means that you could have something like `$USER` in the command argument, and it would resolve that before actually writing to the crontab, but that did not seem to work when I tested on `current`.
- Need to add appropriate error for invalid arguments (e.g. `special_time` param). Not sure how that is handled elsewhere in `pyinfra`.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -416,7 +416,6 @@ class Crontab(FactBase):
         
         chunks = []
         current_chunk = _empty_chunk()
-        #  breakpoint()
         for line in output:
             # Get line type
             line_type = _line_type(line)
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -344,22 +344,41 @@ class Groups(FactBase):
 
 class Crontab(FactBase):
     '''
-    Returns a list of dictionaries where each entry corresponds to a single cron command.
+    Returns a list of dictionaries where each entry corresponds to a single crontab "chunk".
+
+    Each chunk can have one of the following types:
+        - command 
+            - Also includes all comment lines directly above the command line.
+        - env
+            - Also includes all comment lines directly above the command line.
+        - comment 
+        - whitespace
 
     .. code:: python
 
-        {
-            '/path/to/command': {
-                'minute': '*',
-                'hour': '*',
-                'month': '*',
-                'day_of_month': '*',
-                'day_of_week': '*',
+        [
+            {
+                "data": [
+                    "# My special comment.",
+                    "# Multiple comment lines are fine.",
+                    "# pyinfra-name=Task A",
+                    '*/5 * * * * echo "hello"',
+                ],
+                "type": "command",
             },
-            'echo another command': {
-                'special_time': '@daily',
+            {
+                "data": [
+                    "",
+                ],
+                "type": "whitespace",
             },
-        }
+            {
+                "data": [
+                    '@daily echo "My daily task."',
+                ],
+                "type": "command",
+            },
+        ]
     '''
 
     default = list
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -344,7 +344,7 @@ class Groups(FactBase):
 
 class Crontab(FactBase):
     '''
-    Returns a dictionary of cron command -> execution time.
+    Returns a list of dictionaries where each entry corresponds to a single cron command.
 
     .. code:: python
 
@@ -362,9 +362,14 @@ class Crontab(FactBase):
         }
     '''
 
-    default = dict
+    default = list
 
     requires_command = 'crontab'
+
+    # Compile regex
+    whitespace_regex = re.compile(r"^\w*$")
+    comment_regex = re.compile(r"^#.+$")
+    command_regex = re.compile(r"^[\*\d@].+$")
 
     @staticmethod
     def command(user=None):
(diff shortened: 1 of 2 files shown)


pyinfra/facts/server.py
```diff
@@ -344,7 +344,7 @@ def process(output):
 
 class Crontab(FactBase):
     '''
-    Returns a dictionary of cron command -> execution time.
+    Returns a list of dictionaries where each entry corresponds to a single cron command.
```
Fizzadar: This is a much cleaner approach - I always regretted the command -> details dict choice here!

One problem here is that this is a breaking change - could we make this a new class (maybe just `Crons`? Or `CronEntries`?) and leave the old one as-is, I can then add a deprecation notice when used pending removal in v2.

Fizzadar (changes requested):
This is a _huge_ improvement - thank you so much for taking the time to rewrite this!

The only change required is the fact class name to avoid breaking existing implementations, once that's swapped out (and tests/linting pass) I think this is good to go 👍

taranlu-houzz:
@Fizzadar Ah, good to hear! I may not be able to get to this for a little while, but I do plan to continue on this when I have time (will be out on leave for about a month).

In the meantime, these are a couple of questions that I have:
- Could you help me to understand how your operation tests are generated? I found the json that generates the tests to be a little confusing.
- How do operations typically handle errors for invalid parameters? I have a check in place for the `special_time` param, but I wasn't quite sure what the standard process is elsewhere in `pyinfra`.
- I was also hoping to get some clarification on how the `interpolate_variables` param is supposed to function. It didn't seem to do what I was expecting...

I have a few more questions, but I'll get to those when I get back to this PR to tidy it up and get it ready for merging.

pyinfra/facts/server.py
taranlu-houzz: Ah, yeah, I can make that change. Will get to that when I have some more time to work on this PR.

Fizzadar:
Hey @taranlu-houzz! No worries at all hugely appreciate you taking the time to work on this; answers below:

>     * Could you help me to understand how your operation tests are generated? I found the json that generates the tests to be a little confusing.

The JSON files should roughly define, for a given operation, the inputs (args/kwargs) and the commands output as a result. There's also ways to define facts/filesystem objects/etc which have been tacked on over time. I need to do a proper write up of the format of these because I'm sure this will help many contributors - it's certainly quite confusing at the moment (https://github.com/Fizzadar/pyinfra/issues/656).

>     * How do operations typically handle errors for invalid parameters? I have a check in place for the `special_time` param, but I wasn't quite sure what the standard process is elsewhere in `pyinfra`.

There's an `OperationError` class that works nicely for this; example @ https://github.com/Fizzadar/pyinfra/blob/27f4732ae712f3ae158fbb00c98e5319e3f9eefd/pyinfra/operations/mysql.py#L271-L275.

>     * I was also hoping to get some clarification on how the `interpolate_variables` param is supposed to function. It didn't seem to do what I was expecting...

The idea here is that it will double-quote any user provided strings in the output shell - that should then enable environment variable substitution in those strings.

> I have a few more questions, but I'll get to those when I get back to this PR to tidy it up and get it ready for merging.

👍

taranlu-houzz:
@Fizzadar I'm back from leave and will hopefully be able to get this wrapped up soon. I do have a further question about the `interpolate_variables` param: Could I get a couple of example cases showing the difference this setting makes?

In some basic test using the below deploy, it doesn't seem to make any difference (I think I am probably still just misunderstanding what it does):

```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-


"""Test crontab interpolate_variables param."""


####################################################################################################
# Imports
####################################################################################################


# ==Site-Packages==


####################################################################################################
# Main
####################################################################################################


server.crontab(
    name=f"Are my variables interpolated yet?",
    command='echo "$USER" "toot" "oop"',
    special_time="@reboot",
    interpolate_variables=True,
)
```

I did run several variations with `--debug` set, so I can see that it can change the `sed` command that gets run, but the resulting crontab doesn't seem to be different.

Fizzadar:
@taranlu-houzz the variables will only render if not quoted at all, for example:

```py


files.line(
    path='somefile',
    line='this is $USER data',
    interpolate_variables=True,
)

files.line(
    path='somefile',
    line='this is $USER data',
)
```

Results in a file:

```
this is nick data
this is $USER data
```

This has uncovered an unrelated bug (fixed by https://github.com/Fizzadar/pyinfra/commit/adaf051d25bd8c278eb2456eda982349a72f8b9d).

Fizzadar:
https://github.com/pyinfra-dev/pyinfra/pull/1193 has now implemented this (not identical, but functionality wise).

PR #659, opened 2021-09-18
Add pkgin support

### LINKED ISSUE #632: pkgin support
Using https://github.com/Fizzadar/pyinfra/issues/617#issuecomment-884766468 to submit an idea I had!

I have some NetBSD boxes here and there and would like to manage them with pyinfra, one step in that direction would be support for [pkgin](https://pkgin.net/), which is an user-friendly frontend to pkgsrc (which is already "provided" by the `pkg` module). 

A new `pkgin` module would be the idea, providing operations with the same semantics as the `apt` module with `packages()`, `update()`, and `upgrade()`:
 - `packages()` can be implemented because `pkgin list` shows all currently installed packages, so you can discard unecessary reinstallations
 - Maybe `autoremove()` could be done as well?

I can provide a NetBSD VM in my infrastructure for testing purposes, if you wish to!

lun-4:
Close #632!

I based my implementation from `apk`'s facts and operations, save for version pinning as pkgin supports setting minimum and maximum versions, so I was confused on how to express that on the operation. I added a TODO comment on it as I think that would be useful future work, but not immediate.

Successfully tested all operations in a NetBSD 9.2 amd64 VM.

### DIFF
--- a/pyinfra/operations/pkgin.py
+++ b/pyinfra/operations/pkgin.py
@@ -1,0 +1,81 @@
+'''
+Manage pkgin packages.
+'''
+
+from pyinfra.api import operation
+from pyinfra.facts.pkgin import PkginPackages
+
+from .util.packaging import ensure_packages
+
+
+@operation
+def upgrade(state=None, host=None):
+    '''
+    Upgrades all pkgin packages.
+    '''
+
+    yield 'pkgin upgrade'
+
+_upgrade = upgrade  # noqa: E305
+
+
+@operation
+def update(state=None, host=None):
+    '''
+    Updates pkgin repositories.
+    '''
+
+    yield 'pkgin update'
+
+_update = update  # noqa: E305
+
+
+@operation
+def packages(
+    packages=None, present=True, latest=False,
+    update=False, upgrade=False,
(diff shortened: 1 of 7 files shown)


pyinfra/operations/pkgin.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+    '''
+
+    if update:
+        yield _update(state=state, host=host)
+
+    if upgrade:
+        yield _upgrade(state=state, host=host)
+
+    # TODO support glob for specific versions (it isn't as simple
```
Fizzadar, line 73: This is interesting - would require fairly major refactoring of `version_join` in the [`ensure_packages`](https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/operations/util/packaging.py#L29https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/operations/util/packaging.py#L29) function. Particularly a way to confirm if a current version of a package falls within the specified limits - potentially can utilise Python's own packaging version handling for that, if they align.

Fizzadar (approved):
This is fantastic, thank you for putting this together @lun-4!

Issue #656, opened 2021-09-13
Document operation + fact test JSON spec


Fizzadar:

Fizzadar:
I think the existing set is documentation enough, many PRs since this issue with great tests written by contributors so feels redundant.

PR #660, opened 2021-09-19
Add 'available' as an argument to apk.upgrade

lun-4:
From the [Upgrading Alpine](https://wiki.alpinelinux.org/wiki/Upgrading_Alpine) wiki page:

> The --available switch is used to force all packages to be upgraded, even if they have the same version numbers. Sometimes changes in uClibc require doing this.

Thought it would be a good option to add. I'm unsure on how to write a JSON test for it, though!

### DIFF
--- a/pyinfra/operations/apk.py
+++ b/pyinfra/operations/apk.py
@@ -12,6 +12,8 @@ from .util.packaging import ensure_packages
 def upgrade(available=False, state=None, host=None):
     '''
     Upgrades all apk packages.
+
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Thank you for this addition @lun-4! I pushed a quick test in https://github.com/Fizzadar/pyinfra/pull/660/commits/77f8686ccc8ed71e8db9bed182605fd56410613f - currently the process of making these is totally undocumented (see: https://github.com/Fizzadar/pyinfra/issues/656).

Issue #663, opened 2021-09-22
`file.put` fails if `dest` is a directory


oz123:
## Describe the bug
Using `files.put` as:

```
files.put(
   src="/home/oz123/some.deb"
   dest='/home/oz123/'
   )
```

Results in the following error:
```
   [hostxxx] Command socket/SSH error: OSError('Failure',)
```

Which is pretty cryptic...

## To Reproduce
Use as above.

## Expected behavior
One can use `files.put` as  with regular `sftp` or other unix commands `mv`, `cp`, which allow to ommit file name if the path is directory.

## Meta
+ Include output of `pyinfra --support`.
```
oz123@server:~/nesuss-agents$ ./venv/bin/pyinfra --support
--> Support information:

    If you are having issues with pyinfra or wish to make feature requests, please
    check out the GitHub issues at https://github.com/Fizzadar/pyinfra/issues .
    When adding an issue, be sure to include the following:

    System: Linux
      Platform: Linux-4.4.0-18362-Microsoft-x86_64-with-glibc2.29
      Release: 4.4.0-18362-Microsoft
      Machine: x86_64
    pyinfra: v1.4.14
    Executable: ./venv/bin/pyinfra
    Python: 3.8.10 (CPython, GCC 9.4.0)
oz123@server:~/nesuss-agents$

```
+ How was pyinfra installed (source/pip)?
```pip3 install pyinfra```

The issue is actually not with pyinfra, rather with paramiko (I have 2.7.2) see here https://github.com/paramiko/paramiko/issues/1000.

Fizzadar:
Thank you for fixing this @oz123; this is now live in `v1.4.15`.

PR #664, opened 2021-09-22
Fix: don't fail if dest in files.put is a directory

oz123:
See #663.

Thanks for publishing this project. It makes me happy when I think I don't have to use ansible anymore!
This is one possible approach. Another would be to specify the issue in the documentation.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -706,6 +706,9 @@ def put(
     mode = ensure_mode_int(mode)
     remote_file = host.get_fact(File, path=dest)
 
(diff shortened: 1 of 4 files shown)


Fizzadar (approved):
Thank you for finding & fixing this @oz123! I pushed a few extra bits to document it and handle Windows systems path joining. Will release this in `v1.4.15` now!

oz123:
@Fizzadar thank you for attending to my incomplete PR. I lack the time to do a better job :-)

Fizzadar:
No worries at all, happy to help; thank you for taking the time to contribute to pyinfra :)

PR #665, opened 2021-09-23
Fix use of get_sudo_password() with less arguments

themanifold:
`get_sudo_password()` seems have been refactored, but the modules that call it don't seem to have been refactored.

This small change fixes that.


Fizzadar (approved):
Thank you for spotting and fixing this @themanifold! Reminds me that this codepath is currently untested.

PR #674, opened 2021-10-05
Fix files.sync exclude dir multiple

gchazot:
This contains some fixing and clean-up of the `files.sync` operation, with most importantly:
* c43c1df + 278d1a8 -> Fixing handling of list in `exclude_dir` argument
* c41490d and 2b90e80 are 2 distinct solutions to handle excluding "sub-directories"

Discussion is needed around the solutions for excluding "sub-directories". To be more explicit, the question is:

What behaviour do we want when the deploy specifies `exclude_dir="__pycache__"` (for example)? Do we want this to be enough to exclude any directory with this name in the `src` tree, or only the one possibly present right under `src` directory? 

Note that a very similar question should apply to the `exclude=...` argument, in order to keep both arguments behaviour consistent.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -464,8 +464,8 @@ def sync(
         remote_dirpath = dirpath.replace(src, '')
 
         # Filter excluded dirs
-        for child_dir in dirnames:
-            child_path = os_path.join(remote_dirpath, child_dir)
+        for child_dir in dirnames[:]:
+            child_path = os_path.normpath(os_path.join(remote_dirpath, child_dir))
             if exclude_dir and any(fnmatch(child_path, match) for match in exclude_dir):
                 dirnames.remove(child_dir)
 
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -433,6 +433,15 @@ def sync(
             src='files/tempdir',
             dest='/tmp/tempdir',
         )
+
+    Note: ``exclude`` and ``exclude_dir`` use ``fnmatch`` behind the scenes to do the filtering.
+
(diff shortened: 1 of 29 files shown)


gchazot:
Starting the discussion about the `exclude_dir` behaviour, I think that the current behaviour is already the better one, i.e. the one presented with c41490d80769cb62506c6728321af8f9a3113684. It is a little annoying because it is repeating the argument (`__pycache__` + `*/__pycache__`), but this solution allows to say explicitely with the wildcard that the folder name should be excluded in all subfolders.

The repetition is inherent to the use of `fnmatch` as the pattern handling function and because it doesn't treat `/` as a special character, meaning `*/` does not match "no folder" (i.e. `src`)


The same argument goes against the other solution (2b90e80ed5c3dbf76a4e6fc582d38da4a72a93dd), since there, specifying `__pycache__` is the only way to specify that `src/__pycache__` must be excluded, but it also means that *any* other `__pycache__` folder nested below `src` will also be implicitly excluded

Fizzadar:
I'm a strong agree on the explicit first method. Annoying to specify twice but means you know exactly what the operation is going to do. This also matches the current logic thus avoiding a breaking change.

I think putting an example of the double exclude in the docstring will suffice to ensure this is well documented.

gchazot:
I've rebased this on top of `gchazot:rework_filesystem_mock` (#675) to benefit from the new filesystem mock format.

Removed solution 2 commits

Added note to the documentation to explain the subtlety of `exclude_dir`:
image

gchazot:
This is ready I think, just depends on #675 being merged first

Fizzadar (approved):
Thank you @gchazot!

PR #675, opened 2021-10-05
Rework filesystem mock

gchazot:
Suggestion of a change in tests description of local filesystem contents to normalise the structure. This allows to remove redundancy of data, and will allow simplification of the code handling filesystem mocking.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -236,11 +236,6 @@ class patch_files(object):
                 patch_directories,
             )
         else:
-            print(self._parse_files_and_directories(
-                patch_files,
-                patch_directories,
-            ))
-            print(self._parse_local_files(local_files))
             directories, files, files_data = self._parse_local_files(local_files)
 
         self._files = files
--- a/tests/util.py
+++ b/tests/util.py
@@ -242,6 +242,8 @@ class patch_files(object):
         files_data = {}
         directories = {}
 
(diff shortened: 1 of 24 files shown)


tests/util.py
```diff
@@ -229,11 +229,19 @@ def __exit__(self, *args, **kwargs):
...
-            patch_directories,
-        )
+    def __init__(self, patch_files, patch_directories, local_files):
+        if local_files is None:
+            directories, files, files_data = self._parse_files_and_directories(
+                patch_files,
+                patch_directories,
+            )
+        else:
+            print(self._parse_files_and_directories(
```
Fizzadar: Are these leftovers or to be used when debugging tests (if so can we add a quick comment indicating)?

Fizzadar (commented):
BIG fan of this - a much cleaner way to specify the dirs/files and much less error prone 👍

tests/util.py
gchazot: Oops, these are leftovers indeed 59fb3e3

gchazot:
Cool, when I get the time I'll finalise the migration of all tests and remove the old format

gchazot:
I've cleaned this up, it's ready now

Fizzadar (approved):
Thank you for putting this together @gchazot!

PR #680, opened 2021-10-20
Use uname -m instead of -p for fact server.Arch

weakish:
since uname -m is more portable than uname -p.


Fizzadar (approved):
Hi @weakish - sorry this took a while to get to - excellent change thank you for pushing it 👍. Going to merge right away and update the test as the file has moved between your branch and HEAD.

PR #697, opened 2021-11-04
Using `group=group or user` instead of `group=user` in operations.server.user

glassbeads:
Here's a simple (probably even naïve) fix to this particular issue.

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -936,7 +936,7 @@ def user(
     if ensure_home:
         yield files.directory(
             home,
-            user=user, group=user,
+            user=user, group=group or user,
             state=state, host=host,
         )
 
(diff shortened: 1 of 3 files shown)


Fizzadar (commented):
Thank you for the PR @glassbeads. I think this is a reasonable solution to the problem. The [`add_user.json`](https://github.com/glassbeads/pyinfra/blob/current/tests/operations/server.user/add.json#L22) test will need updating to match the change.

Another option would be set group to match the group passed into the function, but as by default (most?) OS's will create a user with the same group name by default, this seems reasonable and non-breaking.

glassbeads:
I've updated the add_user.json test accordingly. Thank you for pointing it out!

Fizzadar (approved):
Thank you for updating that @glassbeads! I made a mistake and missed the other edit user test; pushed a fix in https://github.com/Fizzadar/pyinfra/pull/697/commits/a2f0bc6293f65d9c58c15e8755657642852201fb :)

PR #699, opened 2021-11-04
Unconditionally add os.getcwd() into sys.path

glassbeads:
This commit contains a quick fix for relative Python imports from the top-level project directory.
It simply unconditionally adds os.getcwd() into sys.path right at the beginning, while still allowing deploy_dir to override things later, if necessary.

This commit doesn't fix relative paths for files, like templates, locally referenced deploys/tasks etc. Looks like this functionality needs to be completely revisited, along with merging data from multiple levels of group_data (both top-level and inventory-level).

### DIFF
--- a/pyinfra_cli/main.py
+++ b/pyinfra_cli/main.py
@@ -289,6 +289,9 @@ def _main(
     init_virtualenv()
 
     cwd = getcwd()
+    # Unconditionally adding cwd into sys.path
+    sys.path.append(cwd)
+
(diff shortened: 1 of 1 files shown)


glassbeads:
Fixed lint complaints. I can't believe I pushed this, honestly, but here we are.

Fizzadar (approved):
Thanks for adding this @glassbeads, suitable workaround for the current issue pending further discussion 👍

PR #701, opened 2021-11-04
Using `grep '^{0}:'` instead of `grep '{0}:'` for group/user lookup

glassbeads:
This commit contains a fix for a regular expression that is used in group/user lookup in group/user create operations.
NB: while this will work fine in the vast majority of cases, the particular applicability of `-E` flag might need to be discussed.

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -770,7 +770,7 @@ def group(group, present=True, system=False, gid=None, state=None, host=None):
 
         # Groups are often added by other operations (package installs), so check
         # for the group at runtime before adding.
-        yield "grep -E '^{0}:' /etc/group || groupadd {1}".format(
+        yield "grep '^{0}:' /etc/group || groupadd {1}".format(
             group,
             ' '.join(args),
         )
@@ -883,7 +883,7 @@ def user(
 
         # Users are often added by other operations (package installs), so check
         # for the user at runtime before adding.
-        yield "grep -E '^{1}:' /etc/passwd || useradd {0} {1}".format(
+        yield "grep '^{1}:' /etc/passwd || useradd {0} {1}".format(
             ' '.join(args),
             user,
         )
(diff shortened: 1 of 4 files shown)


pyinfra/operations/server.py
```diff
@@ -770,7 +770,7 @@ def group(group, present=True, system=False, gid=None, state=None, host=None):
 
         # Groups are often added by other operations (package installs), so check
         # for the group at runtime before adding.
-        yield "grep '{0}:' /etc/group || groupadd {1}".format(
+        yield "grep -E '^{0}:' /etc/group || groupadd {1}".format(
```
Fizzadar:
```suggestion
        yield "grep '^{0}:' /etc/group || groupadd {1}".format(
```

I believe this should work OK, as the start/end characters work with non-extended-regex grep.

pyinfra/operations/server.py
```diff
@@ -883,7 +883,7 @@ def user(
 
         # Users are often added by other operations (package installs), so check
         # for the user at runtime before adding.
-        yield "grep '{1}:' /etc/passwd || useradd {0} {1}".format(
+        yield "grep -E '^{1}:' /etc/passwd || useradd {0} {1}".format(
```
Fizzadar:
```suggestion
        yield "grep '^{1}:' /etc/passwd || useradd {0} {1}".format(
```

As above ^

Fizzadar (changes requested):
Nice catch @glassbeads, thank you for putting this together!

One small suggestion + tests need updating which live in https://github.com/glassbeads/pyinfra/tree/issues/700/tests/operations/server.user - should just be a case of updating the command output within those tests.

pyinfra/operations/server.py
glassbeads: Agreed. Done

pyinfra/operations/server.py
glassbeads: This one too. And tests are fixed now, thanks for pointing me to them!

Fizzadar (approved):
Thank you for pushing those fixes, @glassbeads!

PR #713, opened 2021-11-30
Use best practice to check empty output

sahmed007:
It is a Python best practice to use 'if not val' to check if container is empty.


Fizzadar (approved):
Thank you @sahmed007! Nice spot, much cleaner 👍

PR #718, opened 2021-12-09
apt: add dist_upgrade operation

bauen1:
Sometimes, just `apt-get upgrade` is not enough, e.g. when upgrading the kernel.

Tested that it does what it says:

```
Dec 09 13:33:29 de-fra0 sudo[15904]: sysadmin : PWD=/home/sysadmin ; USER=root ; COMMAND=/usr/bin/sh -c DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" dist-upgrade
```

### DIFF
--- a/pyinfra/operations/apt.py
+++ b/pyinfra/operations/apt.py
@@ -354,6 +354,23 @@ _upgrade = upgrade  # noqa: E305 (for use below where update is a kwarg)
 
 
 @operation
+def dist_upgrade(state, host):
+    '''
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thank you for adding this @bauen1!

(I've switched to the `current` branch as the base as I'm planning to move to that in almost every case for simplicity)

Issue #692, opened 2021-10-30
Add support for non-POSIX target environments in the SSH connector


KuxaBeast:
## Description
I tried using pyinfra to write a few operations to deploy my infrastructure consisting of MikroTik RouterBoard-based devices running RouterOS, but found out, that the SSH connector has a hardcoded `sh -c` wrapper function in its code for executing remote commands.

After some investigation I found the culprit is this: (api/connectors/ssh.py:282)
```command = make_unix_command(command, state=state, **command_kwargs)```

I don't know if I am missing something and there is a way to execute remote commands without this wrapper (which can only support devices running POSIX shell), but it would be nice to take into consideration adding support for non-POSIX target environments or at least to have the option to turn this feature off.

## To Reproduce
Try to execute a valid command over SSH on virtually any target which doesn't ship a POSIX-compliant shell.

## Expected behavior
The command executes without any issue on the target device.

## Meta
+ `pyinfra --support`
```
System: Linux
      Platform: Linux-5.14.14-arch1-1-x86_64-with-glibc2.33
      Release: 5.14.14-arch1-1
      Machine: x86_64
    pyinfra: v1.4.17
    Executable: /home/user/.local/bin/pyinfra
    Python: 3.9.7 (CPython, GCC 11.1.0)
```
+ How was pyinfra installed (source/pip)?
pipx, since that's the only way that worked for me on Arch Linux
+ Output with `-vv` and `--debug`
```
$ pyinfra -vv --debug yum.ospf exec -- '/file print'
    [pyinfra_cli.main] Deploy directory remains as cwd
--> Loading config...
--> Loading inventory...
    [pyinfra_cli.inventory] Creating fake inventory...

--> Connecting to hosts...
    [pyinfra.api.connectors.ssh] Connecting to: yum.ospf ({'allow_agent': True, 'look_for_keys': True, 'hostname': 'yum.ospf', 'timeout': 10})
    [yum.ospf] Connected
    [pyinfra.api.state] Activating host: yum.ospf
    [pyinfra.api.operation] Adding operation, {'Server/Shell'}, opOrder=(0,), opHash=784a97bf1955d5f7a2b9dd6c1e371e17b73c42bc

--> Proposed changes:
    Ungrouped:
    [yum.ospf]   Operations: 1   Commands: 1   

--> Beginning operation run...
--> Starting operation: Server/Shell (/file print)
    [pyinfra.api.operations] Starting operation Server/Shell on yum.ospf
[...]

PR #724, opened 2021-12-23
Make `shell_executable=None` pass raw commands but default to `sh -c` wrap

KuxaBeast:
Allow setting `shell_executable` in ssh connector to `None` to pass raw commands instead of wrapping it as a payload to the `shell_executable`. Default to `sh` as default shell for backward compatibility. Solves [#692](https://github.com/Fizzadar/pyinfra/issues/692).


KuxaBeast:
Okay, I implemented the base but it seems like it defaults to `shell_executable=None` instead of `sh`, which I set in the function arguments as default, and I can't find where it is actually overridden.

Fizzadar (approved):
Thank you for pulling this together @KuxaBeast!

KuxaBeast:
Thanks, @Fizzadar. I think it still requires some polishing though. There is still the issue of the default behaviour. As I previously stated, I don't have a clue where is it overridden, because the default value in the function definition didn't seem to do anything.

pyinfra/api/connectors/util.py
```diff
@@ -328,12 +325,12 @@ def make_unix_command(
 
         if shell_executable is not None:
             # Quote the whole shell -c 'command' as BSD `su` does not have a shell option
-            command_bits.append(QuoteString(StringCommand(shell_executable, '-c', command)))
+            command_bits.append(QuoteString(StringCommand(shell_executable, '-c', QuoteString(command))))
         else:
             command_bits.append(StringCommand(command))
```
KuxaBeast: I might have forgotten `QuoteString()` around the `StringCommand()` here.

Fizzadar:
@KuxaBeast my apologies - I jumped the gun a bit here! So the default value is populated here:

https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/api/operation_kwargs.py#L69

Which itself is retrieved from the config, with the default `None` here:

https://github.com/Fizzadar/pyinfra/blob/current/pyinfra/api/config.py#L55

I think we should be safe to change the `Config.SHELL` default and have this behave as expected.

---

For full context any of the "global arguments", such as `shell_executable`, can be passed directly into operation/deploy calls, defined in the config and also as host data. The config value is the default and in most cases this is never used.

KuxaBeast:
Okay, I think I found another problem. When I changed the default of `Config.SHELL` variable to `sh` another problem appeared. Until now, I have been verifying the behaviour using the `exec` operation and the `--shell-executable` argument, but it seems it accepts any value as string, resulting in `None -c command` or `null -c command`.

Calling: `$ pyinfra -vv --debug yum.ospf exec --shell-executable=None -- '/file print'`
results in:
```
[yum.ospf] >>> None -c '/file print'
[yum.ospf] expected command name (line 1 column 1)
```

I am not sure why didn't I notice it before, but the `server.shell` operation does not return the output of executed commands for me, but does indeed support setting the `None` value by passing `shell_executable=null` as the second argument. (`--shell-executable=null` doesn't work with the `exec` op.)

Successful, but does not print command output:
`$ pyinfra -vv --debug yum.ospf server.shell '/file print' shell_executable=null`

pyinfra/api/connectors/util.py
KuxaBeast: Yup. Totally forgot. Will fix in next commit.

Fizzadar:
Using `-vvv` with the second version will print the output.

Exec is a bit of a special case, under the hood it actually calls `server.shell` and sets the print output flags. I think it's acceptable to parse `'None'` into `None` when handling the CLI arg (https://github.com/Fizzadar/pyinfra/blob/current/pyinfra_cli/main.py#L461).

This works with `server.shell` because the arguments are parsed as JSON such that `'null'` becomes `None`, etc.

KuxaBeast:
Oh, I see now. It might be a good idea to mention the JSON-parsing in the docs somewhere. Anyway, I am gonna explicitly convert `'None'` and `'null'` arguments to `None` in the pyinfra_cli/main.py for now. Maybe this would deserve a whole separate cli flag like `--no-shell` instead?

pyinfra/api/connectors/util.py
```diff
@@ -240,7 +240,7 @@ def make_unix_command(
     command,
     env=None,
     chdir=None,
-    shell_executable='sh',
+    shell_executable=None,
```
Fizzadar: I think this should remain `'sh'` for now as the default and `None`/no shell has to be explicitly set.

Fizzadar (commented):
One suggestion - everytihng else looks good!

On the CLI flag I think it's best to keep one arg for the purpose, converting from the string values seems acceptable to me (I think, open to thoughts!). Agreed this needs to be better documented!

pyinfra/api/connectors/util.py
KuxaBeast: Okay. Fixing.

Fizzadar (approved):
Thank you once again for pulling this together @KuxaBeast, really exciting to start opening up pyinfra to different target types!

EDIT: I pushed a tiny fix for some tests checking for the call to `make_unix_command` so I can get this into `v1.6`!

vojta001:
I wish I had known this earlier. The docs aren't really detailed in this regard.

Anyway, is there a way to set this option in the inventory? `{"_shell_executable": None}` is of no luck for me.

Fizzadar:
> I wish I had known this earlier. The docs are really detailed in this regard.
> 
> Anyway, is there a way to set this option in the inventory? `{"_shell_executable": None}` is of no luck for me.

Hrm that should work inside the inventory, mind filing a bug report for that?

vojta001:
@Fizzadar Sure! Thank you for your dedication

Issue #737, opened 2022-01-11
SSH Hostkey checking security issues. (Behaviour deviates from SSH)


bauen1:
## Describe the bug

The configuration option `StrictHostKeyChecking` is ignored, and because of https://github.com/Fizzadar/pyinfra/commit/be6b97c4f6f443227a3fa170d5622fae51e9331b pyinfra will **blindly accept any host key**, allowing man in the middle attacks if the host key is not manually added.

Furthermore when calling `rsync` host key checking is **completely** disabled.

## To Reproduce

- Run pyinfra.

## Expected behavior

Respect `StrictHostKeyChecking`, and most importantly adopt the same default as would be expected from ssh, from ssh_config.5:

> If this flag is set to ask (the default), new host keys will be added to the user known host files only after the user has confirmed that is what they really want to do, and ssh will refuse to connect to hosts whose host key has changed. 

## Meta

+ Include output of `pyinfra --support`.
```
    System: Linux
      Platform: Linux-5.15.0-2-amd64-x86_64-with-glibc2.32
      Release: 5.15.0-2-amd64
      Machine: x86_64
    pyinfra: v1.5
    Executable: /usr/local/bin/pyinfra
    Python: 3.9.9 (CPython, GCC 11.2.0)
```

+ How was pyinfra installed (source/pip)?
Source, basically v1.5 ; in some weird container setup.

I have a patch in the working.

PR #741, opened 2022-01-16
ssh: don't disable host key checking when performing rsync

bauen1:
This will make ssh behave as configured when invoking rsync.

Fixes parts of #737 

Tested.


Fizzadar (approved):
Thank you for pulling this together @bauen1, I want to wait for https://github.com/Fizzadar/pyinfra/pull/742 before merging this (will merge together) as without it may cause even more confusing behaviour!

PR #742, opened 2022-01-16
Draft: ssh connector: respect StrictHostKeyChecking

### LINKED ISSUE #737: SSH Hostkey checking security issues. (Behaviour deviates from SSH)

bauen1:
Fixes #737 

In theory this works and has been sort of tested, but I still need to:

- [x] Figure out what unit tests this breaks
- [x] Test that the strict policy actually does what it says
- [x] ~~If necessary split the commit for userknownhostsfile into a seperate PR~~

### DIFF
--- a/pyinfra/api/connectors/sshuserclient/client.py
+++ b/pyinfra/api/connectors/sshuserclient/client.py
@@ -71,13 +71,22 @@ class SSHClient(ParamikoClient):
         _pyinfra_ssh_config_file=None,
         **kwargs
     ):
-        hostname, config, forward_agent, missing_host_key_policy = self.parse_config(
-            hostname,
-            kwargs,
-            ssh_config_file=_pyinfra_ssh_config_file,
-        )
+        hostname, config, forward_agent, missing_host_key_policy, host_keys_file = \
+            self.parse_config(
+                hostname,
+                kwargs,
+                ssh_config_file=_pyinfra_ssh_config_file,
+            )
         self.set_missing_host_key_policy(missing_host_key_policy)
         config.update(kwargs)
+
+        try:
+            self.load_system_host_keys(filename=host_keys_file)
+        # Unfortunately paramiko bails for any dodge line in known hosts
+        # See: https://github.com/Fizzadar/pyinfra/issues/683
+        except Exception as e:
+            logger.warning('Failed to load known host keys: {0}'.format(e))
+
         super(SSHClient, self).connect(hostname, **config)
 
         if _pyinfra_force_forward_agent is not None:
@@ -124,6 +133,10 @@ class SSHClient(ParamikoClient):
                     'Invalid value StrictHostKeyChecking={}'.format(
                         host_config['stricthostkeychecking']))
 
+        host_keys_file = None
+        if 'userknownhostsfile' in host_config:
+            host_keys_file = path.expanduser(host_config['userknownhostsfile'])
+
         if 'hostname' in host_config:
             hostname = host_config['hostname']
 
@@ -159,7 +172,7 @@ class SSHClient(ParamikoClient):
                 sock = c.gateway(hostname, cfg['port'], target, target_config['port'])
             cfg['sock'] = sock
 
-        return hostname, cfg, forward_agent, missing_host_key_policy
+        return hostname, cfg, forward_agent, missing_host_key_policy, host_keys_file
 
     @staticmethod
     def derive_shorthand(host_string):
--- a/pyinfra/api/connectors/sshuserclient/client.py
+++ b/pyinfra/api/connectors/sshuserclient/client.py
@@ -5,7 +5,9 @@ source has now vanished (https://github.com/tobald/sshuserclient).
 
 from os import path
 
+from gevent.lock import BoundedSemaphore
 from paramiko import (
+    HostKeys,
     MissingHostKeyPolicy,
     ProxyCommand,
     SSHClient as ParamikoClient,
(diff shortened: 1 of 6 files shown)


themanifold:
I did a draft of something similar where I allowed the passing in of custom host keys, but great job.

pyinfra/api/connectors/sshuserclient/client.py
```diff
@@ -18,6 +19,23 @@
...
+class StrictPolicy(MissingHostKeyPolicy):
+    def missing_host_key(self, client, hostname, key):
+        logger.error('No host key for {0} found in known_hosts'.format(hostname))
+        raise SSHException("StrictPolicy: No host key for {0} found in known_hosts".format(hostname))
+
+
+class AskPolicy(MissingHostKeyPolicy):
+    def missing_host_key(self, client, hostname, key):
+        logger.error('No host key for {0} found in known_hosts'.format(hostname))
+        raise SSHException("AskPolicy: not implemented!\n")
```
Fizzadar: This can probably just call `getpass` here the same we do for sudo passwords:

https://github.com/Fizzadar/pyinfra/blob/ff0913d6a172966760b63fe59e55dff9ea852e0d/pyinfra/api/connectors/util.py#L176

pyinfra/api/connectors/sshuserclient/client.py
```diff
@@ -79,6 +106,25 @@ def parse_config(self, hostname, initial_cfg=None, ssh_config_file=None):
...
+            elif v == 'yes':
+                missing_host_key_policy = StrictPolicy()
+            else:
+                raise SSHException(f"Invalid value StrictHostKeyChecking={host_config['stricthostkeychecking']}")
+
+        host_keys_file = None
+        if 'userknownhostsfile' in host_config:
+            host_keys_file = host_config['userknownhostsfile']
+            if host_keys_file.startswith('~'):
+                host_keys_file = getenv('HOME') + host_keys_file[1:]
```
Fizzadar:
```suggestion
            host_keys_file = os.expanduser(host_keys_file)
```

[`os.expanduser`](https://docs.python.org/3/library/os.path.html#os.path.expanduser) is a neater way to do this.

Fizzadar (changes requested):
@bauen1 this is really awesome, a huge improvement to `pyinfra` and very much appreciated.

Left a couple of notes - let me know if there's anything you need a hand with or if you'd like me to sort out the tests.

bauen1:
@themanifold if you want to make your own PR with that It would be quite nice, as my implementation of `UserKnownHostsFile` is incomplete and was only hacked together so pyinfra would work correctly with my setup.

pyinfra/api/connectors/sshuserclient/client.py
bauen1: This methode is called if paramiko does not find the ssh host key in the users `known_hosts` file, I'm not sure what exactly you mean.

Fizzadar: Sorry I wasn't clear - `getpass` is also the wrong function I meant `raw_input`!

```py
class AskPolicy(MissingHostKeyPolicy):
    def missing_host_key(self, client, hostname, key):
        should_continue = raw_input('No host key for {0} found in known_hosts, continue? [y/N] '.format(hostname))
        if should_continue.lower() == 'y':
                ...add key to known hosts
                return
        raise SSHException('Failed due to no valid host key for {0}'.format(hostname))
```

bauen1: I still can't find that function

pyinfra/api/connectors/sshuserclient/client.py
bauen1: Done.

pyinfra/api/connectors/sshuserclient/client.py
bauen1: Changed, the code for adding the host key was taken from paramiko itself, but I haven't tested that it actually works.

bauen1:
I've tested this, but only the StrictHostKeyPolicy, there is also some potential for additional tests.
Apart from my lack of time to test this thoroughly right now this should be good now.

Fizzadar (approved):
Updates all look good to me @bauen1, thank you for updating this! When I get a minute I'll write a quick test or two for the other policies so we can get this merged in!

Fizzadar:
Confirmed things working as expected, pushed some unrelated fixes to get the tests working. Thank you once again @bauen1!

PR #743, opened 2022-01-23
pyinfra v2

Fizzadar:
Tracking the upcoming v2 release, which removes Python 2 support and anything that currently shows a warning. Will also bring in a number of new backwards compatible changes/warnings that will be removed in v3.


Fizzadar:
Closing/merged, now `2.x` branch which is the default.

PR #745, opened 2022-01-26
Fix missing _bash for creating the completion

oz123:
Somehow this got dropped. Without this click does not
do the completion for bash.


Fizzadar (approved):
Thanks for catching & fixing this @oz123!

PR #755, opened 2022-02-18
Passing the success_exit_codes to facts

mohsinhijazee:
Currently, the `success_exit_codes` are not passed down to the facts hence any fact that depends on a command whose only output matters or it has non standard exit codes, cannot be used for fact gathering. This PR fixes that.


pyinfra/api/facts.py
```diff
@@ -196,6 +197,7 @@ def get_facts(
         ignore_errors = current_global_kwargs.get('ignore_errors', ignore_errors)
         timeout = current_global_kwargs.get('timeout', timeout)
         env = current_global_kwargs.get('env', env)
+        success_exit_codes = current_global_kwargs.get('success_exit_codes', [0])
```
Fizzadar:
```suggestion
        success_exit_codes = current_global_kwargs.get('success_exit_codes', success_exit_codes)
```

Fizzadar (approved):
Thank you for implementing this @mohsinhijazee! One minor tweak but otherwise looks good to me.

pyinfra/api/facts.py
mohsinhijazee: Okay, makes sense actually. Will fix.

mohsinhijazee:
> Thank you for implementing this @mohsinhijazee! One minor tweak but otherwise looks good to me.

Most welcome! I am going to add the tests that would bump the coverage back up. :)

tests/test_api/test_api_facts.py
```diff
@@ -143,6 +146,39 @@ def test_get_fact_error_ignore(self):
             print_input=False,
             print_output=False,
             shell_executable='sh',
+            success_exit_codes=[0],
```
mohsinhijazee, line 149: Diff is messed up for some reason. I've however shamelessly copied the boilerplate from the existing test case.

pyinfra/api/facts.py
mohsinhijazee: Fixed.

mohsinhijazee:
@Fizzadar GTG from my side subject to your approval of course! There's another issue related to the facts that only stdout is returned and I am not sure if we should touch it or not, I'll create an issue for that discussion separately.

Fizzadar (approved):
Thank you @mohsinhijazee! Looks good to me, will get into `v1.6.3` I'm about to release.

PR #768, opened 2022-03-03
Pyinstaller binaries

Fizzadar:
Sometimes it's desirable to have a single binary instead of a `pip`/whatever package. Unfortunately in the Python world this isn't a simple task and the resulting binary is pretty slow to execute (10s to start cold). But it works!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,13 +11,13 @@ jobs:
       matrix:
         include:
           - os: ubuntu-18.04
-            python-version: 3.9
+            python-version: "3.10"
             filename: pyinfra
           - os: macos-10.15
-            python-version: 3.9
+            python-version: "3.10"
             filename: pyinfra
           - os: windows-2019
-            python-version: 3.9
+            python-version: "3.10"
             filename: pyinfra.exe
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


v3ss0n:
should be doable with `nuitka`

Fizzadar:
For now I'm killing this, I do wish to revisit probably using `nuitka` but the performance is still too much of a concern, although I have some ideas on dealing with this (bundle the pyinfra module into a zip to avoid all the IO operations on startup). Anyway, will revisit in 3.X where X > 0.

paidhi:
If you revisit this I would recommend taking a look at what [PyApp](https://ofek.dev/pyapp), by Ofek Lev, has to offer. [Hatch](https://hatch.pypa.io) (by the same author) has a simple integration with PyApp via it's [Binary Builder Plugin](https://hatch.pypa.io/latest/plugins/builder/binary/).

I use this at work to build an internal utility that is written in Python and has mulitple 3rd-party dependencies (e.g. rich, requests, psutils). It produces a single (Linux ELF) executable that I can just copy an any machine and run.

A good usage example might be Hatch itself. One can download it as a single executable. PyApp provides the executable with a nice "self" subcommand. E.g.: `hatch self update`.

PR #770, opened 2022-03-04
Enhance systemd operation and facts with full support for `--user` and `--machine` flags

jmpolom:
This PR completes support for interacting with systemd user sessions, to include those of users other than the calling user by way of the [`--machine`][0] flag.

Previously, the systemd facts did not properly check the status of systemd user units despite the presence of the `user_mode` kwarg in the systemd.daemon_reload and systemd.service operations. This led to inconsistent behavior when interacting with user units as opposed to system units. The systemd facts now also accept an argument to indicate user mode is desired and this is incorporated into the systemd operations.

A new feature introduced is the ability to specify both a `machine` and `user_name` to the systemd facts and operations which adds a constructed `--machine=user_name@machine` flag to the systemctl command. This tells systemctl to access the bus of the specified user on the specified machine. This is targeted towards use with systemd-nspawn but most critically can be used to manage service units running as unprivileged systemd user services on the same host when `.host` is specified as the machine.

Possible areas for improvement: tests could/should be enhanced to cover the user mode and machine use cases

[0]: https://www.freedesktop.org/software/systemd/man/systemctl.html#-M

cc: @storrgie @jkl92

### DIFF
--- a/pyinfra/operations/systemd.py
+++ b/pyinfra/operations/systemd.py
@@ -5,25 +5,9 @@ Manage systemd services.
 from __future__ import unicode_literals
 
 from pyinfra.api import operation
-from pyinfra.facts.systemd import SystemdEnabled, SystemdStatus
+from pyinfra.facts.systemd import _make_systemctl_cmd, SystemdEnabled, SystemdStatus
 
 from .util.service import handle_service_control
-
-
-def _make_systemctl_cmd(user_mode=False, machine=None, user_name=None):
-    # base command for normal and user mode
-    systemctl_cmd = 'systemctl --user' if user_mode else 'systemctl'
-
-    # add user and machine flag if given in args
-    if machine is not None:
-        if user_name is not None:
-            machine_opt = '--machine={1}@{0}'.format(machine, user_name)
-        else:
-            machine_opt = '--machine={0}'.format(machine)
-
-        systemctl_cmd = '{0} {1}'.format(systemctl_cmd, machine_opt)
-
-    return systemctl_cmd
 
 
 @operation
--- a/pyinfra/operations/systemd.py
+++ b/pyinfra/operations/systemd.py
@@ -10,6 +10,22 @@ from pyinfra.facts.systemd import SystemdEnabled, SystemdStatus
 from .util.service import handle_service_control
 
 
+def _make_systemctl_cmd(user_mode=False, machine=None, user_name=None):
+    # base command for normal and user mode
+    systemctl_cmd = 'systemctl --user' if user_mode else 'systemctl'
+
+    # add user and machine flag if given in args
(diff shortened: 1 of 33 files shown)


pyinfra/operations/systemd.py
```diff
@@ -10,17 +10,43 @@
 from .util.service import handle_service_control
 
 
+def _make_systemctl_cmd(user_mode=False, machine=None, user_name=None):
```
Fizzadar: To avoid duplicating these functions this should be imported from the fact module alongside the facts.

Fizzadar (changes requested):
This is an excellent addition @jmpolom, thank you for pulling this together! Just one thing that should be tidied up and some tests would be good but can be added later if you're time pressured!

pyinfra/operations/systemd.py
jmpolom: Totally agreed. Was a little unhappy I had to send this with the same thing defined in two places.

I attempted the other way around, with a systemd.py in the operations/util directory which I then tried to import into facts/systemd.py; this did not work. ~~No syntax error so valid python but many tests failed which I suspect has to do with how things get loaded internally in pyinfra and I'm lacking specific knowledge about that.~~ I think I may have attempted to run the tests from the wrong directory previously? Unsure.

I will try it as you suggest.

jmpolom: Updated in [e64cd51](https://github.com/Fizzadar/pyinfra/pull/770/commits/e64cd51ab42bf7481567aa1cbf3593aa076101f0)

Fizzadar: Interesting the utils approach should work as well, hard to know without seeing the error! I think this way also works, it's hard to say which is the best approach for these things really!

Side thought: maybe facts + operations should live alongside each other rather than divided into two modules.

jmpolom: The function is not used in any other operation or fact definitions besides systemd so I think the current approach is understandable.

> Side thought: maybe facts + operations should live alongside each other rather than divided into two modules.

At a minimum having a higher level location for shared utility functions would be nice since apparently there could be some shared logic. The way facts and operations work is quite distinct so I don't know if I'd treat them as equals exactly.

jmpolom (commented):
repeated conditional code block was pulled out into function in facts/systemd.py as requested.

jmpolom:
I have added enhanced tests for systemd facts and operations which now cover the user and machine options.

Fizzadar (approved):
Thank you for this addition and adding those tests @jmpolom! Planning to release this in `v1.7` tomorrow :)

jmpolom:
@Fizzadar thanks for the prompt feedback and quick merge!

PR #774, opened 2022-03-09
Escape exported variables

lun-4:
Environment variables with URL values that have query parameters e.g `https://example.com?a=b&c=d` are interpreted differently than intended, since the `&` turns the export command into a background command, and strips off `c=d` from the variable.

This was manually tested on my deployment scripts, but I'm relying on CI for the test suite this time due to issues trying to spin up a local environment.


Fizzadar (approved):
Thank you @lun-4, looks good to me!

PR #775, opened 2022-03-09
Bugfix fact SystemdEnabled

jmpolom:
The SystemdEnabled fact did not properly report the status of instantiated systemd units (IE: `kresd@1.service`) because it used `systemctl list-unit-files` to generate the list of units to check the status of. Unfortunately there is either a bug in systemd or an error in the documentation suggesting `list-unit-files` should list instantiated units, which it does not seem to do consistently.

The net effect was that sometimes operations to enable an instantiated service were not idempotent when they should have been. The changes in this PR resolve that behavioral inconsistency.

Now, `systemctl list-units` is used which provides a more complete list of units on the system and then `systemctl show` is used to obtain the unit file enabled status.

### DIFF
--- a/pyinfra/facts/systemd.py
+++ b/pyinfra/facts/systemd.py
@@ -92,7 +92,7 @@ class SystemdEnabled(FactBase):
         return (
             '{0} -al --plain --legend=false --state=loaded list-units | '
             'while read -r UNIT REST; do '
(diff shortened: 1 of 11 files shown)


Fizzadar (approved):
Thank you @jmpolom, great improvement!

Issue #772, opened 2022-03-06
Invert `ensure_packages` lower argument default


Fizzadar:
See: https://github.com/Fizzadar/pyinfra/issues/769#issuecomment-1059832256

Fizzadar:
Cancelling as this would actually be breaking, and https://github.com/Fizzadar/pyinfra/issues/779 is a cleaner solution.

PR #778, opened 2022-03-13
Invert ensure_packages lower argument default

sysadmin75:
Fix for https://github.com/Fizzadar/pyinfra/issues/772


Fizzadar:
@sysadmin75 thank you for putting this together - unfortunately I jumped the gun a bit on the original issue and this would be a breaking change; I think the right thing to do here is https://github.com/Fizzadar/pyinfra/issues/779 which entirely removes the lowercasing of package names for v2, leaving v1 as-is. My apologies.

PR #798, opened 2022-04-23
Fix: Remove state/host arguments from apt.dist_upgrade operation

pabloxio:
This PR fixes the following issue with `apt.dist_upgrade()` operation

```bash
TypeError: dist_upgrade() missing 2 required positional arguments: 'state' and 'host'
```


Fizzadar (approved):
Thank you for catching this and adding a test too!

PR #803, opened 2022-05-01
correct link to the license file

Lab-Brat:
I've noticed that when the "license|MIT" icon is pressed on README page, it returns 404 Error code. So I replaced the link to the license file with the one from the repository.


Fizzadar (approved):
Thanks for catching & fixing this @Lab-Brat!

PR #810, opened 2022-05-14
Fix examples link in docs

StefanBRas:
The old link was broken and lead to a 404 page.


Fizzadar (approved):
Thank you for catching & fixing this @StefanBRas!

phlummox:
Hi,

Just to let you know: I'm looking at https://docs.pyinfra.com/en/2.x/examples.html, and the link there is still broken - it looks like this change has not actually propagated to docs.pyinfra.com.

Issue #476, opened 2020-11-05
pylint error on decorated `name` argument


themanifold:
**pylint error on decorated `name` argument**
Due to the way the `name` argument is injected into some functions with a decorator, such as `pyinfra.operations.server.shell`, we get a pylint error like:

```
deploy/deploy.py:3:0: E1123: Unexpected keyword argument 'name' in function call (unexpected-keyword-arg)
```

**To Reproduce**
`deploy.py`:
```
from pyinfra.operations import server

server.shell(
    name='Run command',
    commands=['echo "1234"']
)
```
pylint deploy.py

**Expected behavior**
No pylint errors

**Meta**
+ `System: Linux
      Platform: Linux-5.8.11-1-MANJARO-x86_64-with-glibc2.2.5
      Release: 5.8.11-1-MANJARO
      Machine: x86_64
    pyinfra: v1.2.1
    Executable: /home/user/workspace/common/venv/bin/pyinfra
    Python: 3.8.5 (CPython, GCC 10.2.0)`
+ pip installed

themanifold:
Ah, turns out the support for decorators in pylint is a bit lacking. If you read https://github.com/PyCQA/pylint/issues/259 you'll see that there's a work around. I will test this now.

themanifold:
OK, so following that information, you do something like this in your `.pylintrc`:

```
[TYPECHECK]
signature-mutators=pyinfra.api.operation.operation
```
See this: https://github.com/PyCQA/pylint/blob/master/doc/whatsnew/2.4.rst

What unfortunately happens is that you will get false negatives when you actually pass in incorrect arguments into the pyinfra decorated functions.

Is flake8 fine with decorators?

I'll leave this up to you as to what you do with it.

Fizzadar:
So flake8 doesn't do type checking which has prevented this being an issue before! There's currently a PR/branch (https://github.com/Fizzadar/pyinfra/pull/472) tracking implementation of type checking using mypy.

The operation functions are particularly annoying as they all accept a list of global arguments, defined [here](https://github.com/Fizzadar/pyinfra/blob/master/pyinfra/api/operation_kwargs.py) which is a bit of a nightmare for typing. 

I think long run the right approach is to move these into the operation function definition itself, will need to deduplicate that with the deploy function also (possibly by way of a decorator decorator!).

wookayin:
[...]

PR #817, opened 2022-05-19
Type stub generator for operations.

StefanBRas:
As mentioned in [this issue](https://github.com/Fizzadar/pyinfra/issues/476#issuecomment-1100748320) a way to solve the typing issues on functions with the `@operation` decorator is to generate type stubs. 

This PR introduces that. I have not tested it anymore than it looks to be working and pyright isn't complaining:

image

When using operations, they will now not complain about the global args but still complain if any unknown arg is added:

image

The generator functions are in the file  `scripts/generate_operation_type_stubs.py` and running `python scripts/generate_operation_type_stubs.py` will generate a stub for each *.py file in `pyright/operations/`

This could (should) be extended to also have the correct annotations on each global arg, but I wanted to hear if it is of interrest at all before moving forward.

It introduces a dev dependency on the package "redbaron". It could probably be done with any other AST package or `ast` from standard lib. I only ended up on RedBaron because it seemed to be the most user friendly and documented one.


Fizzadar (commented):
This is fantastic @StefanBRas! Definitely want to get this included in `pyinfra`. Definitely agree it'd be good to add types for global args.

Fun fact: `pyinfra` used to use RedBaron to compile operation files to work around operation ordering with conditional statements (since fixed with stack call ordering): https://github.com/Fizzadar/pyinfra/commit/bd6c48ff8b7b8ed449435a7974ea0c8de67b7d3d#diff-fb2026268ae4b6a8ad26e8f37bf78bb7ff9a0af57ef13b2d571e3200cb02e5d5 - so absolutely OK with the extra dev dependency :)

StefanBRas:
Great! Right now I just get the argument names from `OPERATION_KWARGS` in [`arguments.py`](https://github.com/Fizzadar/pyinfra/blob/dfad6ef512781407187ef91434388e6d705f2d3c/pyinfra/api/arguments.py#L122), which doesn't have the types. Should I infer them from the default values or are they actually typed somewhere?

Apart adding those types and linting, is there anything else that would need to be added/changed? 

It could be added as a part of the build process, but I'm not sure how you would want to integrate it.

Also, maybe the generator should be included as a util function, since operation authors would probably also like to generate stubs for their own operations.

Fizzadar:
> Great! Right now I just get the argument names from `OPERATION_KWARGS` in [`arguments.py`](https://github.com/Fizzadar/pyinfra/blob/dfad6ef512781407187ef91434388e6d705f2d3c/pyinfra/api/arguments.py#L122), which doesn't have the types. Should I infer them from the default values or are they actually typed somewhere?

They're not currently typed anywhere, could add them to the definitions or just infer from the defaults for now, both seem reasonable to me. Perhaps look for a `"type"` field in the spec falling back to the default? ie explicit type would be needed here:

```py
"_env": {
    "description": "Dictionary of environment variables to set.",
    "handler": generate_env,
    "type": dict[str, str],
},
```

I'm also totally happy to do this part, up to you :)

> Apart adding those types and linting, is there anything else that would need to be added/changed?

Only thing would be making all the stub functions use one line per argument (just spotted some are on one super long line). Other than that this is good to go I think.

Long run I'm keen to add type annotations within the operations themselves, and I see this nicely handles that and includes them in the stubs which is ideal.

> It could be added as a part of the build process, but I'm not sure how you would want to integrate it.

Yeah I've been trying to figure out this part, I think first up will be adding a CI lint test to the GitHub workflow that checks the stubs are up to date which could just generate them and check if any files have changed. That should ensure the stubs are up to date similar to linting.

> Also, maybe the generator should be included as a util function, since operation authors would probably also like to generate stubs for their own operations.

Agreed, I've got a local folder in the repo `pyinfra_util/` that may be a suitable place for this (don't think it belongs within `pyinfra/` or `pyinfra_cli/`); perhaps a `pyinfra_util/stubgen.py` that could be used with `python -m pyinfra_util.stubgen`?

StefanBRas:
> They're not currently typed anywhere, could add them to the definitions or just infer from the defaults for now, both seem reasonable to me. Perhaps look for a `"type"` field in the spec falling back to the default? ie explicit type would be needed here:
> 
> ```python
> "_env": {
>     "description": "Dictionary of environment variables to set.",
>     "handler": generate_env,
>     "type": dict[str, str],
> },
> ```
> 
> I'm also totally happy to do this part, up to you :)

Ah, with "infer from the defaults" I actually ment manually infering them with my mind and then adding it. I will give it a try and then just add a comment if some of them are not clear to me.


> Only thing would be making all the stub functions use one line per argument (just spotted some are on one super long line). Other than that this is good to go I think.

Yeah, not sure why that happens actually. It seems almost random. I'll look into it.

> Agreed, I've got a local folder in the repo `pyinfra_util/` that may be a suitable place for this (don't think it belongs within `pyinfra/` or `pyinfra_cli/`); perhaps a `pyinfra_util/stubgen.py` that could be used with `python -m pyinfra_util.stubgen`?

Sounds good! I will let it take an optional argument so that you can point to other operations, right now it's hardcoded to point to the `operations.py` from `pyinfra`.

StefanBRas:
@Fizzadar I implemented the changes now. For some reason RedBaron would format it as one line when the operation decorator was used with parenthsis and multiple lines if not, I have no idea why. I just use black programmaticlly to fix the formatting, that seemed easier than investigating.

I make some `TODO` comments in `pyinfra/api/arguments.py` at the places where I'm not sure about the types. You should probably check all the types, but especially those.

pyinfra/api/arguments.py
```diff
@@ -42,14 +52,17 @@
...
     },
     "_doas": {
         "description": "Execute/apply any changes with ``doas``.",
         "defailt": lambda config: config.DOAS,
+        "type": bool,
     },
     "_doas_user": {
         "description": "Execute/apply any changes with ``doas`` as a non-root user.",
         "default": lambda config: config.DOAS_USER,
+        "type": bool,
```
Fizzadar:
```suggestion
        "type": str,
```

pyinfra/api/arguments.py
```diff
@@ -6,34 +8,42 @@
...
     },
     "_preserve_sudo_env": {
         "description": "Preserve the shell environment when using ``sudo``.",
         "default": lambda config: config.PRESERVE_SUDO_ENV,
+        "type": bool,
     },
     "_su_user": {
         "description": "Execute/apply any changes with ``su``.",
         "default": lambda config: config.SU_USER,
+        "type": bool,
```
Fizzadar:
```suggestion
        "type": str,
```

pyinfra/api/arguments.py
```diff
@@ -6,34 +8,42 @@
...
         "default": lambda config: config.USE_SUDO_PASSWORD,
+        "type": bool,
     },
     "_preserve_sudo_env": {
         "description": "Preserve the shell environment when using ``sudo``.",
         "default": lambda config: config.PRESERVE_SUDO_ENV,
+        "type": bool,
     },
     "_su_user": {
         "description": "Execute/apply any changes with ``su``.",
```
Fizzadar:
```suggestion
        "description": "Execute/apply any changes with this user using ``su``.",
```

pyinfra/api/arguments.py
```diff
@@ -71,51 +84,81 @@ def generate_env(config, value):
...
     },
     "_success_exit_codes": {
         "description": "List of exit codes to consider a success.",
         "default": lambda config: [0],
+        "type": Iterable[int],
+    },
+    "_timeout": {
+        "description": "Timeout for *each* command executed during the operation.",
+        "type": int,
+    },  # TODO not sure
```
Fizzadar: 👍  int is correct here (or none).

pyinfra/api/arguments.py
```diff
@@ -71,51 +84,81 @@ def generate_env(config, value):
...
+    "_get_pty": {
+        "description": "Whether to get a pseudoTTY when executing any commands.",
+        "type": bool,
+    },
+    "_stdin": {
+        "description": "String or buffer to send to the stdin of any commands.",
+        "type": str,  # TODO don't know the type for buffer
```
Fizzadar: This currently has to be a string, list or tuple object, although makes me think this should accept IO-like objects too! For now this should cover it, I'll add a new issue to track reading stdin from IO like objects.

```suggestion
        "type": Union[str, list, tuple],
```

pyinfra/api/arguments.py
```diff
@@ -71,51 +84,81 @@ def generate_env(config, value):
...
+        "type": bool,
     },
-    "_precondition": "Command to execute & check before the operation commands begin.",
-    "_postcondition": "Command to execute & check after the operation commands complete.",
-    "_on_success": "Callback function to execute on success.",
-    "_on_error": "Callback function to execute on error.",
+    "_precondition": {
+        "description": "Command to execute & check before the operation commands begin.",
+        "type": str,
+    },  # TODO not sure
```
Fizzadar: 👍 same for `_postcondition` too

pyinfra/api/arguments.py
```diff
@@ -71,51 +84,81 @@ def generate_env(config, value):
...
+    "_postcondition": {
+        "description": "Command to execute & check after the operation commands complete.",
+        "type": str,
+    },  # TODO not sure
+    "_on_success": {
+        "description": "Callback function to execute on success.",
+        "type": Callable,
```
Fizzadar:
```suggestion
        "type": Callable[[State, Host, str], None],
```

This is probably most correct here, no return values and is passed `(state, host, operation_hash)`.

pyinfra/api/arguments.py
```diff
@@ -71,51 +84,81 @@ def generate_env(config, value):
...
+    },  # TODO not sure if arguments can be typed too
 }
 
 # Execution kwargs are global - ie must be identical for every host
 execution_kwargs = {
     "_parallel": {
         "description": "Run this operation in batches of hosts.",
         "default": lambda config: config.PARALLEL,
+        "type": bool,
```
Fizzadar:
```suggestion
        "type": int,
```

This actually represents the size of each batch (with `None`/default being unlimited).

Fizzadar (commented):
Thanks for doing this so quick @StefanBRas! Have added suggestions and run through all the types!

StefanBRas:
No problem! I added the fixes from the code review - unfortunantly this gives a circular import on `Host` as you can see in the pipeline. Normaly this could be solved with a `if TYPE_CHECKING` import guard, but since they are used at "runtime" (stub-generation-time) that's not possible here. 

I'm not sure if it's best to loosen the type or refactor but in any case I think that's easier for you to do.

I think you should be able to directly edit the branch right?

Fizzadar (approved):
Awesome, thanks @StefanBRas! I pushed up a workaround for now, not ideal but means the type can stay in place and avoids the circular import.

PR #840, opened 2022-06-21
fix 'lock_current_sate' -> 'lock_current_state'

ghost:
Found a small typo when reading the source


Fizzadar (approved):
Thanks for catching & fixing this @bouke-sf!

PR #845, opened 2022-07-03
cli.md: fix shell autocompletion instructions

jaysoffian:
Per [Shell Completion](https://click.palletsprojects.com/en/8.1.x/shell-completion/) the correct form is:

```
{PROG_NAME}_COMPLETE set to {shell}_source
```

Note that it is `{shell}_source` and NOT `source_{shell}`.


Fizzadar (approved):

PR #848, opened 2022-07-08
Fix ssh_demo example

jaysoffian:
The `host.fact.<fact>` syntax was deprecated in 1.x.


Fizzadar (approved):

PR #849, opened 2022-07-09
Fix short facts whose base fact requires arguments

jaysoffian:
kwargs must be passed to get_short_facts, otherwise it doesn't come back to get_fact
when its called on the underlying fact. If the underlying fact requires any arguments 
a TypeError would be thrown.


Fizzadar (approved):

PR #852, opened 2022-07-09
Use unique SUDO_ASKPASS script per connection

jaysoffian:
Using a single SUDO_ASKPASS script prevents running multiple instances
of pyinfra against the same host, as one instance is likely to stomp on
the SUDO_ASKPASS while the other instance is still running.

We can fix the issue by running mktemp on the host to generate a unique
filename under `/tmp`. We chose `/tmp` in case the pyinfra connection
is interrupted before we remove the temp file, as it will get cleaned
out automatically by the host's normal processes for maintaining
temporary directories.

And as long as we're running mktemp on the host to generate the
temporary name, we may as well use that same shell command to generate
the SUDO_ASKPASS script itself, instead of sending it via SFTP. This
also allows us to work in scenarios where the host sshd doesn't accept
SFTP connections.

### DIFF
--- a/pyinfra/connectors/util.py
+++ b/pyinfra/connectors/util.py
@@ -1,6 +1,5 @@
 import shlex
 from getpass import getpass
-from io import StringIO
 from socket import timeout as timeout_error
 from subprocess import PIPE, Popen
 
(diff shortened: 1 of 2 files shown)


jaysoffian:
@Fizzadar what are your thoughts on this PR? FWIW, it's working reliably for me against about 100 macOS machines, but I haven't tried it against other platforms.

pyinfra/connectors/util.py
```diff
@@ -13,17 +12,17 @@
...
-
-def get_sudo_askpass_exe():
-    return StringIO(
-        """#!/bin/sh
-echo ${0}
+SUDO_ASKPASS_COMMAND = r"""
+temp=$(mktemp /tmp/pyinfra-sudo-askpass-XXXXXXXXXXXX)
+cat >"$temp"<<'__EOF__'
+#!/bin/sh
+printf '%s\n' "${0}"
```
Fizzadar, line 19: Why the switch to `printf` from `echo` here?

Fizzadar (approved):
@jaysoffian this looks great, glad to hear it's testing working as well 👍, thank you for doing this! Going to push up some additional SSH end to end tests then will get this merged ahead of `2.4` release.

pyinfra/connectors/util.py
jaysoffian: It's a habit I got into because `printf '%s' ...` is safe, reliable, and portable. `echo` can cause trouble if, e.g., the password starts with a dash, then echo ends up taking that as an option and there's no portable way to disable that behavior, nor a portable way to control how echo interprets escapes. More here:

https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

In fact, I notice now the existing (prior to this PR) script doesn't quote the variable at all, so it won't work with passwords that contain, say, two spaces in a row.

Fizzadar: Good habit to have! Definitely seems sensible to me, thank you!

PR #854, opened 2022-07-13
get_facts: run get_fact with context set

jaysoffian:
User-defined facts might expect to have a context in all situations,
not just when run as part of an operation. e.g, when called via
`pyinfra INVENTORY fact ...`.

Another use case is when setting `CONFIG.use_sudo_password` to a
function which wants to look-up the password as an attribute on a
host's group-data, the function might be surprised to find that host
isn't an instance of Host, but rather is the ctx_host.


Fizzadar (approved):
Thank you for fixing this @jaysoffian!

PR #863, opened 2022-08-03
first version of facts and operations for files.flags

morrison12:
This is a fact and an operation to support file "flags" as found in[ BSD-derived](https://github.com/morrison12/pyinfra/pull/new/flags) OSs (including macOS).
The first example for macOS would be automating the usual `chflags nohidden ~/Library` in many setup scripts.

### DIFF
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -38,8 +38,6 @@ SYMBOL_TO_OCTAL_PERMISSIONS = {
     "--x": "1",
 }
 
-FLAGS_PATTERN = re.compile(r"^[d-][rwx-]{9}.\s+\d+\s+\w+\s+\w+\s+([a-z,]+).*$")
-
 
 def _parse_mode(mode):
     """
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -367,7 +367,7 @@ class Flags(FactBase):
 
     def command(self, path):
         return make_formatted_string_command(
-            "stat -f %Sf {0} || true",
+            "! test -e {0} || stat -f %Sf {0}",
             QuoteString(path),
         )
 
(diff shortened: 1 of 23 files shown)


pyinfra/facts/files.py
```diff
@@ -354,3 +356,21 @@ class FindDirectories(FindFilesBase):
...
+class Flags(FactBase):
+    """
+    Returns a list of the file flags set for the specified file or directory.
+    """
+
+    requires_command = "chflags"  # don't try to retrieve them if we can't set them
+
+    def command(self, path):
+        return make_formatted_string_command(
+            "stat -f %Sf {0} || true",
```
Fizzadar:
```suggestion
            "! test -e {0} || stat -f %Sf {0}",
```

Fact have slowly been removing the use of `|| true` because it hides any error - this way we either do nothing if no file or run the `stat` command if the file exists.

pyinfra/facts/files.py
```diff
@@ -38,6 +38,8 @@
     "--x": "1",
 }
 
+FLAGS_PATTERN = re.compile(r"^[d-][rwx-]{9}.\s+\d+\s+\w+\s+\w+\s+([a-z,]+).*$")
```
Fizzadar: Is this unused now?

Fizzadar (commented):
Firstly - apologies for the delay in reviewing this @morrison12. This is a great addition, and thank you for the extensive testing on both fact & operation! Left a couple of very minor comments otherwise this is good to go 👍

pyinfra/facts/files.py
morrison12: that makes sense. change made.

pyinfra/facts/files.py
morrison12: yes - removed.

Fizzadar (approved):
Thank you once again @morrison12!

PR #868, opened 2022-08-08
Set context before calling use_sudo_password()

jaysoffian:
Set context before calling `use_sudo_password()` in case it wants access
to host data. A use case is setting the sudo password in group data
files, and then later providing it via config.USE_SUDO_PASSWORD. e.g.

    config.USE_SUDO_PASSWORD = lambda: host.data.sudo_password

### DIFF
--- a/pyinfra/connectors/util.py
+++ b/pyinfra/connectors/util.py
@@ -10,6 +10,7 @@ from gevent.queue import Queue
 from pyinfra import logger
 from pyinfra.api import MaskString, QuoteString, StringCommand
 from pyinfra.api.util import memoize
+from pyinfra.context import ctx_host, ctx_state
 
 SUDO_ASKPASS_ENV_VAR = "PYINFRA_SUDO_PASSWORD"
 SUDO_ASKPASS_COMMAND = r"""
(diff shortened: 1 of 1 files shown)


pyinfra/connectors/util.py
```diff
@@ -179,7 +180,9 @@ def _get_sudo_password(host, use_sudo_password):
             host.connector_data["sudo_password"] = sudo_password
         sudo_password = sudo_password
     elif callable(use_sudo_password):
-        sudo_password = use_sudo_password()
+        with ctx_state.use(state):
```
Fizzadar, line 183: This shouldn't be required as the state context is set for the entire run @ https://github.com/Fizzadar/pyinfra/blob/5fedec46b127d16d59ca73537f90dadff647511d/pyinfra_cli/main.py#L380

Fizzadar (commented):
Thank you for the PR @jaysoffian, would be good to have this working!

I'm not sure about setting the host context here though, might be better set for the entire duration of operation execution; I just pushed up https://github.com/Fizzadar/pyinfra/pull/869 to test if this works as expected...

Fizzadar:
@jaysoffian I have just merged in https://github.com/Fizzadar/pyinfra/commit/591ea61e3b29210ccb5ac71665920f1374a51e18 which I believe solves this issue (and any other host state needs within operation execution context).

jaysoffian:
> @jaysoffian I have just merged in [591ea61](https://github.com/Fizzadar/pyinfra/commit/591ea61e3b29210ccb5ac71665920f1374a51e18) which I believe solves this issue (and any other host state needs within operation execution context).

Indeed. Thanks!

Issue #870, opened 2022-08-10
CONTRIBUTING not consistent with code (setup.cfg and pyproject.toml)


morrison12:
## Describe the bug
The third bullet under [Contributing/Code Style](https://docs.pyinfra.com/en/latest/contributing.html#code-style) says _single quotes everywhere possible_ but 
1. in ``setup.cfg`` line 13 is ``inline-quotes = double`` 
and  
2. in ``pyproject.toml``  ``skip-string-normalization = true`` is not present in the ``[tool.black]`` section which ensures ``black`` will normalize quotes to double quotes 

I'm happy to provide a PR but wanted to check whether the style has changed first

## To Reproduce
n/a

## Expected behavior
If  [Contributing/Code Style](https://docs.pyinfra.com/en/latest/contributing.html#code-style) is still correct,  the settings for ``flake8`` and ``black`` should match it.

## Meta
current HEAD on 2022-08-10

Fizzadar:
Thanks for spotting this @morrison12! Updated to black/isort in https://github.com/Fizzadar/pyinfra/commit/94efd2b3d46df5dd9664dadc147d0730e601a1b9.

PR #871, opened 2022-08-10
expanded SELinux support

morrison12:
This is a first attempt at expanding SELinux support in PyInfra, specifically facts and operations for each of:
1. Booleans
1. Port labelling
1. File Contexts
1.  File Context Mappings

There are a few things I'm not sure I like yet:

1. There are two versions (SEPort and SEPortB)  of a fact to get the SELinux ’type' for a port:
- the former transfers a lot of lines (~500 on a vanilla Fedora 36 installation) and has to sort through it to build the port -> SELinux ’type mapping
- the latter transfers only a few lines but requires policycoreutils-dev to be installed which seems unlikely and
I think the former (SEPort) is probably the right answer and SEPortB should be removed but have provided both in case there are other considerations.

1. The name “file context mapping” for the information managed by semanage fcontext seems rather verbose.  Any suggestions would be useful.

1. I'm in a battle with flake8 on quotes (see #870)

Note I’ve deliberately left the support for user, role and range by ``selinux.filecontext``  to someone else as I don’t have a way to test them.

Comments/Corrections/Suggestions/etc welcome


pyinfra/operations/selinux.py
```diff
@@ -0,0 +1,177 @@
...
+"""
+from pyinfra import host
+from pyinfra.api import QuoteString, StringCommand, operation
+from pyinfra.facts.selinux import FileContext, FileContextMapping, SEBoolean, SEPort
+
+
+@operation(
+    pipeline_facts={"seboolean": "boolean"},
+)
+def se_boolean(boolean, value, persistent=False):
```
Fizzadar:
```suggestion
def boolean(boolean, value, persistent=False):
```

Consistent with the `file_context` operations I think `selinux.boolean` is suitable.

pyinfra/operations/selinux.py
```diff
@@ -0,0 +1,177 @@
...
+            yield StringCommand("semanage", "fcontext", "-d", QuoteString(target))
+            host.create_fact(FileContextMapping, kwargs=kwargs, data={})
+        else:
+            host.noop(f"no existing mapping for '{target}'")
+
+
+@operation(
+    pipeline_facts={"seport": "protocol, port"},
+)
+def se_port(protocol, port, se_type=None, present=True):
```
Fizzadar:
```suggestion
def port(protocol, port, se_type=None, present=True):
```

Same as above for boolean, `selinux.port`...

pyinfra/facts/selinux.py
```diff
@@ -28,16 +50,79 @@ def process(self, output):
...
-        return components[1]
+        # example output: /etc       all files          system_u:object_r:etc_t:s0 # noqa: SC100
+        # but lines at end that won't match: /etc/systemd/system = /usr/lib/systemd/system
+        if len(output) != 1:
+            return self.default()
+        m = re.match(r"^.*\s+(\w+):(\w+):(\w+):(\w+)", output[0])
+        return {k: m.group(i) for i, k in enumerate(FIELDS, 1)} if m is not None else self.default()
+
+
+class SEPort(FactBase):
```
Fizzadar: Hm, this fact collects all the ports/protocols so I think we could just rename this to `SEPorts` and return them all in a nested dictionary maybe something like:

```json5
{
     "tcp": { 22: "type" },
     "udp": {...},
}
```

pyinfra/facts/selinux.py
```diff
@@ -28,16 +50,79 @@ def process(self, output):
...
+                    pieces = item.split("-")
+                    start, stop = int(pieces[0]), int(pieces[1])
+                else:
+                    start = stop = int(item)
+                labels.update({port: m.group(1) for port in range(start, stop + 1)})
+
+        return labels.get(self.port, self.default())
+
+
+class SEPortB(FactBase):
```
Fizzadar: With the `SEPorts` fact this can just become `SEPort` and used as an optimisation if `sepolicy` is available on the system.

tests/operations/selinux.se_boolean/bad_value.json
```diff
@@ -0,0 +1,16 @@
...
+        "name": "ValueError",
+        "message": "'value' must be one of 'on,off' but found 'really-off'"
+    }
+}
+
+
+
+
+
+
```
Fizzadar: Extra newlines snuck in here!

Fizzadar (commented):
@morrison12 thank you for implementing these facts & operations, overall looks fantastic!

Left a few comments mainly a proposed solution to the port fact/issue you mentioned. Can't think of a better name than file context mapping - I think this is OK :)

pyinfra/operations/selinux.py
morrison12: changed.

pyinfra/operations/selinux.py
morrison12: changed

pyinfra/facts/selinux.py
morrison12: Done.. And the operation `selinux.port` changed to use the new format.  To minimize the "noise" in the dictionary, I added an `ignore` parameter to allow the caller to exclude certain types with a default that gets rid of the reserved-ish types.

pyinfra/facts/selinux.py
morrison12: Is it acceptable to have the operation `selinux.port` use `files.which` to use the optimization ? Or is it better to only have one way for the operation to execute ?

tests/operations/selinux.se_boolean/bad_value.json
morrison12: fixed.

pyinfra/facts/selinux.py
Fizzadar: Definitely acceptable, even encouraged; there's definitely other operations that do this 👍

morrison12: Done.

Fizzadar (approved):
Thank you @morrison12!

Issue #736, opened 2022-01-08
files.sync ignores mode on destination directory


si14:
`files.sync` always applies host's file permission on the destination directory regardless of `mode` in the command:
https://github.com/Fizzadar/pyinfra/blob/ff0913d6a172966760b63fe59e55dff9ea852e0d/pyinfra/operations/files.py#L573

It might be the intended behaviour, but it's fairly surprising, as code like this

```python
files.sync(
    name="Sync static",
    sudo=True,
    src=str(HOST_REPO_ROOT / "staticweb"),
    dest=str(TARGET_ROOT / "staticweb"),
    user="root",
    group="www-data",
    mode=660,
)
```

might lead to `staticweb` being listable for everyone on the target host.

sysadmin75:
I'm unable to reproduce this issue.  The files permissions and ownership in the deploy are honored.

Maybe this has been fixed since you opened the issue.  Maybe update your version and try again.

If you still have this issue, then provide more details on reproducing.

filips123:
This still happens to me on the latest version from PyPI. I created a simple [example repo](https://github.com/filips123/pyinfra-permissions-bug). After running `pyinfra inventory.py deploy.py`, everything is uploaded to the server. However, only files are given the specified permissions, while directories are set to `777`:

```sh
$ ls -al /opt/test/
total 12
drwxrwxrwx 3 root root 4096 Aug  9 19:25 ./
drwxr-xr-x 5 root root 4096 Aug  9 19:25 ../
-rw-r--r-- 1 root root    0 Aug  9 19:25 first.txt
drwxrwxrwx 2 root root 4096 Aug  9 19:25 nested/
-rw-r--r-- 1 root root    0 Aug  9 19:25 second.txt

$ ls -al /opt/test/nested/
total 8
drwxrwxrwx 2 root root 4096 Aug  9 19:25 ./
drwxrwxrwx 3 root root 4096 Aug  9 19:25 ../
-rw-r--r-- 1 root root    0 Aug  9 19:25 third.txt
```

Output of `pyinfra --support`:

```
    System: Windows
      Platform: Windows-10-10.0.19044-SP0
      Release: 10
      Machine: AMD64
    pyinfra: v2.3
    Executable: C:\Users\filips\AppData\Local\Temp\pyinfra-permissions-bug\venv\Scripts\pyinfra.exe
    Python: 3.10.5 (CPython, MSC v.1929 64 bit (AMD64))
```

[...]

PR #873, opened 2022-08-11
Add dir mode argument to files.sync

filips123:
This adds `dir_mode` argument to `files.sync` operation to allow explicitly setting permissions for directories. It's related to my comment in #736, but that issue probably shouldn't be closed yet, as `files.sync` still doesn't respect local directory permissions (at least on Windows).


Fizzadar (approved):
Thanks for implementing this @filips123! Excellent addition here :)

PR #876, opened 2022-08-14
Host loop cycle prevention

Fizzadar:


pyinfra/api/host.py
```diff
@@ -95,6 +95,14 @@ class Host(object):
     executing_op_hash = None
     nested_executing_op_hash = None
 
+    loop_position = 0
```
julienlavergne: Does it work with multiple embedded loops ? It seems using `host.loop` inside another `host.loop` is going to override the `loop_position`

docs/deploy-process.rst
```diff
@@ -106,40 +106,44 @@ Sometimes it is impossible to know all the facts before executing operations. Fo
...
+
+    # @local: A0 -> B0 -> A1 -> B1
+    # Other:  B0 -> A1 -> B1
+
+Problem is, by default ``pyinfra`` doesn't know the loop position. To work around this duplicate issues have numbers appended, so now we have:
+
+.. code:: shell
+
+    # @local: A -> B -> A-1 -> B-1
+    # Other:  B -> A -> B-1
```
julienlavergne: I am not so sure a user will understand what we are talking about here. I understand because I read your code, but without that, I don't think I would.
It looks like this example could be in the code, and the user documentation just need to make clear that `host.loop` must be used in every loop containing operations.

Fizzadar: Thank you for pointing this out, will have another go at better explaining/documenting this!

pyinfra/api/host.py
Fizzadar: Great spot, definitely won't work, will fix...

Fizzadar: Fixed by https://github.com/Fizzadar/pyinfra/pull/876/commits/838a6f79362e81e9df91b3f871a67f010ff90634 & tested in https://github.com/Fizzadar/pyinfra/pull/876/commits/650b6fc5e9666b7ead8128b010d95a4579cbc6c2

docs/deploy-process.rst
Fizzadar: Docs updated @ https://github.com/Fizzadar/pyinfra/pull/876/commits/c07908ebc02373f0117fda7584be5a3abbe56eba, have kept the technical bit (and cleaned it up) but prioritised the fix.

PR #882, opened 2022-08-29
Refactor

ccrvlh:
The idea here is to make functions smaller, and easier to comprehend (and maintain).
I tried to split the big functions (mainly the `main` cli, `operations` and `operation`) into smaller chunks that made sense (not sure I was able to). I tried to follow the logic already present using the comment blocks.

Example for the `main`:

```python
_setup_log_level()
_validate_operations()
_setup_state()
_set_override_data()
_set_config()
_set_verbosity()
_apply_inventory_limit()
```
Unfortunately this made the diff quite a pain specially for the `cli` part, I'm sorry about that. Hopefully looking at the code directly will make more sense. The good part is that it made the function a lot smaller (from ~350 to less than 100 lines), and easier to understand (in my opinion, of course).

Also tried to slightly reduce the indented and nested parts of the code (this usually happens with log messages) - as a first suggestion, I chose to go to the "more variables, less indentation path" but this is purely a personal choice (I personally feel is easier to read & follow). A lot of these are local/private, and won't be used anywhere else, hence the `_` prefix.

In general I chose to play safe, sometimes returning more information that what's necessary, mainly because there are a few places that change global vars, so I though it would be better to return more than what's necessary.

I made an effort not to change any piece of logic, just purely organizing functions. Tests are passing normally. I hope you guys like it, but I understand there's a lot of personal preferences here, let me now what you think and I'll be happy to adjust accordingly. If none of this makes sense, I'll also understand, no worries at all.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ docs/_deploy_globals.rst
 .idea/
 
 pyinfra-debug.log
+Makefile
--- a/pyinfra/api/operation.py
+++ b/pyinfra/api/operation.py
@@ -179,76 +179,9 @@ def operation(
                 )
             return func(*args, **kwargs) or []
 
-        # If this is a legacy operation function (ie - state & host arg kwargs), ensure that state
-        # and host are included as kwargs.
-        if func.is_legacy:
-            if "state" not in kwargs:
-                kwargs["state"] = state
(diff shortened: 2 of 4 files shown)


ccrvlh:
Closing for a new one with easier Diffs.

PR #883, opened 2022-08-30
Functions decomposition

ccrvlh:
The idea here is to make functions smaller, and easier to comprehend (and maintain).
I tried to split the big functions (mainly the main cli, operations and operation) into smaller chunks that made sense (not sure I was able to). I tried to follow the logic already present using the comment blocks.

Example for the main:
```
_setup_log_level()
_validate_operations()
_setup_state()
_set_override_data()
_set_config()
_set_verbosity()
_apply_inventory_limit()
```

Unfortunately this made the full diff quite a pain specially for the cli part, I’m sorry about that. Hopefully looking at the code directly will make more sense. The good part is that it made the function a lot smaller (from ~350 to less than 100 lines), and easier to understand (in my opinion, of course).

Also tried to slightly reduce the indented and nested parts of the code (this usually happens with log messages) - as a first suggestion, I chose to go to the “more variables, less indentation path” but this is purely a personal choice (I personally feel is easier to read & follow). A lot of these are local/private, and won’t be used anywhere else, hence the _ prefix.
In general I chose to play safe, sometimes returning more information that what’s necessary, mainly because there are a few places that change global vars, so I though it would be better to return more than what’s necessary.

I made an effort not to change any piece of logic, just purely organizing functions. Tests are passing normally. I hope you guys like it, but I understand there’s a lot of personal preferences here, let me now what you think and I’ll be happy to adjust accordingly. If none of this makes sense, I’ll also understand, no worries at all.

Edit: the last PR was hard to read because of the diff. Couldn't really make the diff better, but I split the commits a lot more, so the diff in each commit should be a lot easier to read.


pyinfra/api/operations.py
```diff
@@ -128,18 +128,12 @@ def run_host_op(state, host, op_hash):
             try:
                 status = command.execute(state, host, executor_kwargs)
             except Exception as e:  # Custom functions could do anything, so expect anything!
+                _formatted_exc = format_exception(e)
+                _error_msg = "Unexpected error in Python callback: {0}".format(_formatted_exc)
+                _error_msg_styled = click.style(_error_msg, "red")
+                _error_log = "{0}{1}".format(host.print_prefix, _error_msg_styled)
                 logger.warning(traceback.format_exc())
-                logger.error(
```
Fizzadar, line 132: Heh - this is such a huge improvement to readability, the nesting here makes my eyes burn 😆

pyinfra/api/operation.py
```diff
@@ -222,30 +219,9 @@ def decorated_func(*args, **kwargs):
...
-            "commands": commands,
-            "global_kwargs": global_kwargs,
-            "operation_meta": operation_meta,
-        }
-        state.set_op_data(host, op_hash, op_data)
-
-        # If we're already in the execution phase, execute this operation immediately
-        if state.is_executing:
-            _execute_immediately(state, host, op_data, op_meta, op_hash)
+        state, operation_meta = _update_state_meta(
```
Fizzadar: This needn't return `state` because it mutates it, which I think we should also add a commend indicating.

pyinfra/api/operation.py
```diff
@@ -306,50 +331,29 @@ def _solve_operation_consistency(names, state, host):
...
-        else:
-            name = func.__name__
-
-        names = {name}
-
-    if host.current_deploy_name:
-        names = {"{0} | {1}".format(host.current_deploy_name, name) for name in names}
-
-    return names, add_args
+def _ensure_shared_op_meta(state, op_hash, op_order, global_kwargs, names):
```
Fizzadar, line 313: Let's add a comment indicating this function modifies state.

Fizzadar (approved):
This looks great @lowercase00, thank you for doing this it makes things much more approachable. Left a few comments for myself mainly I'll add them in before merging this hopefully this weekend!

pyinfra/api/operation.py
Fizzadar: https://github.com/Fizzadar/pyinfra/pull/883/commits/f30aa877f7cdb5349d0f41f3459886c3255532af

pyinfra/api/operation.py
Fizzadar: https://github.com/Fizzadar/pyinfra/pull/883/commits/f30aa877f7cdb5349d0f41f3459886c3255532af

pyinfra/api/operations.py
ccrvlh: Yeah, it was a bit confusing with a lot of nesting. I don't know if this is the best solution, but at least it's "flat", a bit of an improvement I would say!

ccrvlh:
@Fizzadar let me know if there's anything else to change in here

Fizzadar:
All good @lowercase00 I'm just behind on things! Thank you for working on this, an excellent improvement 👍!

Issue #565, opened 2021-04-01
Add a `files.block` operation


Fizzadar:
Same kinda thing as Ansible's `blockinfile`.

julienlavergne:
For now, any way to reproduce the behavior with existing operations ?

Fizzadar:
Now added in #885 🎉!

PR #885, opened 2022-09-05
the start of an implementation of files.block

morrison12:
This is the beginning of an implementation of `files.block` (and `files.Block`) as contemplated in #565.

The main missing feature I can see is `assume_file_present` which is need for the common use case of installing a package and then configuring it although there could well be others.

Some "notes:
1. The current implementation uses `awk` which I assumed was a reasonably dependency as even busybody seems to have it ?
1. Not all `awk` implementations support in place editing so the code currently does the "write to a temporary file, if things worked then adjust the user/group/mode and mv it to the desired location" jig...  
1. I've tested this with a few `awk` implementations (both the native and GNU awks on macOS, GNU awk on Fedora and busybox awk in OpenWrt)

Some questions:
1. Given a file is created in the case where it doesn't exist, should I allow the caller to supplier the `user`, `group` and `mode` ?
1.  Despite the temporary file being used, for consistency with other operations would a `backup` option be useful ?
2. I have some, admittedly not speedy, functional tests; is there somewhere they should be stored ?


pyinfra/operations/files.py
```diff
@@ -320,17 +323,17 @@ def line(
...
-    if escape_regex_characters:
-        match_line = re.sub(r"([\.\\\+\*\?\[\^\]\$\(\)\{\}\-])", r"\\\1", match_line)
-
-    # Ensure we're matching a whole line, note: match may be a partial line so we
-    # put any matches on either side.
-    if not match_line.startswith("^"):
-        match_line = "^.*{0}".format(match_line)
-    if not match_line.endswith("$"):
-        match_line = "{0}.*$".format(match_line)
+    match_line = adjust_regex(line, escape_regex_characters)
```
Fizzadar, line 327: Nice!

pyinfra/operations/files.py
```diff
@@ -1526,3 +1529,189 @@ def directory(
...
+
+    Removal ignores ``content`` and ``line``
+
+    **Examples:**
+
+    .. code:: python
+
+        # add entry to /etc/host
+        files.marked_block(
+            name = "add IP address for red server",
```
Fizzadar: For consistency these should not have spaces between the keyword & value.

pyinfra/operations/files.py
```diff
@@ -1526,3 +1529,189 @@ def directory(
...
+            raise ValueError("'line' must be supplied or 'before' and 'after' must be equal")
+
+        the_block = "\n".join([mark_1, content, mark_2])
+
+        if (current is None) or ((current == []) and (before == after)):
+            # a) no file or b) file but no markers and we're adding at start or end. Both use 'cat'
+            redirect = ">" if (current is None) else ">>"
+            stdin = "- " if ((current == []) and before) else ""
+            # here = hex(random.randint(0, 2147483647))
+            here = "HERE"
```
Fizzadar: Should we use something like `PYINFRAHERE` to avoid clashes with any user content in the block?

Fizzadar (commented):
THANK YOU! @morrison12 this has been in the backlog way too long and I've been unable to un-wrangle how to approach it, massive thank you for figuring it out!

PR looks great overall, few very minor comments. Addressing your questions below:

> The main missing feature I can see is assume_file_present which is need for the common use case of installing a package and then configuring it although there could well be others.

Ah yes, very unfortunate gotcha. This operation seems like it'd be particularly complex to support. Maybe it could just append the block, or maybe we don't support it? Happy to leave this out of scope for this PR entirely.

 > The current implementation uses awk which I assumed was a reasonably dependency as even busybody seems to have it ?
 > Not all awk implementations support in place editing so the code currently does the "write to a temporary file, if things worked then adjust the user/group/mode and mv it to the desired location" jig...

👍 


Some questions:

> Given a file is created in the case where it doesn't exist, should I allow the caller to supplier the user, group and mode ?

Good question, I would say no for now given that `files.line` is the same. If needed this can be combined with `files.file` to achieve this.

> Despite the temporary file being used, for consistency with other operations would a backup option be useful ?

I think so for keeping the old version around somewhere.

> I have some, admittedly not speedy, functional tests; is there somewhere they should be stored ?

What do they look like? :) Maybe under `tests/test_operations/test_files.py`?

pyinfra/operations/files.py
morrison12: Fixed.

pyinfra/operations/files.py
morrison12: I changed it.  Originally I had a random number but that broke the idempotence testing.

morrison12:
> > The main missing feature I can see is assume_file_present which is need for the common use case of installing a package and then configuring it although there could well be others.
> 
> Ah yes, very unfortunate gotcha. This operation seems like it'd be particularly complex to support. Maybe it could just append the block, or maybe we don't support it? Happy to leave this out of scope for this PR entirely.
> 
I'll happily file the bug report once this is released and work on a solution but need to find some time to make sure it covers all the cases.  Perhaps mark the operation as "beta" in the meantime ?

Fizzadar (approved):
Thank you @morrison12 for implementing this, fantastic as ever! I have rebased & added the beta warning to it ready to release :)

morrison12:
Thanks @Fizzadar.  I've opened #900 to track the need for `assume_file_present`.

PR #891, opened 2022-10-13
Generic typechecking stub for operations

mariusmuja:
This adds generic type-checking support for the `@operation` decorator. 

One of the nice features of pyinfra is the ability to compose operations, but when doing so a typing stub needs to be added for each new operation (if one wanted to continue having working type-checking). With this all operations typing stubs are no longer required.

Here's a small example of a new operation composing two other operations:

```python
@operation
def configure_firewall(src: str):
    filename = basename(src)
    
    yield from files.put(
        src=src,
        dest=f"/etc/nftables/{filename}",
    )
    
    yield from files.line(
        path="/etc/sysconfig/nftables.conf",
        line=f'include "/etc/nftables/{filename}"',
    )
```
With this typing stub for the `@operation` decorator the type of `configure_firewall` is inferred correctly:
image

The type of all arguments are known to the type checker:
image
image

Argument auto-completion also works:
image

This uses `ParamSpec` from PEP-612, which is part of python 3.10. However, since all typing annotations are inside the `operation.pyi` stub, it doesn't impact the runtime and VSCode/pyright is happy to use the typing annotations with older python versions as well (tested on 3.8).


Fizzadar (commented):
@mariusmuja THANK YOU! Not going to lie I have really been struggling to wrap my head around paramspec and decorator typing. This is such a great improvement and really clears things up!

I have also noticed that managing the stubs is an unnecessary overhead for contributors and this removes that.

I need to make a couple of cleanups on other PR and will get this tested and merged in! Once again huge thank you for this!

Fizzadar (approved):
Thank you once again @mariusmuja, this is a massive improvement!

PR #895, opened 2022-10-21
Fix typo in heading

simonw:


Fizzadar (approved):
Thank you for fixing this @simonw!

PR #896, opened 2022-10-21
Fixed two typos

simonw:


Fizzadar (approved):
Thank you for fixing these @simonw!

PR #897, opened 2022-10-21
Spelling fixes, spotted using codespell

simonw:
I ran `pip install codespell` and then `codespell .` in the `pyinfra` directory. These are the fixes that it spotted which looked like they mattered (it had a bunch of noise too).


Fizzadar (approved):
Thank you for fixing these @simonw!

PR #898, opened 2022-10-21
Include systemd inactive units when checking for status

mariusmuja:
Added `--all` flag to `SystemdStatus` to include the inactive units. From the manual page of `systemctl`:

```
-a, --all
           When listing units, show all loaded units, regardless of their state, including inactive units.
```


Fizzadar (approved):
Good catch, @mariusmuja, thank you!

PR #899, opened 2022-10-21
Fix for RpmPackage fact to work with .rpm files as well as package names

mariusmuja:
Currently the `RpmPackage` fact only works for `.rpm` files and not also package names (because the command starts with `! test -e {1} || `). However the fact is used by the `dnf.rpm` opertation in both ways (from `packaging.ensure_rpm`):

```python
    # Check for file .rpm information
    info = host.get_fact(RpmPackage, name=source)
    exists = False

    # We have info!
    if info:
        current_package = host.get_fact(RpmPackage, name=info["name"])
        if current_package and current_package["version"] == info["version"]:
            exists = True
```

This PR allows the `RpmPackage` fact to work for both `.rpm` files and package names, fixing the `dnf.rpm` operation.

### DIFF
--- a/pyinfra/facts/rpm.py
+++ b/pyinfra/facts/rpm.py
@@ -44,8 +44,9 @@ class RpmPackage(FactBase):
 
     def command(self, name):
         return (
+            'rpm --queryformat "{0}" -q {1} || '
             "! test -e {1} || "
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice! Not only is this way better but it's also more readable 💪. Thank you @mariusmuja!

PR #901, opened 2022-10-24
Add types to get_fact and some facts

simon04:
Hi, based on #881 and https://github.com/Fizzadar/pyinfra/pull/881#issuecomment-1289105072, I tried to add some types to `get_facts`.

The idea is that the following example snippet can infer the type of `linux_distribution` and the existence of the key `"release_meta"`

https://github.com/Fizzadar/pyinfra/blob/2d4792eba3233c77a298d86e9638434688e3a1c5/examples/dnf.py#L18-L19


Fizzadar:
This is awesome, thank you for pulling this together @simon04! Have rebased so mypy properly checks things, I think the issue now comes down to conflating different signatures for `command` and `process` in the facts. Currently command may be either a `str` or static method or method, and `process` either a static method or method.

I think mypy is correct in flagging these as errors, and the way this works is a overly complex type magic. This will probably require some changes I'm happy to make when I have time:

+ add `FactBase.command_str` or similar for fixed commands
+ make `FactBase.command` always a callable method
+ make `FactBase.process` always a callable method

Fizzadar:
@simon04 this is really awesome work, it's going to break some things to land this so I'd like to roll this into upcoming pyinfra v3 if that's OK with you? Happy to fixup and changes to get this merged into the `v3` branch and we can then add types to each fact gradually.

simon04:
> I'd like to roll this into upcoming pyinfra v3 if that's OK with you?

Sure! :smile:

> Happy to fixup and changes to get this merged into the `v3` branch

Do you want me to change the base branch of this PR, or will you take action?

Fizzadar:
> > Happy to fixup and changes to get this merged into the `v3` branch
> 
> Do you want me to change the base branch of this PR, or will you take action?

I'll do it! Hoping to get this merged tonight :)

Fizzadar (approved):
Got there! Thank you so much for this @simon04, a huge improvement. Hoping to get v3 into beta over the next couple of weeks.

PR #902, opened 2022-10-24
Remove `Windows` prefix from all Windows facts.

simon04:
Fixes examples.

Relates to 3a274a04dad3f2a4ae85a02909638132dbdb437b.


Fizzadar (approved):
Thanks for catching & fixing these @simon04!

PR #903, opened 2022-10-24
pylint fixes

marksmayo:
Remedied a bunch of encoding warnings, simplied if/elses and some other small fixes.


Fizzadar (commented):
Thank you for the PR @marksmayo! This looks sensible to me, there's some conflicts with isort by the looks of it on some imports. I do think the conditional type imports should be at the bottom like this so, not sure how to convince isort to do that though...

marksmayo:
Yeah, forgot about isort. I've run it again, will update and see how we go.

Issue #907, opened 2022-11-03
Add `TEMPDIR` fact and use in place of default `config.TEMP_DIR`


Fizzadar:
Inspired by: https://github.com/Fizzadar/pyinfra/issues/880.

Currently temporary files (used for uploads/templates before moving to final location) use the `config.TEMP_FILE` parameter, which defaults to `/tmp`:

https://github.com/Fizzadar/pyinfra/blob/0e4fe02824b262f81e44d397827203d2172d73d8/pyinfra/api/config.py#L15-L16

While this works in many systems, it would be nicer if this was based on any `TEMPDIR` environment variable (unless the user explicitly sets `config.TEMP_DIR`). So the temp directory should be per-host and selected from (in order of preference):

+ An explicit `config.TEMP_DIR` (default changes to `None`)
+ The value of `TEMPDIR` environment variable on the target host
+ The builtin pyinfra default `/tmp` matching the current code, thus should be backwards compatible

TODO:

+ Move the `State.get_temp_filename` function to `Host.get_temp_filename`
+ Create new `server.TempDir` fact to get the correct location
+ make `Host.get_temp_filename` use the config or fact as above

Issue #806, opened 2022-05-04
Nested operations do not support programmatic sudo


themanifold:
## Describe the bug
Trying to use nested operations which require programmatic sudo access throws `AttributeError`

UPDATE: perhaps this should just be that nested operations do not support custom config attributes(?)

## To Reproduce
```
from pyinfra import config

config.USE_REMOTE_SUDO_PASSWORD = "something"

def callback():
    result = server.shell(
        commands=["cat foo"],
        _sudo=True,
        _use_sudo_password=config.USE_REMOTE_SUDO_PASSWORD
    )

def caller(vm_name):
    python.call(
        name="Execute callback function",
        function=callback,
    )

[server.foo] Unexpected error in Python callback: AttributeError("'Config' object has no attribute 'USE_REMOTE_SUDO_PASSWORD'",)
```

## Expected behavior
I would expect a nested operation to support programmtic sudo access

## Meta
+ Include output of `pyinfra --support`.
```
    System: Linux
      Platform: Linux-5.10.105-1-MANJARO-x86_64-with-glibc2.35
      Release: 5.10.105-1-MANJARO
      Machine: x86_64
    pyinfra: v2.1
    Executable: /home/user/workspace/something/venv/bin/pyinfra
    Python: 3.10.4 (CPython, GCC 11.2.0)
```

Fizzadar:
The config variable here should be `USE_SUDO_PASSWORD`, that should work as expected!

themanifold:
OK, so I think what I used to do was create a global variable in the config object call `USE_REMOTE_PASSWORD` and that I passed to `use_sudo_password`. Now what I'm doing is just setting `config.USE_SUDO_PASSWORD` and `config.SUDO` at the top of the code and everything now works apart from the nested operation which just seems to ignore the config settings and not embed the shell snippet within the usual `env SUDO_ASKPASS=pyinfra-sudo-askpass *** sh -c '<shell snippet'>` like I would expect.

Fizzadar:
Very slow getting back on this, apologies. So the reason this fails is because the config object is unique per host and also per stage (first fact gathering then execution) - but the config for one host is not shared between each stage. This the function call here uses the original config without the extra attribute:

https://github.com/Fizzadar/pyinfra/blob/fa23aba0ac8742c246bbfacfb889caab1c796a98/pyinfra/api/command.py#L202

[...]

PR #912, opened 2022-11-11
Get default temp dir from env / host-scoped config

Fizzadar:
For: #907


pyinfra/api/host.py
```diff
@@ -245,6 +247,30 @@ def deploy(self, name: str, kwargs, data, in_deploy: bool = True):
             old_deploy_data,
         )
 
+    def get_temp_filename(self, hash_key: Optional[str] = None, hash_filename: bool = True):
+        """
+        Generate a temporary filename for this deploy.
+        """
+
+        temp_directory = self.state.config.TEMP_DIR
```
Fizzadar, line 255: Config should be host-scoped once state initalized? https://github.com/Fizzadar/pyinfra/issues/806

Fizzadar: Historical context: https://github.com/Fizzadar/pyinfra/commit/9fe1984b968a17552092ef6dc1482724a4125404

Fizzadar:
This is now being deferred for v3 (https://github.com/pyinfra-dev/pyinfra/pull/996) as it is technically a breaking change.

PR #913, opened 2022-11-16
Fix local connector put with spaces

uggedal:
Destination path was not quoted before being copied from it's temp file.


Fizzadar (approved):
Amazing thank you for catching, fixing and testing this @uggedal!

PR #914, opened 2022-11-16
Fix OpenRC status and Apk package fact detection

dchauviere:
This PR fix some detection problem with openrc and apk facts :

* Apk package fact
   include numbers and underscores in package's name search regex
   package like `iproute2` or `linux-firmware-rtl_nic` are now detected 

* OpenRC fact
   include extended status format (started_until date and restart_count) in rc-status output parser
   supervised service with following status are now detected :

   ```
   podman                                   [  started 00:34:14 (10) ]
   podman                                   [  started 15 day(s) 00:34:14 (10) ]
   ```

### DIFF
--- a/pyinfra/facts/apk.py
+++ b/pyinfra/facts/apk.py
@@ -2,7 +2,7 @@ from pyinfra.api import FactBase
 
 from .util.packaging import parse_packages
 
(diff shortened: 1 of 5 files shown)


Fizzadar (approved):
Thank you for catching, fixing & testing this @dchauviere!

PR #921, opened 2022-12-08
Fix handling of mysql privileges

gchazot:
I hope the code change is self-explanatory.

Happy to rewrite the commit history if it's not gonna be a squash merge.

* Fixed the regex that caused the Fact to always return an empty list, whatever the output
* Ignore "USAGE" privilege in the Fact since this is a placeholder that means "no privilege"
* Fix special cases of the "ALL" (aka "ALL PRIVILEGE") case where it conflicts with "GRANT OPTION"


Fizzadar (approved):
Amazing as always @gchazot! Thoroughly tested 💪, thank you!

PR #922, opened 2022-12-08
Fix mysql load command

gchazot:
* Test highlighting the problem: Missing quotes around the filename.
* Fix for that problem

The fix also includes the `< '<filename>'` as 2 additional bits in the `StringCommand`. This is not just for aesthetics, it actually fixed a problem I was having but couldn't get to the bottom of it as the command I see in the operation (`--debug-operation`) works if I exactly copy-paste it in a terminal.

I suspect it was that both bits in `< '<filename>'` used to be passed to the shell as a single argument, but without quoting it's impossible to see that.

Maybe `--debug-operations` should print out the bits rather than the formatted command.

### DIFF
--- a/pyinfra/operations/mysql.py
+++ b/pyinfra/operations/mysql.py
@@ -14,7 +14,7 @@ See the example/mysql.py
 """
 
 from pyinfra import host
-from pyinfra.api import MaskString, OperationError, StringCommand, operation
+from pyinfra.api import MaskString, OperationError, QuoteString, StringCommand, operation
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice catch here @gchazot, thank you for fixing & testing!

Fizzadar:
> Maybe --debug-operations should print out the bits rather than the formatted command.

100% agree, I think debug-operations needs a full rework in general (https://github.com/Fizzadar/pyinfra/issues/924).

PR #925, opened 2022-12-13
Allow passing SSH key directly in CLI

Fizzadar:
This feels terribly wrong, but enables automation flows such as calling pyinfra directly from a Terraform local-exec provisioner if the key is also created by Terraform in the same run (ie it's not available as an output or on the filesystem).


Fizzadar:
Feels too wrong, a better approach would probably be a wrapper bash script to write the key into a temporary file and then call pyinfra.

PR #931, opened 2022-12-17
Do not reuse temporary file in server.script()

link2xt:
Even if `src` is the same, other parameters
such as `_su_user` may be different. In this
case pyinfra unconditionally tries to run `chmod +x`
on the file and fails, because the file is
already uploaded as a different user.

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -172,7 +172,7 @@ def script(src):
         )
     """
 
-    temp_file = state.get_temp_filename(src)
+    temp_file = state.get_temp_filename()
     yield from files.put(src, temp_file)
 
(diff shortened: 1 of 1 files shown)


link2xt:
Random docker test failure seems to be unrelated?

link2xt:
Rebased the PR, maybe #932 is the reason.

Fizzadar (approved):
> Even if `src` is the same, other parameters such as `_su_user` may be different. In this case pyinfra unconditionally tries to run `chmod +x` on the file and fails, because the file is already uploaded as a different user.

Nice spot @link2xt, this makes sense. Scripts are generally very small as well so the saving here really is tiny/over optimisation!

PR #941, opened 2023-01-15
Fix `files.Md5File` fact for BSD-style output

ScottKevill:
First noticed that `files.Md5File` had a copy/paste error in its BSD-style output regex while browsing the code.

Then confirmed `Md5File` was indeed missing the test case even though `Sha1File` and `Sha256File` covered it.

This PR adds the test case, and then fixes the bug.


Fizzadar (approved):
@ScottKevill amazing, thank you for fixing & testing this!

ScottKevill:
No problem, pyinfra has been quite refreshing. :) I have a few more PRs lined up, too.

PR #942, opened 2023-01-16
Refactor `files.HashFile` facts

ScottKevill:
This just refactors `Sha1File`, `Sha256File`, and `Md5File` to derive from a common `HashFileFactBase` to remove the previous duplicated code.

`HashFileFactBase` uses `__init_subclass__()` so that the class variables are only initialised once for each class (rather than for every instance), but this did require tweaking `FactNameMeta` to ignore extra `kwarg` parameters. Python docs are pretty bad about those function signatures.

It also extracts the hash name from the subclass name (stripping the `File` suffix) as a bit of sneaky magic, but the tests will catch any breakage if the subclassses are renamed later.

Two slight changes to the behaviour that should not pose a problem:
- the regexes are tightened to match just hex digits rather than alphanumeric
- the `sha1`, `sha256`, and `md5` commands now also silence `stderr` with `2> /dev/null` the same as the other commands

Did this separately to the previous `Md5File` fix in case you didn't like this one. ;)

### DIFF
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -5,6 +5,7 @@ The files facts provide information about the filesystem and it's contents on th
 import re
 import stat
 from datetime import datetime
+from typing import List, Tuple
 
 from pyinfra.api.command import QuoteString, make_formatted_string_command
 from pyinfra.api.facts import FactBase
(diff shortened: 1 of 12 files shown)


pyinfra/facts/files.py
```diff
@@ -204,91 +205,57 @@ class Socket(File):
...
-    _regexes = [
-        r"^([a-zA-Z0-9]{40})\s+%s$",
-        r"^SHA1\s+\(%s\)\s+=\s+([a-zA-Z0-9]{40})$",
-    ]
+        raw_hash_cmds = ["%s {0} 2> /dev/null" % cmd for cmd in cmds]
+        raw_hash_cmd = " || ".join(raw_hash_cmds)
+        cls._raw_cmd = "test -e {0} && ( %s ) || true" % raw_hash_cmd
+
+        assert cls.__name__.endswith("File")
+        hash_name = cls.__name__[:-4].upper()
```
Fizzadar, line 220: This is probably a bit too magic, but I can't see a nicer way aside from passing the value as an extra kwarg. Will merge for now as you mention the tests will catch any issues here, more just a code understanding problem we can address in the future.

Fizzadar (approved):
Apologies for taking a while to get to this (& the others) @ScottKevill - this looks great a nice cleanup and improvement with the stderr redirects 👍

PR #944, opened 2023-01-16
Fix bug typos

ScottKevill:
These typos did affect code/test behaviour, so I've added them as a separate commit / PR.


Fizzadar (approved):
Thanks for fixing these as well!

PR #946, opened 2023-01-25
remove duplicate DOAS, DOAS_USER

minusf:


Fizzadar (approved):
Nice catch @minusf, thanks for cleanup!

PR #948, opened 2023-01-29
fix typo in docs

belonesox:
remove wrong quote


Fizzadar (approved):
Thanks for fixing this @belonesox!

PR #952, opened 2023-02-10
fix for import error init

v3ss0n:
Current example is using outdated API , replaced `init` with `systemd`

### DIFF
--- a/examples/docker_ce.py
+++ b/examples/docker_ce.py
@@ -1,6 +1,6 @@
 from pyinfra import config, host
 from pyinfra.facts.server import LinuxName, LsbRelease
-from pyinfra.operations import apt, init, python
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thanks for fixing this @v3ss0n!

examples/docker_ce.py
```diff
@@ -1,6 +1,6 @@
 from pyinfra import config, host
 from pyinfra.facts.server import LinuxName, LsbRelease
-from pyinfra.operations import apt, init, python
+from pyinfra.operations import apt, systemd, python
```
Fizzadar:
```suggestion
from pyinfra.operations import apt, python, systemd
```

PR #954, opened 2023-02-15
Cargo Packages Prototype

wowi42:
Hi,

The target is to be able to install some cargo packages, like starship.rs, or vigil-server

```
cargo install vigil-server
```

### DIFF
--- a/tests/operations/cargo.packages/remove_packages.json
+++ b/tests/operations/cargo.packages/remove_packages.json
@@ -1,0 +1,16 @@
+{
+    "args": [["git@1.1", "react", "test@1.1"]],
+    "kwargs": {
+        "present": false
+    },
+    "facts": {
+        "cargo.CargoPackages": {
+            "git": ["1.0"],
+            "react": [""],
+            "test": ["1.1"]
+        }
+    },
+    "commands": [
+        "cargo uninstall react test@1.1"
+    ]
+}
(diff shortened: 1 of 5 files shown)


tests/operations/cargo.packages/remove_packages.json
```diff
@@ -0,0 +1,16 @@
...
+    "facts": {
+        "cargo.CargoPackages": {
+            "git": ["1.0"],
+            "react": [""],
+            "test": ["1.1"]
+        }
+    },
+    "commands": [
+        "cargo uninstall react test@1.1"
+    ]
```
gchazot, line 15:
```suggestion
    "args": [["git@1.1", "react", "test@1.1", "ansible"]],
    "kwargs": {
        "present": false
    },
    "facts": {
        "cargo.CargoPackages": {
            "git": ["1.0"],
            "react": [""],
            "test": ["1.1"],
            "ansible": ["1.2.1.0"]
        }
    },
    "commands": [
        "cargo uninstall react test@1.1 ansible"
    ]
```
Just adding the missing case where the installed package are a version but I assume that no version in the operation call means "any version".

Also, that Fact where `react` has no value in the list of versions, is it a valid case? I would expect that either a version is installed and then listed, or the package is not installed and hence not present at all in the Fact output...

Fizzadar (approved):
Apologies for the delay @wowi42, life has been busy! This looks great to me, I don't have write permissions on the fork but will apply @gchazot's suggestion after merge!

Fizzadar:
For reference: https://github.com/Fizzadar/pyinfra/commit/746976157c2fb9176c0681c37be040fe3f05c742

PR #957, opened 2023-02-20
Appimage builder

v3ss0n:
- added appimage building support , it directly pulls from 2.x branch and then compiles into appimage binary.
- added a build-time dependency for python-appimage 
- run `appimage.sh` to build pyinfra as `AppImage`

### DIFF
--- a/appimage.sh
+++ b/appimage.sh
@@ -1,0 +1,5 @@
+:#!/bin/bash
+pip install python-appimage
+python-appimage install appimagetool &&
+python-appimage install patchelf &&
+python-appimage build app -p 3.11 ./appimage
--- a/appimage.sh
+++ b/appimage.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-pip install python-appimage
+pipx install python-appimage
 python-appimage install appimagetool &&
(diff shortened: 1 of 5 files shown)


v3ss0n:
why this test has to fail ? i didn't add any python code there , just appimage builder bash script.

v3ss0n:
ready to merge.

Fizzadar (commented):
Hey @v3ss0n, apologies for the long delay on reviewing this, I didn't want to rush a decision and life has been extremely busy recently!

Currently I don't think adding this into pyinfra itself is a good idea - it's an additional distribution method for pyinfra. More distribution methods leads to more issues with each and I already have barely enough time to keep things rolling forward.

There is also the [pyinfraller binaries](https://github.com/Fizzadar/pyinfra/pull/768) work which will provide yet more distribution methods (but not Linux only) which I am unsure about continuing for the same reason.

I think this would currently be better suited in a separate repository dedicated to pyinfra-appimage generation/distribution.

v3ss0n:
i see , i will help maintain with a seperate repo , do you want to host inside pyinfra organization?

PR #966, opened 2023-03-13
Fix build and tests

gchazot:
This fixes the current development head for tests to pass.

The diff here also highlights the changes to errors that will be given to the user upon using non-existent facts/operations. 

I don't think these changes are good in terms of user experience. I will try to add a commit to improve the situation with something simple.

### DIFF
--- a/pyinfra_cli/util.py
+++ b/pyinfra_cli/util.py
@@ -138,20 +138,23 @@ def try_import_module_attribute(path, prefix=None):
     module = None
 
     if prefix:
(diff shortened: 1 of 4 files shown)


gchazot:
BTW, the tests were broken on the head branch from this commit:
https://github.com/Fizzadar/pyinfra/commit/63b06d2f4efcaab45d42182fab8788c7e564e2ed

Fizzadar (approved):
Heh, cleaning up my mess 😅 , thank you @gchazot 🙌!

PR #968, opened 2023-03-13
Git config add

gchazot:
Built on top of https://github.com/Fizzadar/pyinfra/pull/966 for CI tests

Allow to add git config key/value pairs for options that support multiple lines such as `safe.directory`.

NOTE: This changes slightly the API of the `git.GitConfig` Fact, since it now always returns a list of values for each key present. I think this is more in line with the possible structure of the configuration file though.

The [git config file documentation](https://git-scm.com/docs/git-config#_configuration_file), shows that there's an undefined number of possible options, with no real way to know (programmatically) whether an option allows multiple values. So the design here (let the `pyinfra` user specify it's a multi_value option) stands.

### DIFF
--- a/pyinfra/facts/git.py
+++ b/pyinfra/facts/git.py
@@ -33,7 +33,7 @@ class GitConfig(FactBase):
 
         for line in output:
             key, value = line.split("=", 1)
-            items[key] = value
+            items.setdefault(key, []).append(value)
 
         return items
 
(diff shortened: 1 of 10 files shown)


Fizzadar (approved):
Awesome stuff as always @gchazot! I think the small change to the fact is OK here, as it better represents the underlying config file.

PR #969, opened 2023-03-19
make put_file work with doas

minusf:


Fizzadar (approved):
Thank you for adding this @minusf! I pushed a quick test & fix pre-merge 👍

Issue #961, opened 2023-03-03
Cronjobs/line replacements ending `2>&1` break


Fizzadar:
## Describe the bug

When ending a line replacement with `2>&1` the line is not replaced properly and a broken line, growing each deploy, is instead. Repro below.

## To Reproduce
Steps to reproduce the behavior, please include where possible:

```
files.line(
    name="Ensure step cron",
    path="/usr/lib/cron/tabs/root",
    line="/usr/local/bin/step",
    replace="0 * * * * /usr/local/bin/step ca renew --root ca.crt --ca-url https://ca.com mtls.crt mtls.key >> /tmp/step-renew.log 2&>1",
)
```

## Expected behavior

Replace the line properly

Fizzadar:
Fixed in https://github.com/pyinfra-dev/pyinfra/commit/d1092ad63b5c1659a9cfc1e224c9b6ce3a311ae7

PR #972, opened 2023-03-26
Escape & character during sed replacement during line.replace

sysadmin75:
fixes issue #961


Fizzadar (approved):
Thank you for fixing this @sysadmin75 🙏

PR #978, opened 2023-04-16
Install requirement failed as codecov package is no longer available

ioO:
On April 12, 2023, codecov package was removed on Pypi repository. I have forked the project to add a new operation but I got an error during `pip install -e '.[dev]'`. This pull request remove the package from dependecies and update the workflow to use codecov v3 as v1 was deprecated.


Fizzadar (approved):
Thank you for fixing this @ioO!

PR #980, opened 2023-04-18
Manage locales on linux systems

maisim:
- Add server.Locales fact to retrieve enabled locales
- Add server.locale operation to enable a locale


Fizzadar (approved):
Great addition @maisim, thank you! Pushed a small test change (https://github.com/Fizzadar/pyinfra/pull/980/commits/128680f0b2c60495c894f1f49909897d6ce08262), now merging!

PR #991, opened 2023-06-18
Fix misspelling in docs

bnmoch3:
Minor change: "direct" to "directly"


Fizzadar (approved):
Thanks for fixing this @bnm3k!

PR #993, opened 2023-07-14
Escape database name in mysql.database operation

tissieres:
Hello,
Thanks for pyinfra!
I noticed that the database name was not correctly escaped in the mysql.database operation.
Cheers


Fizzadar (approved):
@tissieres thank you for the PR! Sorry it took so long to review, looks good to me, I think the tests got lost between changes on the 2 branch but all looks good.

PR #995, opened 2023-07-19
Eval operations at execution where state unknown

### LINKED ISSUE #805: What are the advantages of gathering facts beforehand ?
I have been modifying a local version of pyinfra in order to solve different issues (some of them posted as github issues here), and I came to the conclusion that there is not much advantages in gathering facts before operations.
Actually, most of the limitations and weird behaviors I see would be solved by gathering facts along the way.

I would be interested to discuss to pros and cons and contribute in modifying the way pyinfra work if necessary.

I have a small list of the advantages of gathering facts before operations that requires them:
- Since there is no need to come up with a list of facts to gather beforehand, the whole problem of ordering operations goes away. Especially, the code that browse the stack to come up with operation order is unnecessary, and this code does not output the correct order in several cases.
- There is no need for the `preserve_loop_order` magic anymore. It was anyway very counter-intuitive to have operations in loops not executed in the expected order.
- Most usage of `assume_present` arguments becomes unnecessary since facts will reflect the correct state of the machine right before an operation is performed.
- The execution flow is easier to understand, for example, simple things like creating a directory and checking if that directory exists would simply works.
- The whole concept of nested operations is moot or becomes very limited. It seems the main use case was to run an operation, get the output and perform more operations based on the output. But with facts gathered along the way, the output of the command is available immediately and conditional logic can be written directly within the main python script (instead of the callback).
- The is no need to support dynamic facts, since facts are dynamic by default.
- We can still operate a cache to make sure facts are not gathered unnecessary and, just like ansible, provide ways to invalidate the cache for arbitrary operations like `server.shell`.

Fizzadar:
Closes: #805


Fizzadar:
Cancelling this in favour of #996, there’s just too much magic already and this change adds yet more. V3 will radically simplify and address these issues properly.

PR #996, opened 2023-07-23
pyinfra v3 WIP

Fizzadar:


DonDebonair:
Hey @Fizzadar, after getting increasingly frustrated with Ansible over the past months, I was researching alternatives and found PyInfra. I'm inclined to try it out, but I noticed this draft PR which seems to suggest V3 is being actively worked on.

Can you elaborate how much V3 will differ from V2 in terms of API? How smooth will the migration path from stable to V3 be? In other words: is it safe start using the stable version of PyInfra right now with a relatively smooth upgrade to V3 once it hits?

I'm also interested in contributing. I saw that there are no Docker/Docker Compose related operations yet, and I need those. But it would be a problem if the API is going to change significantly and operations need to be changed a lot for V3.

Could you elaborate?

Fizzadar:
> Hey @Fizzadar, after getting increasingly frustrated with Ansible over the past months, I was researching alternatives and found PyInfra. I'm inclined to try it out, but I noticed this draft PR which seems to suggest V3 is being actively worked on.
> 
> Can you elaborate how much V3 will differ from V2 in terms of API? How smooth will the migration path from stable to V3 be? In other words: is it safe start using the stable version of PyInfra right now with a relatively smooth upgrade to V3 once it hits?
> 
> I'm also interested in contributing. I saw that there are no Docker/Docker Compose related operations yet, and I need those. But it would be a problem if the API is going to change significantly and operations need to be changed a lot for V3.
> 
> Could you elaborate?

Hi @DonDebonair! Apologies for taking ages to respond here. Ultimately for the public stuff (ie anything in the docs) there should be no significant changes. Any changes will display a warning in v2. The majority of v3 changes are in the CLI (approval step by default) and internal APIs. I am intending for v3 to provide official support for API usage as well (which v1 had).

So in short: migration between v2/v3 should be trivial in almost every case, any changes needed will show as warnings in v2.

Final thing - for operations there’s no significant change to those APIs so any ops designed for 2 will work in 3 (and any contributions would be very welcomed on both versions!)

Hope this helps!

DonDebonair:
Hey @Fizzadar thanks for the response! I will start using PyInfra then. Do you want new operations - like for example ops to manage docker containers and images and interact with docker compose - to be part of this repo, or should they live somewhere separate? When I'm ready to contribute, I can make a PR to this or another repo

Fizzadar:
Github please will you run the workflow!

PR #997, opened 2023-08-08
Prevent Paramiko size mismatch errors in utf-8 strings

sbj-ss:
Length of unicode strings consisting of multibyte characters differs from their size in bytes. Hence uploads of files and templates using unicode characters will fail with messages like
```--> Starting operation: tasks/app.py | scheduler crontab 
    Failed to upload file, retrying: size mismatch in put!  208 != 190
    Failed to upload file, retrying: size mismatch in put!  208 != 190
    Failed to upload file, retrying: size mismatch in put!  208 != 190
    [172.33.2.57] Command socket/SSH error: OSError('size mismatch in put!  208 != 190',)
    [172.33.2.57] Error: executed 0/1 commands
```
Using BytesIO will prevent this.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -885,7 +885,7 @@ def put(
 
         files.put(
             name="Upload a BytesIO object",
-            src=BytesIO("file contents"),
+            src=BytesIO("file contents".encode("utf-8")),
             dest="/etc/motd",
         )
     """
@@ -1052,7 +1052,7 @@ def template(src, dest, user=None, group=None, mode=None, create_remote_dir=True
         {% for entry in foo_list %}
             - "{{ entry }}"
         {% endfor %}
-        """)
+        """.encode("utf-8"))
 
         files.template(
             name="Create a templated file",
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
@sbj-ss thank you for the PR, this definitely seems sensible to me! I’ll fix up those test issues as they’re hidden in some of the legacy test stuff and get this merged!

Fizzadar:
As part of the v3 work I incorporated this change at a lower level which I think better handles these cases automatically:

- auto convert: https://github.com/pyinfra-dev/pyinfra/commit/6be2aa40ff964029111dbcbcb819c6de04f4ef98
- tests: https://github.com/pyinfra-dev/pyinfra/commit/d8c120503cfe268529fc9f576d70ff67bcbae780

Usually I hate implicit behavior (lots of v3 work removes it!) but in this case I think this is best since passing `StringIO` objects in/out of operations is standard and recommended. This ensures that the output file always matches what is requested in `get_file_io`. Thank you once again for the PR, inspiration and fix @sbj-ss!

Issue #888, opened 2022-10-04
Support HTTP/S proxies for various operations


themanifold:
## Is your feature request related to a problem? Please describe
I work with various "airgapped" machines that require using a corporate proxy to communicate to the outside internet, which pyinfra does not support the use of

## Describe the solution you'd like
The ability to pass a custom http/s proxy to various pyinfra operations. Examples include:
* files.download()
* apt operations

The way I see this, there's two ways to achieve this:
* Add extra arguments to each operation one by one.
* Add the ability to set environment variables to all pyinfra operations, e.g. `export HTTP_PROXY=<whatever>` etc.

## Workarounds
I work around this by altering the pyinfra code and adding a proxy to the curl command, for example.

Fizzadar:
I _think_ this is resolved, the `files.download` PR covers that and the `_env` global argument could be used to set the `HTTP_PROXY` env for `apt.*`. I'm not sure there's much value adding extra proxy specific args to all other operations.

Issue #646, opened 2021-08-18
Global or per-operation class env vars


taranlu-houzz:
## Is your feature request related to a problem? Please describe
Not sure if this is actually a feature that would make sense to add, but this is the issue that I am dealing with:

The command `brew update` used by the `brew.update()` operation automatically stashes any local changes that have been made to brew recipes (one might do this if you need to patch a brew recipe to get at a specific version that doesn't have an `@version`). The way to get around this is to set the environment var `HOMEBREW_DEVELOPER` when running the brew commands. In order to do this via `pyinfra`, I need to pass that in via the `env` param to each `brew.*` operation, which is tedious and can be error prone (since `brew` will just stash the modifications and carry on as though nothing out of the ordinary happened).

## Describe the solution you'd like
I'd like a way to specify a set of environment vars that will be passed automatically to all invocations of a particular operation class, possibly as some sort of special `group_data` key?

Maybe, something like this:
```python
pyinfra_operations_brew = {
    "HOMBREW_DEVELOPER": 1,
}
```
There is probably a more elegant solution, but that is the gist of what I am hoping for.

Fizzadar:
One option here is to use the `ENV` config var which would apply to all operations within a deploy; ie:

```py
# config.py

ENV = {
    "HOMBREW_DEVELOPER": 1,
}

# or, new style recommended in v1.5+

from pyinfra import config

config.ENV = {
    "HOMBREW_DEVELOPER": 1,
}
```

I believe this would provide a suitable solution? Setting environment variables for a given set of operations would be a bit confusing, I think, vs. this global approach.

Fizzadar:
The `_env` global argument supports this now. https://docs.pyinfra.com/en/3.x/arguments.html#shell-control-features

PR #998, opened 2023-08-09
Add proxy support for files.download()

themanifold:
This should resolve the most basic version of https://github.com/Fizzadar/pyinfra/issues/888 - it'll support using a basic http proxy with `files.download()`  via the `download` parameter. It should work with curl and wget

In terms of tests, I need some guidance as I'm not understanding the dynamic nature of them in this project.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -69,6 +69,7 @@ def download(
     md5sum=None,
     headers=None,
     insecure=False,
+    proxy=None,
 ):
     """
     Download files from remote locations using ``curl`` or ``wget``.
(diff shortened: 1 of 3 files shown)


themanifold:
I just noticed that there's an ability to set global environment variables https://github.com/Fizzadar/pyinfra/issues/646#issuecomment-903298143

Fizzadar (approved):
Thank you for the PR @themanifold! My apologies for the delay getting back here; for tests this should basically be a case of duplicating and modifying one of the JSON operation tests here: https://github.com/Fizzadar/pyinfra/tree/2.x/tests/operations/files.download

Made a note to better document this as well.

Fizzadar:
I just pushed a quick couple of tests up to get this out here: https://github.com/pyinfra-dev/pyinfra/pull/998/commits/b239dfd338b0f97ed286c2569c21202f0687c76b. Will be making sure the contributing guide is properly updated to document the test generation!

PR #1000, opened 2023-08-12
Fix issue where we strip a newline when public key is read

### LINKED ISSUE #999: Adding a public key does not work
## Describe the bug
Adding a public key to a user does not work when there are already existing keys.

## To Reproduce
I have this deploy:

```python
@deploy("ssh key")
def ssh_keys():
    with open("a.pub") as f:
        k1 = f.read()
    with open("b.pub") as f:
        k2 = f.read()
    server.user("user", public_keys=[k1, k2])
```

I add another key:

```python
@deploy("ssh key")
def ssh_keys():
    with open("a.pub") as f:
        k1 = f.read()
    with open("b.pub") as f:
        k2 = f.read()
    with open("c.pub") as f:
        k3 = f.read()
    server.user("user", public_keys=[k1, k2, k3])
```

## Expected behavior
I expect k3 to be added.

## Meta

```
    System: Linux
      Platform: Linux-6.4.7-100.fc37.x86_64-x86_64-with-glibc2.36
      Release: 6.4.7-100.fc37.x86_64
      Machine: x86_64
    pyinfra: v2.7
    Executable: /home/sebels/venv/bin/pyinfra
    Python: 3.11.4 (CPython, GCC 12.3.1 20230508 (Red Hat 12.3.1-1))
```

pip

## My Debug

I was able to trace the issue down to :
https://github.com/Fizzadar/pyinfra/blob/44ce3fa8ed689ffca6cc79aadfa61f4b72754bac/pyinfra/operations/files.py#L473

As commented in the code this is the situation we have: Line(s) exists and we want to ensure they're correct. But we never enter the "if" in the line below:
https://github.com/Fizzadar/pyinfra/blob/44ce3fa8ed689ffca6cc79aadfa61f4b72754bac/pyinfra/operations/files.py#L476
This is because "replace" is an empty string "".
Could it be the if should be "replace is None" instead of "replace is ''"?

sysadmin75:
The public key is read with the newline appended to the string.  This caused the `process` command in FindInFile to return any existing files.

Fixes #999

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -862,7 +862,7 @@ def user_authorized_keys(
 
         if path.exists(try_path):
             with open(try_path, "r") as f:
-                return f.read()
+                return f.read().strip()
 
         return key
 
--- a/tests/operations/server.user/key_files.json
+++ b/tests/operations/server.user/key_files.json
@@ -7,7 +7,7 @@
     "local_files": {
         "files": {
             "somefile.pub": "somekeydata",
-            "anotherfile.pub": "someotherkeydata"
+            "anotherfile.pub": "someotherkeydata\n"
         },
         "dirs": {}
     },


Fizzadar (approved):
Thanks for fixing this @sysadmin75!

PR #1001, opened 2023-08-18
fix fact lookup for systemd units containing '.'

cawo-odoo:
e.g.:
- unit name is `postgresql@9.5-main.service`
- user calls `server.service(service="postgresql@9.5-main", ...)` Because the unit name contains a dot, the ".service" extension is not added and the fact lookup always returns False, even if the service is in fact enabled.

Fix: explicitly check for unit type endings instead of just a dot.

### DIFF
--- a/pyinfra/operations/systemd.py
+++ b/pyinfra/operations/systemd.py
@@ -85,7 +85,21 @@ def service(
         user_name=user_name,
     )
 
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Thank you for the PR @cawo-odoo!

PR #1004, opened 2023-08-24
Really do not create home when server.user create_home is False

julienlavergne:
Home folders may be created by default following a setting in /etc/login.defs.
See https://linux.die.net/man/8/useradd.

Make sure it is not created when create_home is False.


julienlavergne:
@Fizzadar

Fizzadar (approved):
Apologies for the long delay on this one @julienlavergne, thank you for the PR!

PR #1005, opened 2023-08-25
Include docstring for uid/gid in server.{user,group}

jessexoc:
Update docstring for server.user() to include uid parameter and server.group() to include gid parameter


Fizzadar (approved):
Thanks for the PR @jessexoc!

PR #1009, opened 2023-08-28
Add "/" to connector prefix in docs

ghost:
Add a "/" the connector prefix in the Inventory & Data docs section, so that users know the full syntax immediately.


Fizzadar (approved):
Thanks for the PR @jarpy!

PR #1012, opened 2023-09-02
Add arguments to `operations.script_template`

### LINKED ISSUE #964: Add `args` to `server.script` & `server.script_template`

chriskingio:
I was going to fix `operations.script` as well, but it seems like someone got to that first (and gave prior art to copy).

fixes #964


Fizzadar (approved):
Thanks for adding this @chriskingio! It was me that added it to `server.script` and I have no idea how I failed to add this as well 🤦

PR #1013, opened 2023-09-03
Update `server.mount` to allow for `fs_type` and `device` options

### LINKED ISSUE #963: Expand `server.mount`: type and arbitrary mount location support
ie `mount -t apfs /dev/diskXYZ /mnt` kind of thing.

chriskingio:
This might require a review round before it's good to merge, depending on how "complete" you want to make this. Areas for potential improvement (either now or later - your choice):

1. Updating facts with more 'complete' information (fs type, device, etc)
2. Handling the case where mount is already mounted but fs_type or device is different (the former I'm not sure how that'd work, but I could see someone cloning data to another partition and wanting to swap to the cloned device)

Fixes #963

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -276,10 +276,12 @@ def modprobe(module, present=True, force=False):
 @operation
 def mount(
     path,
+    device=None,
+    fs_type=None,
     mounted=True,
     options=None,
     # TODO: do we want to manage fstab here?
-    # update_fstab=False, device=None, fs_type=None,
+    # update_fstab=False,
 ):
     """
     Manage mounted filesystems.
@@ -296,7 +298,6 @@ def mount(
         This operation does not attempt to modify the on disk fstab file - for
         that you should use the `files.line operation <./files.html#files-line>`_.
     """
-
     options = options or []
     options_string = ",".join(options)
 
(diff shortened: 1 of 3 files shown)


chriskingio:
I see from the unit tests why the spacing was important. I'll work on those next.

pyinfra/operations/server.py
```diff
@@ -276,10 +276,12 @@ def partition(predicate, iterable):
 @operation
 def mount(
     path,
+    device=None,
+    fs_type=None,
     mounted=True,
     options=None,
```
Fizzadar:
```suggestion
    mounted=True,
    options=None,
    device=None,
    fs_type=None,
```

Let's keep the existing order just in case anyone is passing args not kwargs.

pyinfra/operations/server.py
```diff
@@ -305,10 +306,17 @@ def mount(
...
+        if fs_type:
+            args.extend(["-t", fs_type])
+        if options_string:
+            args.extend(["-o", options_string])
+        if device:
+            args.append(device)
+        args.append(path)
+
+        yield StringCommand("mount", *args)
+        # Should we update facts with fs_type, device, etc?
```
Fizzadar, line 319: Can ignore this for now, v3 removes all of this nonsense!

Fizzadar (approved):
Thank you for the PR @chriskingio!

PR #1015, opened 2023-09-04
Fix Users and Groups module for FreeBSD

wowi42:
This PR took an unexpected turn; initially, my intention was to address the user-related issues on FreeBSD. However, I found myself delving deep into the codebase.

The current version appears to be functioning correctly. I've implemented a check to ensure compatibility with FreeBSD by utilizing the pw command. Nevertheless, I'm uncertain about retaining these exceptions within the codebase.


wowi42:
Found a very interesting bug:

by default, the shell in csh, which is very different from bash and sh. It will make the for loop completely fail.

Fizzadar (approved):
Apologies for the delay reviewing this @wowi42 - looks great thank you for adding FreeBSD support here!

PR #1036, opened 2023-11-20
Manage /etc/security/limits.conf

maisim:
Add operation and fact to set and fetch values from /etc/security/limits.conf


Fizzadar (approved):
Really nice @maisim 👍 thank your for the PR!

PR #1039, opened 2023-11-21
Add some pipx operations & facts

maisim:
Start pipx support,

Some tests do not pass but I am not sure of the way to write them in the correct way

WIP


pyinfra/facts/pipx.py
```diff
@@ -0,0 +1,81 @@
...
+        }
+    """
+
+    default = dict
+    pipx_command = "pipx"
+
+    def requires_command(self, pipx=None):
+        return pipx or self.pipx_command
+
+    def command(self, pipx=None):
```
Fizzadar, line 42: I think we can just always use `pipx` as the command, there's no case where we'd have multiple versions alongside each other I think?

pyinfra/operations/pipx.py
```diff
@@ -0,0 +1,100 @@
...
+        pipx.packages(
+            name="Install ",
+            packages=["pyinfra"],
+        )
+    """
+
+    # We should always use the --force flag because
+    # if install it's called, that meen the version missmatch
+    # so we need the --force flag to get the good version installed
+    install_command = [pipx, "install", "--force"]
```
Fizzadar, line 47: I think force should be a keyword argument for the operation, rather than added by default. We should be hitting the upgrade command if a package is already installed?

pyinfra/operations/pipx.py
```diff
@@ -0,0 +1,100 @@
...
+    install_command = " ".join(install_command)
+
+    uninstall_command = " ".join([pipx, "uninstall"])
+    upgrade_command = " ".join([pipx, "upgrade"])
+
+    # Handle passed in packages
+    if packages:
+        current_packages = host.get_fact(PipxPackages, pipx=pipx)
+
+        if current_packages is None:
```
Fizzadar, line 59: This should never be the case, facts return a default value so can drop this check.

Fizzadar (changes requested):
Hi @maisim! Apologies for the epic delay getting a proper review on this. Left a few comments, let me know what you think!

maisim:
superseded by #1243

PR #1040, opened 2023-11-22
add `password` option to `server.user`

disser:


pyinfra/facts/server.py
```diff
@@ -417,6 +417,7 @@ class Users(FactBase):
                 "uid": user_id,
                 "gid": main_user_group_id,
                 "lastlog": last_login_time,
+                "password": encrypted_password,
```
maisim, line 420: Hi @disser !
I am no sure that get encrypted password in the Fact is a good idea.
Do you have a use case for this ?
Maybe just password = True/False should be enough ?
Regards,

disser: The use case is to allow changing the password of an existing account to be idempotent [here](https://github.com/pyinfra-dev/pyinfra/pull/1040/files#diff-93ced8bf570709d17b4670036e1757810812c8af258c2740196f8ac2cd82a2e4R1121-R1123).

Fizzadar (approved):
This is a great add @disser, thank you for the PR!

PR #1041, opened 2023-11-22
Fix mypy error and update ssh test to get around paramiko config parser issue

maisim:
Fix error:
pyinfra/api/config.py:57: error: Unsupported operand types for in ("Version" and "Requirement") Found 1 error in 1 file (checked 122 source files)


Fizzadar (approved):
Appreciate you fixing these two @maisim 🙏

PR #1045, opened 2023-12-08
Fix server Users fact when user description contains a pipe character

julienlavergne:
There is an unrecoverable exception when one of the user description does contains a pipe.

@Fizzadar It makes the entire Users fact not usable, and prior versions of pyinfra also have the issue. It would be great to have a new version released after this is merged.


maisim:
Hi @julienlavergne 
Well see
It should be nice to remove the 2 last commits from you pr has they are useless, you can reset your branch to 1d32b10
On your 'patch-1' branch you can:

`git reset 1d32b100d09db580ccd0438f1d5d69af013ad515`
`git push origin --force`
where 'origin' is your fork remote name

PR #1060, opened 2024-01-17
Retry ssh connection on failure

fwiesel:
Especially with lots of host going through a jump-host, there is a chance that the jump-host is overloaded. 
Instead of simply failing, we can retry and back-off with some random interval to avoid the stampede

### DIFF
--- a/pyinfra/connectors/ssh.py
+++ b/pyinfra/connectors/ssh.py
@@ -235,7 +235,8 @@ def connect(state: "State", host: "Host"):
     retries = host.data.get(DATA_KEYS.connect_retries, 0)
 
     for tries_left in range(retries, -1, -1):
-        if con := _connect(state, host, tries_left):
+        con = _connect(state, host, tries_left)
+        if con:
             return con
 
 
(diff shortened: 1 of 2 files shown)


pyinfra/connectors/ssh.py
```diff
@@ -227,6 +229,14 @@ def connect(state: "State", host: "Host"):
     design so can be run in parallel.
     """
 
+    retries = state.config.CONNECT_RETRIES
```
Fizzadar: I think this should be defined as host data (see the `DATA_KEYS` in this file) rather than config, that gives us more flexibility to control things per host.

So I'd have like defaults as so:

```
host.data.ssh_connect_retries = 0
host.data.ssh_connect_retry_min_delay = 0.1
host.data.ssh_connect_retry_max_delay = 0.5
```

Fizzadar (changes requested):
Big fan of this, @fwiesel, thank you for the PR! One change to move the configuration into host data and this is good to go :)

pyinfra/connectors/ssh.py
fwiesel: Sure, I moved it it to host data and added a unit-test, which fixed also a bug (range is excluding stop).
```diff
-  for tries_left in range(retries, 0, -1):
+  for tries_left in range(retries, -1, -1): 
```

Fizzadar (approved):
Great stuff @fwiesel - thank you for the PR!

PR #1063, opened 2024-01-28
Deal with tab aligned dpkg output correctly

martenlienen:
Hey, thanks for the great package!

When I use the `DebPackage` fact on a .deb file, I noticed that it always returned an empty dictionary. This is caused on my system because at least some versions of dpkg align their output with tabs to make it look nice.

In this PR, I have added a complete and real output of dpkg with tab alignment and extended the regexes to parse this output correctly.

This should probably also go in the 3.x branch.

PS: Is there a way to signal an error in the `process` of a `Fact`? If so, it might make sense to signal an error if `DebPackage` is about to return an empty dictionary.

### DIFF
--- a/pyinfra/facts/deb.py
+++ b/pyinfra/facts/deb.py
@@ -48,8 +48,8 @@ class DebPackage(FactBase):
     """
 
     _regexes = {
-        "name": r"^Package: ({0})$".format(DEB_PACKAGE_NAME_REGEX),
-        "version": r"^Version: ({0})$".format(DEB_PACKAGE_VERSION_REGEX),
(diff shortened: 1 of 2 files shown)


martenlienen:
Looks like two checks failed because github was unavailable?! Should I force-push to trigger another run?

Fizzadar:
> Looks like two checks failed because github was unavailable?! Should I force-push to trigger another run?

Very annoying! Kicked off a retry

martenlienen:
Great, this time it worked and all checks pass.

Fizzadar (approved):
Thank you for the PR and for checking in @martenlienen!

PR #1072, opened 2024-02-27
Allow tab aligned dpkg output

martenlienen:
I noticed that this PR was merged into the 2.x branch but not 3.x and 3.x looks great, so it would be a shame if the bug would resurface when we upgrade. Therefore, I sent the same PR again against the 3.x branch.

Best,
Marten


Fizzadar (approved):
Thanks for carrying this over @martenlienen, haven’t got round to cherry picking everything into 3 yet!

PR #1076, opened 2024-03-21
Fix docker.get_file for binary files

matthijskooijman:
The file to get is copied to a temp file, and then opened to read into memory. This would open the file in text mode (the default), which would fail on binary files that could not be decoded as utf8.

This changes the file opening to use binary mode. It also simplifies the rest of the code, which would try to handle both strings and bytes, but this is likely a remnant of py2 code (or the maybe copied from files.put where the passed file could possibly be a stream in text mode).

### DIFF
--- a/pyinfra/connectors/docker.py
+++ b/pyinfra/connectors/docker.py
@@ -265,17 +265,10 @@ class DockerConnector(BaseConnector):
             )
 
             # Load the temporary file and write it to our file or IO object
-            with open(temp_filename, encoding="utf-8") as temp_f:
+            with open(temp_filename, "rb") as temp_f:
                 with get_file_io(filename_or_io, "wb") as file_io:
                     data = temp_f.read()
-                    data_bytes: bytes
-
-                    if isinstance(data, str):
-                        data_bytes = data.encode()
-                    else:
-                        data_bytes = data
-
-                    file_io.write(data_bytes)
+                    file_io.write(data)
         finally:
             os.close(fd)
             os.remove(temp_filename)


matthijskooijman:
To reproduce this error:

```
$ cat binary_test.py 
files.put(src="binary_file", dest="/binary_file")
files.get(src="/binary_file", dest="/tmp/binary_file")
$ echo -ne '\xf3' > binary_file
$ pyinfra @docker/debian:bookworm-slim binary_test.py --yes
--> Starting operation: files.get (src=/binary_file, dest=/tmp/binary_file)
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/api/operations.py", line 184, in _run_host_op_with_context
    return run_host_op(state, host, op_hash)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/api/operations.py", line 53, in run_host_op
    return _run_host_op(state, host, op_hash)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/api/operations.py", line 121, in _run_host_op
    status = command.execute(state, host, connector_arguments)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/api/command.py", line 191, in execute
    return host.get_file(
           ^^^^^^^^^^^^^^
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/api/host.py", line 429, in get_file
    return self.connector.get_file(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/matthijs/.local/pipx/venvs/pyinfra/lib/python3.11/site-packages/pyinfra/connectors/docker.py", line 270, in get_file
    data = temp_f.read()
           ^^^^^^^^^^^^^
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf3 in position 0: unexpected end of data
2024-03-21T21:54:39Z <Greenlet at 0x72ef51c844a0: _run_host_op_with_context(<pyinfra.api.state.State object at 0x72ef52013e50>, Host(@docker/debian:bookworm-slim), '9d1afe224119c48d16c1b791f0422627e757bed3')> failed with UnicodeDecodeError
```

I have not added any unit tests, since the normal docker unit tests patch out all of the actual docker stuff, so testing that the right calls are done does not seem meaningfull. Maybe a test could be added to the end-to-end tests, but it was not immediately clear to me if that would be a good idea and how to approach that (also, I need to get back to my actual project and pop a few levels of yak shaving here ;-p).

Fizzadar (approved):
Apologies this took so long to review @matthijskooijman, massive thank you for fixing this!

PR #1080, opened 2024-04-04
Just added a missing comma in the example for the `apt.upgrade` operation

la3mmchen:
Without that comma, the example is not working :)


Fizzadar (approved):
Nice catch @la3mmchen! Thank you for fixing :)

PR #1087, opened 2024-04-22
Fix for extra whitespace in DnfRepositories fact entries

bbonf:
On a clean RHEL 9.3 install, `pyinfra.facts.dnf.DnfRepositores` doesn't work as expected because there are extra spaces appended to dictionary keys, and prepanded to values.

That's because the parsing code splits on `'='` but doesn't handle whitespace.

Also relevant for the `3.x` branch

### DIFF
--- a/pyinfra/facts/util/packaging.py
+++ b/pyinfra/facts/util/packaging.py
@@ -32,7 +32,7 @@ def _parse_yum_or_zypper_repositories(output):
             current_repo["name"] = line[1:-1]
 
         if current_repo and "=" in line:
-            key, value = line.split("=", 1)
+            key, value = re.split(r"\s*=\s*", line, maxsplit=1)
             current_repo[key] = value
 
     if current_repo:


Fizzadar:
GH seems to be confused (as usual) and is throwing a total nonsense error for _this PR only_ despite no changes in that area.

Fizzadar (approved):
Thank you for fixing this @bbonf! Do you have an example repo line by any chance? Would be great to get a test added.

bbonf:
Sure, here's an excerpt from one file:
```
[rhel-atomic-7-cdk-3.6-source-rpms]
name = Red Hat Container Development Kit 3.6 /(Source RPMs)
```

And here's another (which is parsed correctly without this change):
```
[epel]
name=epel
```

Fizzadar:
Apologies for taking so long here @bbonf; I've cherry-picked this into v3 with a test in https://github.com/pyinfra-dev/pyinfra/commit/2c5aa7493573807ed75e003f7da7436a2777a075, thank you for the fix!

PR #1091, opened 2024-04-30
fix sphinx syntax

szepeviktor:
> for inline literals, use double backticks

```bash
pip install --user sphinx-lint
git ls-files --cached -z -- '*.rst' \
    | xargs --null -- python -m sphinxlint --enable all --disable line-too-long
```

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,6 +1,6 @@
 Using the API
 =============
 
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Nice catch @szepeviktor, thank you for the fix!

szepeviktor:
Glad to contribute.

PR #1094, opened 2024-05-04
Fix sudo password handling

CD3:
First, thank you for pyinfra. I have been using pyinfra 2 to manage my personal dev machines after trying fabric and ansible, and really liked it, but with version 3's addition of `_if`, it is perfect for what I need.

The only issue I had when switching to 3.x was that the sudo password pompt no longer worked. Pyinfra would ask for the sudo password, but then immediately fail.

The problem was that the connector utility function that handled setting the sudo password was checking to see if `_sudo_password` was in the argument list, but not checking if it was set. Something like (i've shorted some var names)

```
a['_sudo_password'] = a.get( '_sudo_password', h.data.get('prompted_sudo_password')
```
but `a['_sudo_password']` is set to `None`, so the prompted password never get used.

The other issue was that the command string that was generated to pass the sudo password to an askpass executable was not quoting the password

```
PYINFRA_SUDO_PASSWORD=password
```

which does not work if the password contains special shell characters like `;' or spaces.

### DIFF
--- a/pyinfra/connectors/util.py
+++ b/pyinfra/connectors/util.py
@@ -339,7 +339,7 @@ def make_unix_command(
             [
                 "env",
                 "SUDO_ASKPASS={0}".format(_sudo_askpass_path),
-                MaskString("{0}={1}".format(SUDO_ASKPASS_ENV_VAR, _sudo_password)),
+                MaskString("{0}='{1}'".format(SUDO_ASKPASS_ENV_VAR, _sudo_password)),
             ],
         )
 
--- a/pyinfra/connectors/util.py
+++ b/pyinfra/connectors/util.py
@@ -274,12 +274,11 @@ def make_unix_command_for_host(
         # If no sudo, we've nothing to do here
         return make_unix_command(command, **command_arguments)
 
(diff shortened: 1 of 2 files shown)


pyinfra/connectors/util.py
```diff
@@ -339,7 +338,7 @@ def make_unix_command(
             [
                 "env",
                 "SUDO_ASKPASS={0}".format(_sudo_askpass_path),
-                MaskString("{0}={1}".format(SUDO_ASKPASS_ENV_VAR, _sudo_password)),
+                MaskString("{0}='{1}'".format(SUDO_ASKPASS_ENV_VAR, _sudo_password)),
```
pbrkr: I think `shlex.quote()` would be better here - that way it's safe even if the password includes single-quotes.

```
MaskString("{0}={1}".format(SUDO_ASKPASS_ENV_VAR, shlex.quote(_sudo_password))),
```

CD3: Oh yes, of course. I fixed it.

Fizzadar (approved):
Fantastic stuff @CD3, thank you for fixing this AND adding a great test case 🎉

PR #1105, opened 2024-05-20
PyinfraOperation: Put *args first so that mypy correctly handles inner operation args

cdleonard:
This is not a mypy issue - unnamed arguments are passed in order so if PyinfraOperation were a real class instead of a protocol then PyinfraOperation.__call__("a", "b") would pass "a" and "b" to _sudo and _sudo_user.

This is the minimum fix to avoid errors like these:

```
pyinfra_deploy.py:6: error: Too few arguments for "__call__" of "PyinfraOperation"  [call-arg]
pyinfra_deploy.py:7: error: Argument 1 to "__call__" of "PyinfraOperation" has incompatible type "str"; expected "bool"  [arg-type]
```

I saw https://github.com/pyinfra-dev/pyinfra/pull/1082 and this change is not included

### DIFF
--- a/pyinfra/api/arguments_typed.py
+++ b/pyinfra/api/arguments_typed.py
@@ -32,6 +32,11 @@ class PyinfraOperation(Generic[P], Protocol):
     def __call__(
         self,
         #
+        # op args
+        # needs to be first
+        #
+        *args: P.args,
+        #
         # ConnectorArguments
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Aha! Great catch @cdleonard, thank you for the PR!

PR #1107, opened 2024-05-23
Update dev dependencies and clean up mypy errors

martenlienen:
I was about to start working on a small fix when I noticed that `pytest` gave me over 1,000 warnings on python 3.12. So I took the opportunity to update the test dependencies. Everything still works, warnings are down to 47 and code coverage gives the same numbers while feeling a bit faster. I did not measure. While I was at it, I also upgraded the formatting dependencies and tacked on a small fix of two typing errors that mypy complains about.

Best,
Marten

### DIFF
--- a/pyinfra/api/arguments_typed.py
+++ b/pyinfra/api/arguments_typed.py
@@ -77,5 +77,4 @@ class PyinfraOperation(Generic[P], Protocol):
         # op kwargs
         #
         **kwargs: P.kwargs,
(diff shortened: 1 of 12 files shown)


martenlienen:
@Fizzadar Now that the minimum supported python version is 3.8, this PR also works

Fizzadar (approved):
Thank you for this (and the other) PRs @martenlienen 🙏!

PR #1108, opened 2024-05-23
Update minimum python version to 3.8

martenlienen:
This PR updates the minimum required python version to 3.8 and specifies that explicitly in `setup.py`. Version 3.7 has been [end-of-lifed for a year](https://devguide.python.org/versions/) at this point.

This should also fix the error in PR #1107.

Does pyinfra have an explicit compatibility policy? Does it make sense for this project to be compatible with all python versions out there? Since this is not a library but a standalone tool like ansible, it does not really have to support all active python versions, but could also rely on some newer language features, or do I see that incorrectly?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-latest]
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
I think this is sensible. Generally try to cover roughly 1yr post EOL but no reasoning behind that other than wider compatibility.

PR #1109, opened 2024-05-23
Allow single function _if

martenlienen:
Always requiring `_if` to be a list requires a lot of unnecessary syntax in the common special case of exactly one condition, so this commit allows to specify a single condition directly.


Fizzadar (approved):
Nice, thanks @martenlienen!

PR #1111, opened 2024-05-24
Do not mistake inventory files for python modules

### LINKED ISSUE #1069: pyinfra3 inventory.py is neither a valid Python import or hostname - warning
## Describe the bug

pyinfra3 throws a warning on "Loading inventory" when using "inventory.py":

"inventory.py is neither a valid Python import or hostname."

... but it gets loaded anyway.

using ./inventory.py instead of inventory.py removes the error

## To Reproduce
try with pyinfra v2.9 vs 3.0.b0

```
pyinfra ./inventory.py exec -- echo hello
pyinfra inventory.py exec -- echo hello
```

## Expected behavior
A consistent behaviour, (so one option) or both examples possible.

## Meta
+ How was pyinfra installed: pip

martenlienen:
Check for inventory files first, so that they are never mistaken for python modules or hostnames and avoid unnecessary warnings. Also, support IPv6-only hosts.

Fixes #1069.

### DIFF
--- a/pyinfra/api/inventory.py
+++ b/pyinfra/api/inventory.py
@@ -37,6 +37,10 @@ class Inventory:
 
     state: "State"
 
+    @staticmethod
+    def empty():
+        return Inventory(([], {}))
+
     def __init__(self, names_data, override_data=None, **groups):
         # Setup basics
         self.groups = defaultdict(list)  # lists of Host objects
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
Thanks for cleaning this up @martenlienen, much improved!

PR #1112, opened 2024-05-26
Remove debug output

martenlienen:
A recent commit left in a debug output of facts. This PR removes it.


Fizzadar (approved):
🤦  nice catch, thanks @martenlienen!

PR #1115, opened 2024-05-30
Ensure that Fact#command does not take global arguments

martenlienen:
A recent commit c89e21142523b540ba8cf18dc8d54daf5113773c changed getting facts so that arguments to `FactBase#command` clashing with global arguments are filtered from the kwargs. This includes the often used argument `name` and facts like `DebPackage` broke as a result.

This PR introduces a check that subclasses of `FactBase` don't take any global arguments in their `command` method and updates all existing facts and tests accordingly. There is also a tiny bit of improved shell quoting in the deb, pacman and rpm facts.

### DIFF
--- a/pyinfra/api/facts.py
+++ b/pyinfra/api/facts.py
@@ -10,6 +10,7 @@ other host B while I operate on this host A).
 
 from __future__ import annotations
 
+import inspect
 import re
 from inspect import getcallargs
 from socket import error as socket_error, timeout as timeout_error
@@ -32,7 +33,7 @@ from paramiko import SSHException
 
 from pyinfra import logger
 from pyinfra.api import StringCommand
-from pyinfra.api.arguments import pop_global_arguments
+from pyinfra.api.arguments import all_global_arguments, pop_global_arguments
 from pyinfra.api.util import (
     get_kwargs_str,
     log_error_or_warning,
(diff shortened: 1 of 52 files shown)


Fizzadar:
> A recent commit [c89e211](https://github.com/pyinfra-dev/pyinfra/commit/c89e21142523b540ba8cf18dc8d54daf5113773c) changed getting facts so that arguments to `FactBase#command` clashing with global arguments are filtered from the kwargs. This includes the often used argument `name` and facts like `DebPackage` broke as a result.

heh, guess we know _why_ the old handling existed! Lack of comments on that meant I had long forgotten the context which makes sense! This approach is cleaner, and although the global name argument clash is annoying it makes sense to keep things consistent between facts and operations.

Not quite got time to review code right now but thank you for pulling this one together @martenlienen 🙏

pyinfra/api/facts.py
```diff
@@ -76,6 +77,17 @@ def __init_subclass__(cls) -> None:
         module_name = cls.__module__.replace("pyinfra.facts.", "")
         cls.name = f"{module_name}.{cls.__name__}"
 
+        # Check that fact's `command` method does not inadvertently take a global
+        # argument, most commonly `name`.
+        if hasattr(cls, "command") and callable(cls.command):
+            command_args = set(inspect.signature(cls.command).parameters.keys())
```
Fizzadar, line 83: This is neat, will copy this to operations as well!

Fizzadar (approved):
Thank you @martenlienen, this is great work. Also good to change this now ahead of 3.0 with the breaking arg change 🙏

PR #1122, opened 2024-06-07
server/user_authorized_keys: Do not hardcode homedir

matthijskooijman:
This PR makes setting authorized_keys for root work by no longer hardcoding `/home/{user}` as a user's homedir. For this, the `server.Home` fact is extended to allow looking up the homedir of other users than the current user too.

See individual commit messages for more details.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -25,12 +25,14 @@ class User(FactBase):
     command = "echo $USER"
 
 
-class Home(FactBase):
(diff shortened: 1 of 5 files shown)


matthijskooijman:
Force-pushed to fix/add testcase for Home fact. user_authorized_keys does not have a testcase yet, and it is a bit too complicated for me to add one now.

Fizzadar (approved):
This is great @matthijskooijman, thank you! Did not know about tilde expansion for other users - super helpful!

PR #1127, opened 2024-06-14
Add ZFS operations and facts

taliaferro:
This MR adds facts and operations for dealing with ZFS filesystems, and tests to cover them.

Most of the logic is in `pyinfra.facts.zfs.Datasets` and `pyinfra.operations.zfs.dataset` ("dataset" being the broad ZFS term for either a filesystem, a block volume or a snapshot) but I've also added convenience functions for each specific type of dataset, to make it a little easier to understand when in use.

### DIFF
--- a/pyinfra/facts/zfs.py
+++ b/pyinfra/facts/zfs.py
@@ -1,0 +1,37 @@
+"""
+Manage ZFS filesystems.
+"""
+
+from pyinfra.api import FactBase
+
+
+class _ZFSResourceBase(FactBase):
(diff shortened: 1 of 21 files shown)


taliaferro:
I've cleared up the linter issues -- is there anything else you'd want changed?

Fizzadar (approved):
Thank you for adding this @taliaferro - this looks perfect. I'd say this is about as perfect a PR as can get for adding new operations/facts to pyinfra :)

PR #1132, opened 2024-06-20
Quote systemd service name

martenlienen:
Mount unit names can contain backslashes, because systemd escapes dashes in file names and requires unit names to match mount points. So a mount unit for `/my-mount` has to be started with `systemctl start my\x2dmount.mount` which fails at the moment.


Fizzadar (approved):
Nice, thanks for fixing this @martenlienen!

PR #1134, opened 2024-06-26
facts/Which: switch to shell built-in

lemmi:
`which` isn't a shell built-in and distributions started deprecating its usage [0].

The recommendation is to use `command -v` for scripts as a replacement.

[0]: https://salsa.debian.org/debian/debianutils/-/blob/5aeca1e3b14c08b5ab92995e91efcc2b3806a639/debian/NEWS#L7-16

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -121,12 +121,12 @@ class Command(FactBase[str]):
 
 class Which(FactBase[Optional[str]]):
     """
-    Returns the path of a given command, if available.
+    Returns the path of a given command according to `command -v`, if available.
     """
 
     @staticmethod
     def command(command):
-        return "which {0} || true".format(command)
+        return "command -v {0} || true".format(command)
 
 
 class Date(FactBase[datetime]):
--- a/tests/facts/server.Which/which.json
+++ b/tests/facts/server.Which/which.json
@@ -1,6 +1,6 @@
 {
     "arg": "python",
-    "command": "which python || true",
+    "command": "command -v python || true",
     "output": ["/usr/local/bin/python"],
     "fact": "/usr/local/bin/python"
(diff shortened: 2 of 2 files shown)


romain-dartigues:
Bump.
In some distributions `which` is not installed by default and, as you said, [`command -v`](https://man7.org/linux/man-pages/man1/command.1p.html) is [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html).

When used inside BASH, it's part of the [builtins](https://man7.org/linux/man-pages/man1/bash.1.html#SHELL_BUILTIN_COMMANDS).

**But**, currently BusyBox (1.36) only implement `which` and **not** `command`: https://busybox.net/downloads/BusyBox.html#commands.

When `command` is not found, BusyBox at least respect the standard and [return 127](https://pubs.opengroup.org/onlinepubs/000095399/utilities/exit.html); maybe an explorer is needed to know beforehand if `command` or `which` is to be used.

lemmi:
I fully expected BusyBox to work. Maybe it's just not documented correctly. This works:
```bash
podman run --rm busybox sh -c "type command"
command is a shell builtin
podman run --rm busybox sh -c "command -v ls"
/bin/ls
```

And I found:

https://git.busybox.net/busybox/tree/shell/ash.c#n8896

romain-dartigues:
My bad, you are right; it's implemented by `sh`, as per POSIX, and not as a BusyBox "applet" (ie: `busybox which busybox` vs `busybox command -v busybox`).

Good news :^)

lemmi:
Ah, this makes sense. Still good to confirm this :)

Fizzadar (approved):
Thanks for fixing this @lemmi, will land in 3.0.1 later today :)

PR #1141, opened 2024-07-06
Documentation: Update using-operations.rst

scoufman:
- Fix commands parameter name in server.shell call.


Fizzadar (approved):
Thanks for fixing this @scoufman!

PR #1147, opened 2024-07-15
Add AppJail connector

DtxdF:
This is the connector for AppJail, a jail management framework for FreeBSD.

More information in its documentation:

* https://appjail.readthedocs.io/

### DIFF
--- a/pyinfra/connectors/appjail.py
+++ b/pyinfra/connectors/appjail.py
@@ -1,0 +1,223 @@
+from __future__ import annotations
+
+import os
+from tempfile import mkstemp
+from typing import TYPE_CHECKING
+
+import click
+from typing_extensions import List, TypedDict, Union, Unpack
+
+from pyinfra import local
+from pyinfra.api import QuoteString, StringCommand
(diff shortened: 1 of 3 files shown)


Fizzadar (commented):
Hi @DtxdF! Firstly - thank you for PR-ing this, it's great to see the connector framework/API being used!

I'm not going to merge this into pyinfra core, however, because I would like to reduce the number of connectors that are maintained within pyinfra itself (I am thinking v4 will only contain `@ssh` and `@docker` and will move all others into their own packages).

If you have time and are willing I'd highly encourage you to turn this into a pypi package (`pyinfra-connector-appjail`?) which could then be linked to on the [connectors index in the docs](https://docs.pyinfra.com/en/3.x/connectors.html).

DtxdF:
@Fizzadar 

No problem, when I have some time I will create a pypi package to use it.

Thanks!

Fizzadar:
@DtxdF awesome, drop me a note when you do (or make an issue here) and I'll get it linked in the connector docs!

PR #1155, opened 2024-07-27
Fix apt sources not being parsed correctly

CondensedTea:
Hello! 

I recently started playing with pyinfra and encountered issue that after installing MongoDB I could not find related data in `AptSources` fact. 

Seems like that issue rooted in pyinfra expecting distribution name to comply with regexp `[a-z-]+`, but after quick research I did not found any requirements that distribution name cannot contain other symbols and as we can see mongodb developers do use those symbols. 

This PR updates regexp in `parse_apt_repo` function to be more permissive to distribution names and adds test for this scenario.

Reproduction of issue:
```python
from pyinfra import host
from pyinfra.operations import apt, server, python
from pyinfra.facts.apt import AptSources


def print_fact():
    print("facts", host.get_fact(AptSources))


apt.update()

apt.packages(
    packages=["curl", "gpg"]
)

server.shell(
    commands=[
        "curl -fsSL https://www.mongodb.org/static/pgp/server-7.0.asc | gpg -o /usr/share/keyrings/mongodb-server-7.0.gpg --dearmor",
        "echo 'deb [ arch=amd64,arm64 signed-by=/usr/share/keyrings/mongodb-server-7.0.gpg ] https://repo.mongodb.org/apt/ubuntu jammy/mongodb-org/7.0 multiverse' | \
         tee /etc/apt/sources.list.d/mongodb-org-7.0.list > /dev/null"
    ]
)

python.call(
    function=print_fact
)
```  

Mongodb docs: https://www.mongodb.com/docs/manual/tutorial/install-mongodb-on-ubuntu/

### DIFF
--- a/pyinfra/facts/apt.py
+++ b/pyinfra/facts/apt.py
@@ -9,7 +9,7 @@ from .util import make_cat_files_command
 
 
 def parse_apt_repo(name):
-    regex = r"^(deb(?:-src)?)(?:\s+\[([^\]]+)\])?\s+([^\s]+)\s+([a-z-]+)\s+([a-z-\s]*)$"
+    regex = r"^(deb(?:-src)?)(?:\s+\[([^\]]+)\])?\s+([^\s]+)\s+([^\s]+)\s+([a-z-\s]*)$"
 
     matches = re.match(regex, name)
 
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Thanks for catching & fixing this @CondensedTea!

PR #1160, opened 2024-07-30
docs/using-operations.rst: fix _if example and extend it

bauen1:
The current example for `_if` doesn't work, I've fixed it and extended it to also cover building "OR" cases, which are probably quite a common occurrence.


Fizzadar (approved):
Thanks for fixing this @bauen1!

PR #1164, opened 2024-08-05
Fix docker container

apecnascimento:
Some users report an error:
`    [@myserver] "docker container inspect" requires at least 1 argument.
    [@myserver] See 'docker container inspect --help'.
    [@myserver] 
    [@myserver] Usage:  docker container inspect [OPTIONS] CONTAINER [CONTAINER...]
    [@local] Display detailed information on one or more containers
    [@local] Error: could not load fact: docker.DockerContainers `

This fact errors occurs if there aren't any containers on the system already. 
This PR change:
-  Use DockerContainer instead DockerContainers
-  Use DockerVolume instead DockerVolumes
-  Use DockerNetwork insead DcokerNetworks

This changes fix the error in docker.container, docker.volume and docker.network operations

### DIFF
--- a/pyinfra/operations/docker.py
+++ b/pyinfra/operations/docker.py
@@ -4,7 +4,7 @@ Manager Docker Containers, Volumes and Networks
 
 from pyinfra import host
 from pyinfra.api import operation
-from pyinfra.facts.docker import DockerContainer, DockerNetworks, DockerVolumes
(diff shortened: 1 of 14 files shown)


Fizzadar (approved):
Thanks for fixing this @apecnascimento! It's actually going to be much more efficient as well instead of loading all the containers/volumes each time :)

PR #1167, opened 2024-08-09
pyinfra/operations: fix crontab comparison

andrew-d:
The Crontab fact tries to convert numerical values from 'str' to 'int', but the crontab operation was doing a direct comparison of the values provided (whether they were strings or ints).

Mimic what the Crontab fact does and use 'try_int' to compare. Also, add a test to confirm this behaviour doesn't regress.


Fizzadar (approved):
Great catch, thanks for fixing this @andrew-d!

PR #1172, opened 2024-08-11
Allow passing arguments to jinja2.Environment() when rendering templates

DonDebonair:
It seems that with the current test setup we can only create unit tests for operations that check the commands that are being yielded by those operations. I don't see an easy option to create a test to verify that the `jinja_env_kwargs` passed into the `files.template()` operation are being passed onto the jinja2 `Environment()`.

If you want to me to write a test for that, you'll have to give me some pointers :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ venv/
 .cache/
 .mypy_cache/
 .pytest_cache/
+.python-version
 
 .vagrant*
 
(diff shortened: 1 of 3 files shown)


DonDebonair:
@Fizzadar what do you need from me in order to merge this PR?

Fizzadar (approved):
Thank you for pulling this together, @DonDebonair, a great add!

PR #1176, opened 2024-08-12
Check if SSH alias exists (#1170)

### LINKED ISSUE #1170: Deploying to SSH aliases is no longer possible
## Describe the bug

In pyinfra 2.7, it was possible to run a deploy to a SSH alias defined in `~/.ssh/config` file, like this:

```sh
pyinfra server-alias deploy.py
```

In pyinfra 3.0.2, this is no longer possible, and fails with:

```
server-alias is neither an inventory file, a (list of) hosts or connectors nor refers to a python module
```

## To Reproduce

Add an alias to your `~/.ssh/config`:

```
Host server-alias
    HostName example.com
    User user
    Port 22
```

Try to deploy to that alias:

```sh
pyinfra server-alias deploy.py
```

## Expected behavior

pyinfra should detect that this is a host alias and use it to connect to the server.

## Meta

+ How was pyinfra installed (source/pip)? poetry

```
    System: Windows
      Platform: Windows-11-10.0.22631-SP0
      Release: 11
      Machine: AMD64
    pyinfra: v3.0.2
    Executable: E:\Projects\GimVicUrnik\venv\Scripts\pyinfra.exe
    Python: 3.12.4 (CPython, MSC v.1940 64 bit (AMD64))
```

simonhammes:
Fixes #1170

### DIFF
--- a/pyinfra_cli/inventory.py
+++ b/pyinfra_cli/inventory.py
@@ -6,6 +6,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union
 
 from pyinfra import logger
 from pyinfra.api.inventory import Inventory
+from pyinfra.connectors.sshuserclient.client import get_ssh_config
 from pyinfra.context import ctx_inventory
 
 from .exceptions import CliError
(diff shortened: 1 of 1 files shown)


pyinfra_cli/inventory.py
```diff
@@ -88,7 +89,33 @@ def _resolves_to_host(maybe_host: str) -> bool:
...
+            socket.getaddrinfo(alias, port=None)
+            return True
+        except socket.gaierror:
+            return False
+
+
+def _get_ssh_alias(maybe_host: str) -> Optional[str]:
+    logger.debug('Checking if "%s" is an SSH alias', maybe_host)
+
+    ssh_config = get_ssh_config()
```
Fizzadar, line 107:
```suggestion
    # Note this does not cover the case where `host.data.ssh_config_file` is used
    ssh_config = get_ssh_config()
```

Fizzadar: Note: I don't have a good suggestion for that right now, so I think let's land this as-is with the comment as it fixes the majority of cases with this issue.

Fizzadar (approved):
One small nit but I think we can just comment that for now, looks great otherwise - thanks @simonhammes!

PR #1177, opened 2024-08-16
systemd: assume localhost if user is given but not machine

bauen1:
Should make the `machine=".host"` redundant in `systemd` facts and operations.


bauen1:
Just tested it to supply `.host` if no `machine` is given, but `user` is given.

Perhaps `user_name` should also imply `user=True`, but that's another can of worms.

Fizzadar (approved):
This seems reasonable to me, thanks for fixing @bauen1!

PR #1179, opened 2024-08-17
apt upgrade change detection

bauen1:
Just a draft for adding proper change detection to `apt-get dist-upgrade` and possibly extending this to cover `apt-get upgrade` too.

Currently based on #1178 

Also there's a note on `/var/lib/apt/periodic/update-stamp` vs. `/var/lib/apt/periodic/update-success-stamp` to be addressed.


docs/using-operations.rst
```diff
@@ -166,6 +166,11 @@ All operations return an operation meta object which provides information about
         _if=lambda: create_user.did_change() or create_otheruser.did_change(),
     )
 
+    # The functions `any_changed` and `all_changed` are provided for common use cases, e.g.
```
Fizzadar, line 169: This is an excellent addition here!

pyinfra/facts/apt.py
```diff
@@ -7,6 +7,23 @@
...
+        (
+            '-o Dpkg::Options::="--force-confdef"',
+            '-o Dpkg::Options::="--force-confold"',
+            command,
+        ),
+    )
+
+    return " ".join(args)
+
+APT_CHANGES_RE = re.compile(r"^(\d+) upgraded, (\d+) newly installed, (\d+) to remove and (\d+) not upgraded.$")
```
Fizzadar: Interesting! This is not an approach I'd thought of using at all, I like it.

Fizzadar (commented):
This is really neat @bauen1!

pyinfra/facts/apt.py
bauen1: And this line should also be present when executing e.g. `apt-get autoremove`, hence my question https://matrix.to/#/!dPvLqxqwbQZYeQrNnu:matrix.org/$ygUK_H5ACvxHxmEbKiwfSBsCQsoiTINm9WqBl5tSatE?via=beeper.com&via=matrix.org&via=t2bot.io if there was a way for operations to generate a command and then decide, based on the output, if any change was actually performed.
That could then also apply to `apt-get update` and then almost all apt operations could support proper change detection.

bauen1:
Changes for `apt-get upgrade` are untested, but you get the idea.

bauen1:
@Fizzadar The operations unit tests `FakeHost.get_fact` does not actually support positional arguments, how can I fix that ?

pyinfra/operations/apt.py
```diff
@@ -21,21 +27,18 @@
 APT_UPDATE_FILENAME = "/var/lib/apt/periodic/update-success-stamp"
 
 
-def noninteractive_apt(command: str, force=False):
-    args = ["DEBIAN_FRONTEND=noninteractive apt-get -y"]
-
-    if force:
-        args.append("--force-yes")
+def _simulate_then_perform(command: str):
+    changes = host.get_fact(SimulateOperationWillChange, command)
```
maisim, line 31: Regarding your question that you asked on matrix, I think you should do:
 changes = host.get_fact(SimulateOperationWillChange, command=command)

bauen1: My problem is, that the `get_fact` method of the `FakeHost` of the framework does not support positional arguments.
But there are already other facts that use positional arguments, so this seems to be an oversight :thinking:

maisim: Are you sure ? it seems to me that only the name of the Fact is a positional argument, all the others have to be named arguments, let me check that

An example where get_fact is called with positional arguments?

bauen1: Oh, that's a good point, I only though about the definiton of the `command` method using positional arguments, e.g. for `FindInFile`:

```python
def command(self, path, pattern, interpolate_variables=False):
```

But even that is called using a keyword argument:

```python
pyinfra/operations/server.py
19:from pyinfra.facts.files import Directory, FindInFile, Link
1193:        FindInFile, path=locales_definitions_file, pattern=rf"^.*{locale}[[:space:]]\+.*$"
```

So, maybe that's an oversight or some quirk that should be documented ?

In the meantime I added some rudimentary support (or rather, ignorance of) positional arguments for `get_fact`

bauen1: And the real `get_fact` method does support passing in positional arguments, but then apparently does some wild magic to "merge" positional and keyword arguments: 

https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra/api/facts.py#L121-L143

maisim: You're right, it is possible, just not used :)
You're fix in tests/utils.py seems justified :+1:

bauen1:
So, I've "fixed" the test, but this also kind of breaks existing tests, since the facts kwarg string is now different.
Actually, it doesn't even break all tests, e.g. https://github.com/pyinfra-dev/pyinfra/pull/1179/commits/c884c43370c015d17bd5fb0ed70c13dabc427550 doesn't change the outcome.

One of the reasons is, that missing fact entries currently don't raise an error or even warning, so I would propose https://github.com/pyinfra-dev/pyinfra/pull/1179/commits/b964fe557e9899674b9f66373042589fe4b20b70 or even better, make this a fatal error for the test case.


Additionally, the `idempotent` and `disable_idempotent_warning` fields should be cleaned up like https://github.com/pyinfra-dev/pyinfra/pull/1179/commits/6040d3449684dec85b9919decf2f1e3cab66dd52 - I will try to split this cleanup into it's own MR to keep this one simpler...

tests/util.py
```diff
@@ -194,23 +195,33 @@ def _get_fact_key(fact_cls):
...
             ), f"Argument {key} is not a real argument in the `{fact_cls}.command` method"
 
-    def get_fact(self, fact_cls, **kwargs):
+    def get_fact(self, fact_cls, *args, **kwargs):
         fact_key = self._get_fact_key(fact_cls)
         fact = getattr(self.fact, fact_key, None)
         if fact is None:
             raise KeyError("Missing test fact data: {0}".format(fact_key))
+
+        # This does the same thing that pyinfra.apit.facts._handle_fact_kwargs does
```
Fizzadar, line 210: Nice, good to have this be the same even if it is a bit weird!

pyinfra/operations/apt.py
Fizzadar: Agreed this is a good change! The facts API is... old and messy, unfortunately hard to fix it without breaking it.

Fizzadar (approved):
This is awesome, thank you for this @bauen1 and sorry for taking so long!

I totally agree with your comment re: test keys missing. I was going to leave it as a warning but actually it seems like a good chance to fix this properly. I've pushed a quick fix for the failing windows test but will follow (today, if time allows) with updated keys in all the tests.

pyinfra/facts/apt.py
```diff
@@ -94,3 +118,39 @@ def command(self) -> str:
...
+class AptSimulationDict(TypedDict):
+    upgraded: int
+    newly_installed: int
+    removed: int
+    not_upgraded: int
+
+
+class SimulateOperationWillChange(FactBase[AptSimulationDict]):
+    """
+    Simulate an 'apt-get' operation and try to detect if any changes would be performed.
```
Fizzadar, line 132: Didn't mention in my review, this is a very cool trick, love it

pyinfra/operations/apt.py
bauen1: > Agreed this is a good change! The facts API is... old and messy, unfortunately hard to fix it without breaking it.

I mean, if you ever do decide to change it (in a probably breaking way), one of the things that's been bothering me is, that while you usually use operations in a "namespace preserving way" i.e. `files.put` you don't do the same with Facts, i.e. `FileExists` ...

pyinfra/facts/apt.py
bauen1: Technically speaking, this primarily works around the inability to tell pyinfra if a command run actually changed something.

Ideally I'd run the `apt-get` command and additionally inform pyinfra after it run if it actually changed something.

But glad you like it, and it does improve my update scripts a bit by avoiding unnecessary reboots :)

PR #1180, opened 2024-08-24
A few additional type hints

bauen1:
Not a 100% sure if they are right, but they seem to work.

### DIFF
--- a/pyinfra/api/host.py
+++ b/pyinfra/api/host.py
@@ -218,17 +218,19 @@ class Host:
             self.print_prefix_padding,
         )
 
-    def log(self, message, log_func=logger.info):
+    def log(self, message: str, log_func: Callable[[str], Any] = logger.info) -> None:
         log_func(f"{self.print_prefix}{message}")
 
-    def log_styled(self, message, log_func=logger.info, **kwargs):
+    def log_styled(
+        self, message: str, log_func: Callable[[str], Any] = logger.info, **kwargs
+    ) -> None:
         message_styled = click.style(message, **kwargs)
         self.log(message_styled, log_func=log_func)
 
     def get_deploy_data(self):
         return self.current_op_deploy_data or self.current_deploy_data or {}
 
-    def noop(self, description):
+    def noop(self, description: str) -> None:
         """
         Log a description for a noop operation.
         """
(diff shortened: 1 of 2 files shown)


pyinfra/api/host.py
```diff
@@ -218,17 +218,18 @@ def style_print_prefix(self, *args, **kwargs) -> str:
             self.print_prefix_padding,
         )
 
-    def log(self, message, log_func=logger.info):
+    def log(self, message: object, log_func: Callable[[str], Any] = logger.info) -> None:
```
jankatins: Shouldn't message be a str? (if something is not a str, I think it should be stringified `str(...)` on the caller side...)

pyinfra/api/host.py
```diff
@@ -218,17 +218,18 @@ def style_print_prefix(self, *args, **kwargs) -> str:
...
+    def log_styled(self, message: object, log_func: Callable[[str], Any] = logger.info, **kwargs
+    ) -> None:
         message_styled = click.style(message, **kwargs)
         self.log(message_styled, log_func=log_func)
 
     def get_deploy_data(self):
         return self.current_op_deploy_data or self.current_deploy_data or {}
 
-    def noop(self, description):
+    def noop(self, description: object) -> None:
```
jankatins: Same here:  a description should be a str...

pyinfra/api/host.py
```diff
@@ -218,17 +218,18 @@ def style_print_prefix(self, *args, **kwargs) -> str:
             self.print_prefix_padding,
         )
 
-    def log(self, message, log_func=logger.info):
+    def log(self, message: object, log_func: Callable[[str], Any] = logger.info) -> None:
         log_func(f"{self.print_prefix}{message}")
 
-    def log_styled(self, message, log_func=logger.info, **kwargs):
+    def log_styled(self, message: object, log_func: Callable[[str], Any] = logger.info, **kwargs
```
jankatins: Same here: a description should be a str...

pyinfra/api/host.py
bauen1: I actually used `object` since it's what's used in https://github.com/python/typeshed/blob/main/stdlib/logging/__init__.pyi#L115

bauen1:
I've fixed all the linting errors and changed `object` to `str` instead.

Fizzadar (approved):
Nice, thanks @bauen1!

PR #1181, opened 2024-08-25
Fix sphinx warnings

simonhammes:


Fizzadar (approved):
Amazing, thank you for all these fixes @simonhammes!

Issue #662, opened 2021-09-22
Syntactically correct python in inventory file fails if not matching inventory specification


themanifold:
If you specify a syntactically python inventory file, but also one that does not match what pyinfra requires in its inventory spec, you'll see no error which is confusing.

For example, an inventory file (called `bad_inventories.py`) can look like this:

```
app_server = [
    'a',
    'b', ({"foo": "bar"})    
]
``` 
which is incorrect as the bracket for specifying data associated with the host `'b'` is in the wrong place so that there's a third host that is a tuple.

If you run the above snippet with pyinfra as an inventory file, you'll see:

```
$ pyinfra --debug -vvvv inventories/bad_inventory.py debug-inventory
    [pyinfra_cli.main] Checking potential directory: inventories
    [pyinfra_cli.main] Deploy directory remains as cwd
--> Loading config...
--> Loading inventory...
    [pyinfra_cli.inventory] Creating fake inventory...
```
Obviously, the inventory file should look like this:
```
app_server = [
    'a',
    ('b', {"foo": "bar"})    
]
``` 
but we should protect the users by having some more validation on the inventory file format. If there's a subset of python we accept, then it needs validation.

Fizzadar:
This is super important, a real gotcha without any validation currently.

simonhammes:
@Fizzadar This can also be closed

PR #1182, opened 2024-08-25
Validate inventory file (#662)

simonhammes:


pyinfra_cli/inventory.py
```diff
@@ -258,7 +270,7 @@ def make_inventory_from_files(
         for hosts in groups.values():
             # Groups can be a list of hosts or tuple of (hosts, data)
             hosts = _get_any_tuple_first(hosts)
-
+            # TODO: Generator expressions are silently discarded (?)
```
simonhammes: I'll take a look at this.

pyinfra_cli/inventory.py
```diff
@@ -23,18 +23,30 @@ def _is_inventory_group(key: str, value: Any):
...
+    elif isinstance(value, GeneratorType):
+        # Expand any generators of hosts
         value = list(value)
+    else:
+        logger.debug('Ignoring variable "%s" in inventory file since it is not a list, tuple or generator expression', key)
+        return False
 
-    return all(isinstance(item, ALLOWED_HOST_TYPES) for item in value)
+    if not all(isinstance(item, ALLOWED_HOST_TYPES) for item in value):
+        logger.warning('Ignoring host group "%s". Host groups may only contain strings (host) or tuples (host, data).', key)
```
simonhammes: Should this be an error instead? 

On one hand, an inventory file might contain other data structures since you can do whatever you want in an inventory file (similar to dynamic inventory in Ansible), so even this warning might be too verbose?

On the other hand, _silently_ ignoring variables due to invalid types is also problematic (see #662).

Any thoughts?

Fizzadar: I think warning is the correct behavior for now - don't want to break anything existing but also need the warning so at least people will know!

Fizzadar (commented):
This looks great so far, thank you @simonhammes!

I think long term pyinfra should be moving towards [inventory functions](https://github.com/pyinfra-dev/pyinfra-examples/tree/main/inventory-functions) as the default to avoid the weird "list or tuple" thing for hosts/groups.

There's also potential for a DSL kind of inventory, but that's a future problem :)

pyinfra_cli/inventory.py
simonhammes: Generator expressions do not work (same behavior as on `3.x`):

<details><summary>Screenshot</summary>

image

</details>

simonhammes:
> This looks great so far, thank you @simonhammes!

:pray: 

> I think long term pyinfra should be moving towards [inventory functions](https://github.com/pyinfra-dev/pyinfra-examples/tree/main/inventory-functions) as the default to avoid the weird "list or tuple" thing for hosts/groups.

That seems like a wise decision imho.   
Should I even bother to fix the generator support (see the open thread in this PR)? Or just update the documentation?

EDIT: I'll also add some tests.

Fizzadar:
> That seems like a wise decision imho. Should I even bother to fix the generator support (see the open thread in this PR)? Or just update the documentation?

Let's just update the docs, would rather not increase complexity of the current situation!

Fizzadar (commented):
Change looks great so far, tests would tie this up nicely!

simonhammes:
> > That seems like a wise decision imho. Should I even bother to fix the generator support (see the open thread in this PR)? Or just update the documentation?
> 
> Let's just update the docs, would rather not increase complexity of the current situation!

Thanks for the feedback, I have updated the docs.

> Change looks great so far, tests would tie this up nicely!

Thank you! I agree, I'll add some tests.

simonhammes:
Let me know if you had something else in mind re: tests :sweat_smile:

tests/test_cli/inventories/invalid.py
simonhammes, line 1: Did not know where to put this - maybe into `deploy/inventories`?

Fizzadar: This seems good as-is, there's never a correct answer when placing files 😆

Fizzadar (approved):
Awesome stuff, @simonhammes, thank you!

PR #1185, opened 2024-08-27
pyinfra/operations: respect `pull` value in git.worktree

Pirols:


Fizzadar (approved):
Neat, thanks for fixing this @Pirols!

PR #1186, opened 2024-08-27
pyinfra/operations: `git worktree add --force` support

Pirols:
-- Why is this needed?

Current scenario:

1. During pyinfra deployment: `git worktree add /path/to/wk commitish`
2. `rm -f /path/to/wk` happens
3. Deployment is run again, however `git worktree add /path/to/wk commitish` fails:

   ``` fatal: '../wk' is a missing but already registered worktree; use 'add -f' to override, or 'prune' or 'remove' to clear ```

   Requiring manual intervention: `git worktree prune` and even `git branch -d wk` if the branch was created.

Solution:

Use `-f`/`--force` (i.e. `force`) to avoid the failure entirely.


Fizzadar (approved):
Great add @Pirols, thank you!

PR #1188, opened 2024-08-27
Support more arguments in FindFilesBase

### LINKED ISSUE #1174: Make get_fact(FindFiles) more versatile
Many times, I want to filter certain kinds of files, yet find can only show if a file exists. This could also be a performance issue when the number of files are larger.

I would like to add a few arguments to Find.

Below are some maybe useful arguments I copied from `man find`. One shall also be able to pass custom arguments to get_fact(Find). 

Pay special attention to `-size n[ckMGTP]`. `1k/M/G/T/P` has pitfall and behave like all non-empty files. one should always use n c(chars) to work around 

```
     -depth  Always true; same as the non-portable -d option.  Cause find to perform a
             depth-first traversal, i.e., directories are visited in post-order and all
             entries in a directory will be acted on before the directory itself.  By
             default, find visits directories in pre-order, i.e., before their contents.
             Note, the default is not a breadth-first traversal.

             The -depth primary can be useful when find is used with cpio(1) to process
             files that are contained in directories with unusual permissions.  It
             ensures that you have write permission while you are placing files in a
             directory, then sets the directory's permissions as the last thing.

     -depth n
             True if the depth of the file relative to the starting point of the
             traversal is n.

     -empty  True if the current file or directory is empty.
     -fstype type
             True if the file is contained in a file system of type type.  The lsvfs(1)
             command can be used to find out the types of file systems that are
             available on the system.  In addition, there are two pseudo-types, “local”
             and “rdonly”.  The former matches any file system physically mounted on the
             system where the find is being executed and the latter matches any file
             system which is mounted read-only.

     -gid gname
             The same thing as -group gname for compatibility with GNU find.  GNU find
             imposes a restriction that gname is numeric, while find does not.

     -group gname
             True if the file belongs to the group gname.  If gname is numeric and there
             is no such group name, then gname is treated as a group ID.

     -ignore_readdir_race
             Ignore errors because a file or a directory is deleted after reading the
             name from a directory.  This option does not affect errors occurring on
             starting points.

     -ilname pattern
             Like -lname, but the match is case insensitive.  This is a GNU find
             extension.

     -iname pattern
             Like -name, but the match is case insensitive.

     -name pattern
             True if the last component of the pathname being examined matches pattern.
             Special shell pattern matching characters (“[”, “]”, “*”, and “?”) may be
             used as part of pattern.  These characters may be matched explicitly by
             escaping them with a backslash (“\”).

     -regex pattern
             True if the whole path of the file matches pattern using regular
             expression.  To match a file named “./foo/xyzzy”, you can use the regular
             expression “.*/[xyz]*” or “.*/foo/.*”, but not “xyzzy” or “/foo/”.

     -samefile name
             True if the file is a hard link to name.  If the command option -L is
             specified, it is also true if the file is a symbolic link and points to
             name.

     -size n[ckMGTP]
             True if the file's size, rounded up, in 512-byte blocks is n.  If n is
             followed by a c, then the primary is true if the file's size is n bytes
             (characters).  Similarly if n is followed by a scale indicator then the
             file's size is compared to n scaled as:

             k       kilobytes (1024 bytes)
             M       megabytes (1024 kilobytes)
             G       gigabytes (1024 megabytes)
             T       terabytes (1024 gigabytes)
             P       petabytes (1024 terabytes)

```

JakkuSakura:
This is an attempt to make FindFiles more useful. 

~It has not been tested yet~

closes #1174

### DIFF
--- a/pyinfra/facts/util/units.py
+++ b/pyinfra/facts/util/units.py
@@ -1,29 +1,59 @@
+# from https://stackoverflow.com/a/60708339, but with a few modifications
+from __future__ import annotations  # for | in type hints
+
 import re
 
-units = {"B": 1, "KB": 10 ** 3, "MB": 10 ** 6, "GB": 10 ** 9, "TB": 10 ** 12,
-         "KIB": 2 ** 10, "MIB": 2 ** 20, "GIB": 2 ** 30, "TIB": 2 ** 40}
+units = {
+    "B": 1,
+    "KB": 10**3,
+    "MB": 10**6,
+    "GB": 10**9,
+    "TB": 10**12,
+    "KIB": 2**10,
+    "MIB": 2**20,
+    "GIB": 2**30,
+    "TIB": 2**40,
+}
 
 
 def parse_human_readable_size(size: str) -> int:
     size = size.upper()
-    # print("parsing size ", size)
-    if not re.match(r' ', size):
-        size = re.sub(r'([KMGT]?I?[B])', r' \1', size)
+    if not re.match(r" ", size):
+        size = re.sub(r"([KMGT]?I?[B])", r" \1", size)
     number, unit = [string.strip() for string in size.split()]
     return int(float(number) * units[unit])
+
 
 def parse_size(size: str | int) -> int:
     if isinstance(size, int):
         return size
     return parse_human_readable_size(size)
+
+
 def test_parse_human_readable_size():
     example_strings = [
-        "1024b", "10.43 KB", "11 GB", "343.1 MB",
-        "10.43KB", "11GB", "343.1MB", "10.43 kb",
-        "11 gb", "343.1 mb", "10.43kb", "11gb",
-        "343.1mb", "1024Kib", "10.43 KiB", "11 GiB",
-        "343.1 MiB", "10.43KiB", "11GiB", "343.1MiB",
-        "10.43 kib", "11 gib"
+        "1024b",
+        "10.43 KB",
+        "11 GB",
+        "343.1 MB",
+        "10.43KB",
+        "11GB",
+        "343.1MB",
+        "10.43 kb",
+        "11 gb",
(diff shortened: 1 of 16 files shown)


Fizzadar (commented):
This looks great, huge improvement! Few linting things to fix and this is ready I think.

JakkuSakura:
Just fixed linting.
one notable change is `fname`
mypy caught that `name` is defined in FactBase, thus we can't use `name` directly

JakkuSakura:
I'm not sure how to express `args: List[str]` especially it may contain QuoteString

JakkuSakura:
finally...
I also managed to enhance the tests to support dict as argument.

pyinfra/facts/util/units.py
```diff
@@ -0,0 +1,59 @@
...
+    return int(float(number) * units[unit])
+
+
+def parse_size(size: str | int) -> int:
+    if isinstance(size, int):
+        return size
+    return parse_human_readable_size(size)
+
+
+def test_parse_human_readable_size():
```
Fizzadar: This should be moved somewhere under `tests/` so it gets covered in test runs.

Fizzadar (approved):
Nice! Just the test thing (I'll move and merge if I get to it before you :)).

Fizzadar (approved):
@JakkuSakura this is really awesome, thank you! Sorry it's taken me so long to review it. I've pushed up a quick fix for the tests (also made a mess on rebase first, but fixed).

JakkuSakura:
Much appreciated! It's my first PR for pyinfra and the lining has cost me some time, but eventually figured out how to do it properly

Fizzadar:
> Much appreciated! It's my first PR for pyinfra and the lining has cost me some time, but eventually figured out how to do it properly

Would love any comments/ideas/etc on how to make the experience better when contributing 🙏

PR #1199, opened 2024-09-06
AptSources now returns components as list instead of set to make updates possible

### LINKED ISSUE #1198: AptSources uses set for components
## Describe the bug
AptSources uses a set for the list of components.
This makes removals of existing entries with at.repo() nearly impossible.

## To Reproduce
sources.list:
`deb [arch=amd64] https://deb.debian.org/debian/ bullseye main contrib non-free`

code:
`sources = host.get_fact(AptSources)`
components returned in this order: contrib non-free main

## Expected behavior
The components should be a list in this order: main contrib non-free

I'm currently working on a patch and will add an pull request.

rsfzi:
resolves #1198

### DIFF
--- a/pyinfra/facts/apt.py
+++ b/pyinfra/facts/apt.py
@@ -7,7 +7,6 @@ from pyinfra.api import FactBase
 from .gpg import GpgFactBase
 from .util import make_cat_files_command
 
-
 def parse_apt_repo(name):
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
Ah I see to enable ordering them - this makes sense, thanks for the PR @rsfzi!

Issue #1208, opened 2024-09-18
pyinfra.operations.files.get raises UnicodeDecodeError when attempting to get a file that is not utf8 encoded


evoldstad:
## Describe the bug
pyinfra.operations.files.get fails with an exception when attempting to get a file that is not utf-8 encoded.  

## To Reproduce
Added an example file that demonstrates the issue by cloning and building ipxe:  [example.zip](https://github.com/user-attachments/files/17040300/example.zip)
Had to zip it due to github not allowing .py files. 

Run it with: `pyinfra @local encoding-bug-example.py`

## Expected behavior

The behavior of files.get is the same regardless of encoding, or documentation clearly explains that the operation only supports utf-8 encoded files.  

## Meta
+ Include output of `pyinfra --support`.
    System: Linux
      Platform: Linux-6.10.8-1-default-x86_64-with-glibc2.40
      Release: 6.10.8-1-default
      Machine: x86_64
    pyinfra: v3.1
      black: v24.4.2
      black: v24.4.2
      click: v8.1.7
      configparser: v7.1.0
      coverage: v7.5.1
      coverage: v7.5.1
      distro: v1.9.0
      flake8: v7.0.0
      flake8: v7.0.0
      flake8-black: v0.3.6
      flake8-black: v0.3.6
      flake8-isort: v6.1.1
      flake8-isort: v6.1.1
      flake8-spellcheck: v0.12.1
      gevent: v24.2.1
      importlib-metadata: v8.5.0
      ipdb: v0.13.13
      ipdbplugin: v1.5.0
      ipython: v8.27.0
      isort: v5.13.2
      isort: v5.13.2
      jinja2: v3.1.4
      mypy: v1.11.2
      mypy: v1.11.2
      myst-parser: v2.0.0
      myst-parser: v2.0.0
      packaging: v24.1
      paramiko: v3.5.0
      pyinfra-guzzle-sphinx-theme: v0.16
      pyinfra-guzzle-sphinx-theme: v0.16
      pytest: v8.2.1
      pytest: v8.2.1
      pytest-cov: v5.0.0
      pytest-cov: v5.0.0
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      redbaron: v0.9.2
      setuptools: v65.5.0
      sphinx: v6.2.1
      sphinx: v6.2.1
      twine: v5.1.1
      typeguard: v4.3.0
      types-PyYAML: v6.0.12.20240917
[...]

PR #1214, opened 2024-09-23
pyinfra/connectors: Fixed broken get for non-utf8 files (#1208)

evoldstad:
A small fix that allows for using the "get" operator on files with different encoding than utf8, which currently results in a UnicodeDecodeError. This fix should allow the get operation to be more encoding-agnostic. 

There is an example project in #1208 that demonstrates the currently broken behavior.


Fizzadar (approved):
Oh nice catch, thanks for fixing @evoldstad!

PR #1215, opened 2024-09-26
Add psql_database argument postgres operations/facts

hamishfagg:
This PR adds the ability to specify which database `psql` should connect to for postgres/postgresql operaitons and facts.

I can into an issue where I wanted to create a postgres role `my_role` with `CREATEDB`, and then connect as that new user to create a database that they owned. Because postgres defaults to connecting to the db with the same name as the role, I would get an error like: `FATAL:  database "my_role" does not exist`.

This method is necessary when using managed postgres services from cloud providers that don't provide superuser access, meaning you can't change the owner of a database.

### DIFF
--- a/pyinfra/facts/postgres.py
+++ b/pyinfra/facts/postgres.py
@@ -58,6 +58,7 @@ class PostgresFactBase(FactBase):
         psql_password=None,
         psql_host=None,
         psql_port=None,
+        psql_database=None,
     ):
         return make_execute_psql_command(
             self.psql_command,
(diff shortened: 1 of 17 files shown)


Fizzadar (approved):
Hey @hamishfagg! Thank you for putting this together - this looks great! Sorry it took a while to get to it, I'll get this released shortly :)

hamishfagg:
Hey no problem, thanks for getting to it!

hamishfagg:
Hey @Fizzadar sorry to bother you about this again, do you have an ETA for this to be released?

Thanks

Fizzadar:
@hamishfagg no worries at all, thank you for reminding me! 3.2 will be live in about 5 minutes :)

PR #1221, opened 2024-10-17
Read group and user information using getent to respect nsswitch.conf

### LINKED ISSUE #1008: server.Users and related does not respect nsswitch.conf
## Describe the bug

Using the `server.Users` fact (and related, such as `server.Groups`) does not respect the configuration of `/etc/nsswitch.conf`, e.g. when LDAP is configured.

## To Reproduce

Configure e.g. LDAP/sssd in `/etc/nsswitch.conf` or `libnss_systemd` for the `passwd` database and attempt to query a user from that source using the `server.Users` fact.
Since the `server.Users` and `server.Groups` fact only looks at `/etc/passwd` and `/etc/groups` instead of using e.g. libnss these user entries are not returned.

If necessary, I can provide a more detailed example, but I believe the main point comes across.

## Expected behavior

Enumerate and return users/groups according to the `/etc/nsswitch.conf` configuration.

## Meta

```
--> Support information:

    System: Linux
      Platform: Linux-6.1.0-9-amd64-x86_64-with-glibc2.37
      Release: 6.1.0-9-amd64
      Machine: x86_64
    pyinfra: v2.7
    Executable: /opt/venv/bin/pyinfra
    Python: 3.11.4 (CPython, GCC 12.2.0)
```

## Suggested Solution

Use `getent` to query the `passwd` and `groups` database, but fall back to `/etc/passwd` and `/etc/group` in the unlikely case that `getent` is not provided, e.g. on a busybox system.

Some additional ideas:

If `getent` is used, it could also be helpful to allow the user to provide the `--service` argument to `getent`.

Perhaps the `server.User` fact could be extended to not just return the name of the current user, but to also include the same information as `server.Users` provides and to allow specifying a different user to return information about.
And add a new fact `server.Group` that does the same.

The Name Service Switch has a few other database that might be interesting to implement, e.g. `aliases`, `ethers`, `hosts`, `initgroups`, `networks`, `protocols`, `rpc`, `services` and some special cases `sudoers`, `subid`.

If such a change would be welcome, I could look into implementing these suggestions.

Thanks for this awesome tool :heart:

bauen1:
This will fix #1008 .

Additionally I messed around with adding a `Group` fact, to get more useful information about e.g. members of a group.

Draft, since I need to fix the user stuff and unit tests, this is just here so I don't forget about it.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -441,11 +441,13 @@ class Users(FactBase):
         }
     """
 
-    def command(self):
-        return """
+    def requires_command(self) -> str:
+        return "getent"
 
-        for i in `cat /etc/passwd | cut -d: -f1`; do
-            ENTRY=`grep ^$i: /etc/passwd`;
+    def command(self) -> str:
+        return """
+        for i in `getent passwd | cut -d: -f1`; do
+            ENTRY=`getent passwd | grep ^$i:`;
             LASTLOG_RAW=`(lastlog -u $i 2> /dev/null || lastlogin $i 2> /dev/null)`;
             LASTLOG=`echo $LASTLOG_RAW | grep ^$i | tr -s ' '`;
             PASSWORD=`grep ^$i: /etc/shadow | cut -d: -f2`;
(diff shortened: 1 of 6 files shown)


Fizzadar:
Closing as stale, happy to review a rebased PR!

PR #1222, opened 2024-10-18
Parse os-release for Servers Incompatible with lsb_release

wowi42:
# Summary

This PR adds support for extracting OS info from servers that don't support `lsb_release` by parsing the `os-release` file.

# Benefits

- Ensures compatibility with lightweight distributions that lack `lsb_release`.
- Provides more consistent OS information across different environments.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -307,6 +307,43 @@ class LsbRelease(FactBase):
         return items
 
 
+class OsRelease(FactBase):
+    """
+    Returns a dictionary of release information stored in ``/etc/os-release``.
+
+    .. code:: python
+
+        {
+          "name": "EndeavourOS",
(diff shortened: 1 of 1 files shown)


pyinfra/facts/server.py
```diff
@@ -307,6 +307,43 @@ def process(self, output):
...
+
+        for line in output:
+            if "=" not in line:
+                continue
+
+            key, value = line.split("=", 1)
+            key = key.strip().lower()
+
+            value = value.strip()
+            value = value.strip('"')
```
simonhammes: Nitpick: You could chain these two calls to `strip()`

wowi42: I did even better :-)

Fizzadar (approved):
Thanks for the PR @wowi42! Apologies it took so long to get merged int

PR #1225, opened 2024-10-22
Grand total for hosts, success, error and no change in results output

NichtJens:
Often, I would like to know if there was any error in a long deployment, or the expected amount of successes/no changes.

For this, a grand total line at the end of the results output would be very useful.

This PR adds such a final line:
```
Operation                                  Hosts   Success   Error   No Change 
[...]
Grand total                                23      5         -       99
```


Fizzadar (approved):
This is really nice @NichtJens! Thank you for the PR!

PR #1229, opened 2024-10-30
CI: Use macos-13 instead of macos-12

simonhammes:
The image is deprecated: https://github.com/actions/runner-images/issues/10721

There will be _brownouts_ (failed builds) in November and it is destined to be removed in early December.


Fizzadar (approved):
Thanks @simonhammes!

PR #1230, opened 2024-10-30
Bump dev dependencies

simonhammes:


Fizzadar (approved):
Thanks @simonhammes! I'm going to tweak the compatibility policy slightly to be non-EOL Python versions so we can drop EOL Python support without bumping the major version each time. For now this is great 🙏!

simonhammes:
> Thanks @simonhammes! I'm going to tweak the compatibility policy slightly to be non-EOL Python versions so we can drop EOL Python support without bumping the major version each time. For now this is great 🙏!

Thanks, that makes sense!

PR #1237, opened 2024-11-06
pyinfra/operations: Handle latest value in installing pip requirements file

amiraliakbari:
Currently in the pip.packages operation, the `latest` parameter is only handled when packages argument is passed. This PR adds `--upgrade` to the final command when only a requirements file is specified.


Fizzadar (approved):
Nice! Thank you for the PR @amiraliakbari!

PR #1243, opened 2024-11-21
Add some pipx operations & facts

maisim:
Minimal pipx support

This one replace the previous PR #1039, wich was for pyinfra 2

HI @Fizzadar  I took into account your previous review in this implementation


Fizzadar (approved):
Thank you for implementing this @maisim!

PR #1244, opened 2024-11-22
Add informations on cli output about module when fact or operation was not found

maisim:
Hello folks,

If the operation or fact was not found, it will display the ones available, and if no one is available, Warn the user about the possibility that he have a py file or a folder with the same name as the module

If you have a directory wich have the same name as the pyinfra module you want to use, pyinfra will try to look inside form the operations/facts (wich seems to be the wanted behavior)

So I added a message to inform about this possibility if the module/fact is not found:
```
find . -name incus
./incus
pyinfra inventory.py --debug --limit=lx1 fact incus.IncusProfiles
--> pyinfra error: No such attribute in module incus: IncusProfiles
No facts found. Maybe you have a file or folder named `incus` in the current folder ?
```

In the same time I added this:

```
pyinfra inventory.py --debug --limit=lx1 fact incus.IncusProfilesNOEXIST
--> pyinfra error: No such attribute in module incus: IncusProfilesNOEXIST
Available facts in module are: IncusInstances, IncusProfiles, IncusTestFact
```

This is a suggestion, tell me what you think.

### DIFF
--- a/tests/test_cli/test_cli_exceptions.py
+++ b/tests/test_cli/test_cli_exceptions.py
@@ -1,3 +1,4 @@
+import re
 import sys
 from os import path
 from unittest import TestCase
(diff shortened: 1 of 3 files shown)


tests/test_cli/test_cli_exceptions.py
```diff
@@ -44,7 +44,13 @@ def test_no_fact_module(self):
     def test_no_fact_cls(self):
         self.assert_cli_exception(
             ["my-server.net", "fact", "server.NotAFact"],
-            "No such attribute in module server: NotAFact",
+            (
+                "No such attribute in module server: NotAFact\n"
+                "Available facts in module are: User, Home, Path, TmpDir, Hostname, Kernel, "
```
maisim: Each update will implies an updtate of the test, maybe use a regex slould be better?

maisim:
It does not work as expected for operations in case of misspelling

Fizzadar (commented):
@maisim this is awesome! Really big QoL improvement.

> It does not work as expected for operations in case of misspelling

What do you mean by this? I'll have a play around with the branch when I get the chance, just wondering.

Fizzadar:
Closing as stale but more than happy to rereview a fresh PR, the idea here is great.

PR #1247, opened 2024-11-26
Add support for SHA384 hash in files.download()

simonhammes:
- [x] Add tests for operation

### DIFF
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -276,6 +276,12 @@ class Sha256File(HashFileFactBase, digits=64, cmds=["sha256sum", "shasum -a 256"
     """
 
 
+class Sha384File(HashFileFactBase, digits=96, cmds=["sha384sum", "shasum -a 384"]):
+    """
+    Returns a SHA384 hash of a file, or ``None`` if the file does not exist.
+    """
+
+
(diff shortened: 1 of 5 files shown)


Fizzadar (approved):
This looks good to go I think @simonhammes? I see tests 🙌

pyinfra/facts/files.py
```diff
@@ -276,6 +276,12 @@ class Sha256File(HashFileFactBase, digits=64, cmds=["sha256sum", "shasum -a 256"
     """
 
 
+class Sha384File(HashFileFactBase, digits=96, cmds=["sha384sum", "shasum -a 384"]):
```
simonhammes: I just remembered why I left this PR as a draft :slightly_smiling_face: 

I was unsure whether to add `sha384` to `cmds` - the other facts have the respective executables. But the only executables called `shaXXX` that I could find are symlinks registered by `hashalot`, and they seem to do a different thing...

@Fizzadar Any idea?

Fizzadar: The non-sum ending variants are a macOS (possibly BSD) thing! By default macOS has `sha384` but not `sha384sum`, both are identical in behavior. So yes let's add them in 👍

simonhammes: Thank you! I've added `sha384`.

Fizzadar (approved):
Thank you, @simonhammes!

Issue #1209, opened 2024-09-19
Known_hosts file loses content when adding new host


pikeas:
## Describe the bug

Pyinfra deletes lines from `~/.ssh/known_hosts`.

## To Reproduce

https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra/connectors/sshuserclient/client.py#L43-L49:

```
with HOST_KEYS_LOCK:
    host_keys = client.get_host_keys()
    host_keys.add(hostname, key.get_name(), key)
    # The paramiko client saves host keys incorrectly whereas the host keys object does
    # this correctly, so use that with the client filename variable.
    # See: https://github.com/paramiko/paramiko/pull/1989
    host_keys.save(client._host_keys_filename)
```

This happens because of the way Paramiko parses the file - it throws away comment lines, newlines, and lines it doesn't understand.

https://github.com/paramiko/paramiko/blob/main/paramiko/hostkeys.py#L89-L97:

```
with open(filename, "r") as f:
    for lineno, line in enumerate(f, 1):
        line = line.strip()
        if (len(line) == 0) or (line[0] == "#"):
            continue
        try:
            entry = HostKeyEntry.from_line(line, lineno)
        except SSHException:
            continue
```

So, when Pyinfra calls `hosts_keys.save(...)`, this doesn't just add a new entry, it actually overwrites the file with the contents **as parsed by Paramiko**. This is unexpected - I've been fighting disappearing contents in known_hosts for years and had no idea it was caused by PyInfra.

## Expected behavior
Pyinfra should either never modify known_hosts, or only add new hosts.

I understand the intent here, to make running PyInfra as convenient as possible. But IMO this behavior is a foot-gun for infrastructure maintainers and should be changed. It's frustrating to lose formatting, painful to lose comments, and breaks general SSH usage when `@cert-authority` (not yet support by Paramiko) silently disappears.

evoldstad:
Hi, 

I created a [possible fix](https://github.com/evoldstad/pyinfra/tree/known_hosts_fix) for this, making it so that pyinfra will only append to the known_hosts file. Do you have the possibility of testing it with a known_hosts file that is known to break? Alternatively providing an example that demonstrates the behavior. I would like to add a test for this to ensure we don't repeat the behavior, as I agree that this has a very negative impact for certain users.

pikeas:
Here's a reproduction:

```
$ mv ~/.ssh/known_hosts ~/.ssh/know_hosts.bak
[...]

PR #1248, opened 2024-11-27
pyinfra/connectors: Fix overwriting of users known_hosts file (#1209)

evoldstad:
This fixes issue #1209, making it so that we append new keys to the users known_hosts file instead of overwriting it.  

I also added a testcase that should discover this breaking in the future. 

Additionally: 
- Broke out the append functionality into a "append_hostkey" function, making it so we don't needlessly reuse code for the AskPolicy and AcceptNewPolicy.

Previous behaviour when adding a new key: 
- Create a paramiko.HostKeys object
- Read the users known_hosts file
- Add the new key to the object
- Save the object, overwriting the users host_keys file. 

New behaviour: 
- Create a paramiko.HostKeyEntry object using the new hostname and corresponding key. 
- Append this key to the existing known_hosts file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# v3.1.1
+
+- Improve errors with 2.x style `@decorator` (vs `@decorator()`) functions
+- Document adding custom connectors (@simonhammes)
+- Add basic API example to docs (@pirate)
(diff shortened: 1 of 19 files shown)


evoldstad:
Forgot to do linting and squashing commits, will redo this.

PR #1252, opened 2024-12-02
pyinfra/connectors: Fix overwriting of users known_hosts file (#1209)

evoldstad:
This fixes issue #1209, making it so that we append new keys to the users known_hosts file instead of overwriting it.

Additionally:
- Added a testcase that should discover this breaking in the future
- Broke out the append functionality into a "append_hostkey" function, making it so we don't needlessly reuse code for AskPolicy 
- Linting actually correct this time.
- Merged commits correctly this time. 

Previous behaviour when adding a new 
- Create a paramiko.HostKeys object 
- Read the users known_hosts file  
- Add the new key to the paramiko object.  
- Save the object, overwriting the users host_keys file without comments and entries incompatible with paramiko. 

New behaviour:
- Create a paramiko.HostKeyEntry object using the new hostname and
corresponding key.
- Append this key to the existing known_hosts file.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -13,6 +13,7 @@ extend-ignore-re = [
     "forr something in",
     "unknown tag 'forr'",
     "nd6 options=",
+    "4Ue1", # Ignore false positive in test_sshuserclient.py (EXAMPLE_KEY_1)
 ]
 
 [default.extend-words]
--- a/pyinfra/connectors/sshuserclient/client.py
+++ b/pyinfra/connectors/sshuserclient/client.py
@@ -14,6 +14,7 @@ from paramiko import (
     SSHException,
 )
 from paramiko.agent import AgentRequestHandler
+from paramiko.hostkeys import HostKeyEntry
 
 from pyinfra import logger
 from pyinfra.api.util import memoize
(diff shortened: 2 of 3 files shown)


evoldstad:
The spelling issue seems to be a false positive, as it reacts on a part of the example key used in the test. Any way to override this behaviour, or mark it as a non-issue?

simonhammes:
> The spelling issue seems to be a false positive, as it reacts on a part of the example key used in the test. Any way to override this behaviour, or mark it as a non-issue?

You can either ignore the whole file or ignore certain expressions based on regular expressions: https://github.com/pyinfra-dev/pyinfra/blob/3.x/.typos.toml

evoldstad:
> > The spelling issue seems to be a false positive, as it reacts on a part of the example key used in the test. Any way to override this behaviour, or mark it as a non-issue?
> 
> You can either ignore the whole file or ignore certain expressions based on regular expressions: https://github.com/pyinfra-dev/pyinfra/blob/3.x/.typos.toml

Thanks, that fixed it!

Fizzadar (approved):
Amazing, thank you for fixing this @evoldstad!

PR #1231, opened 2024-10-31
facts: add efibootmgr.EFIBootMgr

bauen1:
Add a fact that parses the output of `efibootmgr`.

Currently based on ~~#1179~~ #1250

PR #1254, opened 2024-12-19
facts: process_data is not a staticmethod

bauen1:
This makes the code consistent.


Fizzadar:
Merged as part of #1231 (https://github.com/pyinfra-dev/pyinfra/commit/aba88d087105dd7ff503eb2485a86b4715fe7e9f)

PR #1255, opened 2024-12-19
Require explicit override decorator

bauen1:
Not sure if this is a good idea, hence a draft MR, there would be some 304 missing `@override` decorators.

### DIFF
--- a/pyinfra/api/command.py
+++ b/pyinfra/api/command.py
@@ -236,7 +236,7 @@ class FunctionCommand(PyinfraCommand):
         if "state" in argspec.args and "host" in argspec.args:
             return self.function(state, host, *self.args, **self.kwargs)
 
-        def execute_function():
+        def execute_function() -> None:
             with ctx_config.use(state.config.copy()):
                 with ctx_host.use(host):
                     self.function(*self.args, **self.kwargs)
--- a/pyinfra/api/command.py
+++ b/pyinfra/api/command.py
@@ -6,7 +6,7 @@ from string import Formatter
 from typing import IO, TYPE_CHECKING, Callable, Union
 
 import gevent
-from typing_extensions import Unpack
+from typing_extensions import Unpack, override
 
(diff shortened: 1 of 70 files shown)


Fizzadar:
I think this is a good change, an explicit typing improvement with no actual functionality change. Only concern might be breaking type-checking with other modules but that doesn't fall under the version compatibility anyway so probably OK...

bauen1:
Well, it passes the pipeline now, but locally I still get some errors:

```
jh@pns-admin:~/dev/pyinfra$ mypy
Warning: Unpack is already enabled by default
pyinfra/context.py:83: error: Incompatible types in assignment (expression has type "type[local]", base class "ContextObject" defined the type as "type[container]")  [assignment]
pyinfra_cli/__main__.py:34: error: Missing positional argument "greenlet" in call to "signal"  [call-arg]
pyinfra_cli/__main__.py:37: error: Module has no attribute "signal"  [attr-defined]
Found 3 errors in 2 files (checked 145 source files)
```

So I'm guessing there's some different version of greenlet installed in my container ?

Fizzadar:
> Well, it passes the pipeline now, but locally I still get some errors:
> 
> ```
> jh@pns-admin:~/dev/pyinfra$ mypy
> Warning: Unpack is already enabled by default
> pyinfra/context.py:83: error: Incompatible types in assignment (expression has type "type[local]", base class "ContextObject" defined the type as "type[container]")  [assignment]
> pyinfra_cli/__main__.py:34: error: Missing positional argument "greenlet" in call to "signal"  [call-arg]
> pyinfra_cli/__main__.py:37: error: Module has no attribute "signal"  [attr-defined]
> Found 3 errors in 2 files (checked 145 source files)
> ```
> 
> So I'm guessing there's some different version of greenlet installed in my container ?

They look unrelated though? Possibly greenlet related, I'll pull this and poke at it a bit when I can.

PR #1257, opened 2024-12-20
configparser no longer required

bkmgit:


Fizzadar (approved):
Nice catch, thanks @bkmgit!

PR #1259, opened 2024-12-28
Fix docker.network() documentation and network removal

lorenzph:
The `docker.network()` operation's documentation no longer matched the actual
API. Adapt it accordingly.

Additionally, fix network removal which passed an incorrect `command` to
`handle_docker`.


Fizzadar (approved):
Thanks for fixing these @lorenzph!

PR #1266, opened 2025-01-12
Load test specs with PyYAML instead of json

taliaferro:
YAML provides a number of advantages over JSON for this application. It supports comments, which could make it easier to communicate what a test is doing and why, and it also allows multiline strings, which makes it much easier to write fact tests (which currently must transform their output into an array of strings, one per line).

YAML is a superset of JSON (valid JSON is also valid YAML) so nearly all the existing tests can be carried forward with no changes. (A few did contain inconsistent use of tabs and spaces for indentation -- JSON allows this but YAML does not so these tests were corrected in commit 13610681).
However, I have also converted the tests for the ZFS facts and operations to YAML for demonstration purposes, and as an example of how future tests could be written.


Fizzadar:
@taliaferro this is awesome! What a great idea :) Linting thing to address but otherwise this looks decent.

PR #1270, opened 2025-01-15
CI: Prevent docs job from running on forks

simonhammes:
The job always fails due to the missing token and causes annoying notifications when rebasing `3.x` on a fork


Fizzadar (approved):
heh, nice catch, thanks @simonhammes!

PR #1271, opened 2025-01-16
enable -h as shorthand for --help

NichtJens:
This is a tiny improvement that enables the CLI switch `-h` as shorthand for `--help`.

I often find myself trying `-h`, failing and then remembering that specific tools only offer `--help`. `-h` is commonly available from CLI tools, argparse has it a default, etc. Thus, it would be nice if pyinfra would allow it too.


Fizzadar (approved):
Nice, good little win there, thank you @NichtJens!

PR #1272, opened 2025-01-17
`docker.container`: Refactor to support container recreation

minor-fixes:
This change refactors the way the `docker.container` operation manages containers, in preparation for work to make recreation more intelligent. This change is (mostly) a pure refactor; future changes will diff the current container against the operation parameters to determine if a container needs to be recreated. This will fix an issue where changing any of the operation arguments does not result in actual container changes upon execution.

In this refactor, container parameters are moved to a dedicated class, which centralizes the `docker container` command-line argument generation logic and the future diffing logic. The diff function is roughed in, though it currently reports "no diff" (to match the operation's current behavior). Since conditional recreation complicates the operation's logic on which commands to execute, the decisions are boosted into boolean variables to increase readability.

As a side benefit, supporting additional docker container parameters should be more straightforward due to the centralization in said dedicated class (I'm planning on adding support for container args, uid, and other Docker params currently not supported).

The only behavioral change is that creating and starting a container is no longer done in one exec (joined by `;`) but rather two separate docker commands. This sidesteps questions about whether `;` is the correct joiner (as opposed to `&&`) and reduces the amount of `kwargs` fishing in the implementation.

Tested:
* `scripts/dev-test.sh` and `scripts/dev-test-e2e.sh` both pass, save for warnings also present prior to this change

### DIFF
--- a/tests/operations/docker.container/add_existent_container.json
+++ b/tests/operations/docker.container/add_existent_container.json
@@ -9,242 +9,240 @@
         "force": true
     },
     "facts": {
-        "docker.DockerContainer":
-            { 
-                "object_id=nginx": [
-            
-                        {
-                                "Id": "9bb5a79e7c4da9ad38e7183aa7c943ee41dcbbab70e4832fe29b2788ab88d4b4",
-                                "Created": "2024-05-26T22: 01: 24.10525839Z",
-                                "Path": "/docker-entrypoint.sh",
-                                "Args": [
-                                    "nginx",
-                                    "-g",
-                                    "daemon off;"
-                                ],
-                                "State": {
-                                    "Status": "running",
-                                    "Running": "True",
-                                    "Paused": "False",
-                                    "Restarting": "False",
-                                    "OOMKilled": "False",
-                                    "Dead": "False",
-                                    "Pid": 8407,
-                                    "ExitCode": 0,
-                                    "Error": "",
-                                    "StartedAt": "2024-05-26T22: 01: 24.502384646Z",
-                                    "FinishedAt": "0001-01-01T00: 00: 00Z"
-                                },
-                                "Image": "sha256:e784f4560448b14a66f55c26e1b4dad2c2877cc73d001b7cd0b18e24a700a070",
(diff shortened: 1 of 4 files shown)


tests/operations/docker.container/add_existent_container.json
```diff
@@ -9,242 +9,240 @@
...
-                                            "IPAddress": "172.17.0.2",
-                                            "IPPrefixLen": 16,
-                                            "IPv6Gateway": "",
-                                            "GlobalIPv6Address": "",
-                                            "GlobalIPv6PrefixLen": 0,
-                                            "DriverOpts": "None",
-                                            "DNSNames": "None"
-                                        }
-                                    }
+        "docker.DockerContainer": {
```
minor-fixes, line 12: Apologies for the diff here; the diffs in this fact object should be purely whitespace (I think formatting this JSON does help readability, so I'd prefer to leave it in if possible)

simonhammes:
Just some food for thought: Would supporting `docker compose` natively in pyinfra (tracked in https://github.com/pyinfra-dev/pyinfra/issues/1276) make sense instead? 

`docker compose up -d` essentially handles all of the diffing.
Idempotency could be achieved by checking whether the file `.yml` file to template has changed.

minor-fixes:
> Just some food for thought: Would supporting `docker compose` natively in pyinfra (tracked in #1276) make sense instead?
> 
> `docker compose up -d` essentially handles all of the diffing. Idempotency could be achieved by checking whether the file `.yml` file to template has changed.

Yes, I think that'd make a ton more sense - otherwise the routines here will essentially replicate the functionality of `docker compose`. Is someone already working on that?

simonhammes:
> Is someone already working on that?

Not that I'm aware of :slightly_smiling_face:

Fizzadar:
Re: compose - I think this is somewhat different since this will work for people just wanting to manage a container w/o any compose file?

Fizzadar (approved):
@minor-fixes sorry for taking ages to review this! And THANK YOU for the PR! This looks awesome, really nice refactor 👌

Fizzadar:
I'm going to merge this to unblock the others per above I think compose is slightly different (and still worth supporting alongside).

PR #1277, opened 2025-01-20
docker.container: Recreate container when args change

minor-fixes:
This PR allows the `docker.container` operation to tear down and
recreate the container when operation arguments change, instead of
reporting `No change` and doing nothing. This is intended to reduce the
possibility for human error/need for manual intervention when changing
args to `docker.container` operations.

Since it is not possible to extract all operation args from e.g. `docker
inspect` output, this PR takes a similar approach to Docker Compose to
tackle this issue - it serializes the operation args in a deterministic
way, hashes the serialized bytes, and stores this as a label on the
container. If the hash differs from a currently-running container, the
container is recreated.

Tested: Added additional tests for behavior when args are
changing/static in different scenarios

### DIFF
--- a/pyinfra/operations/docker.py
+++ b/pyinfra/operations/docker.py
@@ -8,7 +8,7 @@ from pyinfra import host
 from pyinfra.api import operation
 from pyinfra.facts.docker import DockerContainer, DockerNetwork, DockerVolume
 
-from .util.docker import handle_docker
+from .util.docker import ContainerSpec, handle_docker
 
 
 @operation()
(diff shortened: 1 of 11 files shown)


Fizzadar (commented):
Oh now I see how this is similar to compose - did not know that's how compose managed this, TIL!

Looks good to me overall, I think this still has value as it allows us to properly handle diff cases without using docker compose..

Fizzadar:
Thank you for working on this! The functionality has landed via #1272 so closing this one now!

PR #1285, opened 2025-02-04
Add note re: global arguments to operations

simonhammes:
See the note at the bottom:

Preview


Fizzadar (approved):
Very nice, thank you for adding this @simonhammes!

PR #1289, opened 2025-02-08
Add a note to the docs about using `_inner` when calling operations from other operations

CSDUMMI:
While this behavior is shown in the example, it is not explicit that this is a requirement.


Fizzadar (approved):
Oh nice catch @CSDUMMI, thanks for the fix!

PR #1293, opened 2025-02-15
server.sysctl: handle 0 integer values correctly

lemmi:
Previously, sysctl operations would incorrectly trigger changes when checking against zero values since `not 0` is `True` in python.

This drops checking the `existing_value` and relies on the comparison against the desired value instead.

Added test case verifying correct noop behavior when sysctl value is 0.


Fizzadar (approved):
Awesome, thank you for fixing this @lemmi!

PR #1250, opened 2024-11-30
docs: document any_changed, all_changed helpers for _if

bauen1:

PR #1295, opened 2025-02-16
docs: correct import path for any_changed, all_changed

lemmi:
Fixes: #1250


Fizzadar (approved):
Thanks for fixing this too @lemmi!

Issue #1288, opened 2025-02-08
Current LXD support doesn't support remote servers


goetzk:
(I'm bringing this in from the matrix chat so it can be tracked)

## Describe the bug
I've had a quick look at existing LXD support in pyinfra and noticed the existing lxd support requires the remote to be lxc's default remote.

facts/lxd.py has no way to specify a remote server to list the instances from
operations/lxd.py has no way to specify a remote server to add new instances to / remove unwanted instances from

## To Reproduce

The hard coding is visible in the source code:

```
:~/source/pyinfra$ grep 'lxc ' pyinfra/operations/lxd.py
                yield "lxc stop {0}".format(id)
            yield "lxc delete {0}".format(id)
            yield "lxc launch {image} {id} < /dev/null".format(id=id, image=image)
:~/source/pyinfra$ grep 'lxc ' pyinfra/facts/lxd.py
        return "lxc list --format json --fast"
```

Its possible this will use the default `remote` configured for lxc but I haven't tested.

## Expected behavior

Some way to access a remote server to manage instances.

As noted in #677#issuecomment-2644410227 I'm doing some experiments with pyinfra and LXD so might be able to look at this issue as part of that process.

romain-dartigues:
#1368 let you use local and remote incus/LXD servers.

goetzk:
It'll need testing, but https://github.com/pyinfra-dev/pyinfra/issues/677#issuecomment-4336287885 may help resolve this.

PR #1301, opened 2025-02-24
Add documentation on extending

goetzk:
While working on my custom stuff I gathered up my notes (and will likely gather more). 
I haven't worked with custom facts or operations yet so haven't put anything in those sections but if I do (eg for #1288) I'd be able to expand this further.

@Fizzadar Would this sort of information be useful if I rounded out the page created in this commit? Would you like the information, but rather it was placed somewhere else in the documentation?


Fizzadar (commented):
I need to think about where this fits in, there's a lot of overlap with the contributing doc: https://docs.pyinfra.com/en/3.x/contributing.html, might be best to expand there?

goetzk:
> I need to think about where this fits in, there's a lot of overlap with the contributing doc: https://docs.pyinfra.com/en/3.x/contributing.html, might be best to expand there?

There is definitely overlap and I'm really not sure of the best angle to approach. The perspective I took with this PR was "What did I need to know/do in order to create an installable custom connector package", and left stub bits in to explain how (eg) custom libraries and facts could be added to that same installable package.

Perhaps what really drives this PR is 'how can custom pyinfra bits be made easily distributable'. If thats a question you want answered in the documentation it would lead to a restructuring of what I've done along these lines:
-  `docs/extending.rst` retitled something like 'packaging custom code'
- custom libraries, custom facts, custom connectors, tests sections all removed from the page
- exceptions section moved elsewhere? unsure
- debugging section moved elsewhere?

Then I imagine writing operations/facts/connectors would all have a footnote added saying 'look at  packaging plugins section for information on making your code distributable'

Fizzadar:
I'm going to close this as most of it has been added to the refreshed docs already!

PR #1304, opened 2025-02-27
Document host, state, inventory in files.template

mpilone:
Documented that the host, state, and inventory are passed to the files.template operation.


Fizzadar (approved):
Nice, thanks for adding this @mpilone!

PR #1308, opened 2025-03-09
facts.xbps.XbpsPackages: allow . in package names

lemmi:
Package names including dots `.` were split at the wrong location. Parts that included dots were counted as versions:

    {
        "python3": [
            "3.13.2_1",
            "jaraco.classes",
            "jaraco.context"
        ],
    }

This patch fixes the regex and adds a test for this case:

    {
        "python3": [
            "3.13.2_1"
        ],
        "python3-jaraco.classes": [
            "3.4.0_2"
        ],
        "python3-jaraco.context": [
            "6.0.1_2"
        ]
    }

### DIFF
--- a/pyinfra/facts/xbps.py
+++ b/pyinfra/facts/xbps.py
@@ -21,7 +21,7 @@ class XbpsPackages(FactBase):
 
     default = dict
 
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice catch & fix, @lemmi, thank you!

PR #1310, opened 2025-03-12
Add RebootRequired fact for detecting system reboot needs

wowi42:
Introduces a new fact to check if a system requires a reboot across different Unix-like systems:
- Linux: Checks /var/run/reboot-required files and Alpine Linux kernel comparison
- FreeBSD: Compares running and installed kernel versions

Supports detection methods for Linux and FreeBSD systems with a flexible shell script approach.


wowi42:
not sure if it is the best way to do it

tests/facts/server.RebootRequired/alpine_reboot_required.json
```diff
@@ -0,0 +1,13 @@
+{
+    "command": "\n# Get OS type\nOS_TYPE=$(uname -s)\n\nif [ \"$OS_TYPE\" = \"Linux\" ]; then\n    # Check if it's Alpine Linux\n    if [ -f /etc/alpine-release ]; then\n        RUNNING_KERNEL=$(uname -r)\n        INSTALLED_KERNEL=$(ls -1 /lib/modules | sort -V | tail -n1)\n        if [ \"$RUNNING_KERNEL\" != \"$INSTALLED_KERNEL\" ]; then\n            echo \"reboot_required\"\n            exit 0\n        fi\n    else\n        # Check standard Linux reboot required files\n        if [ -f /var/run/reboot-required ] || [ -f /var/run/reboot-required.pkgs ]; then\n            echo \"reboot_required\"\n            exit 0\n        fi\n    fi\nelif [ \"$OS_TYPE\" = \"FreeBSD\" ]; then\n    RUNNING_VERSION=$(freebsd-version -r)\n    INSTALLED_VERSION=$(freebsd-version -k)\n    if [ \"$RUNNING_VERSION\" != \"$INSTALLED_VERSION\" ]; then\n        echo \"reboot_required\"\n        exit 0\n    fi\nfi\n\necho \"no_reboot_required\"\n",
+    "output": [
+        "reboot_required"
+    ],
+    "fact": true,
+    "setup": [
```
Fizzadar: I don't think this key does anything?

pyinfra/facts/server.py
```diff
@@ -796,3 +796,52 @@ def process(self, output):
...
+    - On Alpine Linux, compares installed kernel with running kernel
+
+    On FreeBSD systems:
+    - Compares running kernel version with installed kernel version
+    """
+
+    def command(self) -> str:
+        return """
+# Get OS type
+OS_TYPE=$(uname -s)
```
Fizzadar, line 924: Ah yes, the problem of facts being unable to call each other. Unfortunately I think this is the best way we can currently do this.

Fizzadar (commented):
I think there's some extra test args but otherwise this looks good. The script is somewhat hacky but there's no real alternative with the way facts work today.

pyinfra/facts/server.py
```diff
@@ -899,3 +899,52 @@ def process(self, output):
...
+
+    On Linux systems:
+    - Checks /var/run/reboot-required and /var/run/reboot-required.pkgs
+    - On Alpine Linux, compares installed kernel with running kernel
+
+    On FreeBSD systems:
+    - Compares running kernel version with installed kernel version
+    """
+
+    def command(self) -> str:
```
Fizzadar, line 921:
```suggestion
    @override
    def command(self) -> str:
```

pyinfra/facts/server.py
```diff
@@ -899,3 +899,52 @@ def process(self, output):
...
+    if [ "$RUNNING_VERSION" != "$INSTALLED_VERSION" ]; then
+        echo "reboot_required"
+        exit 0
+    fi
+fi
+
+echo "no_reboot_required"
+"""
+
+    def process(self, output) -> bool:
```
Fizzadar, line 953:
```suggestion
    @override
    def process(self, output) -> bool:
```

Fizzadar (commented):
Two override decorators needed to fix type check, then good to go I think?

Fizzadar:
@wowi42 do you mind giving me push access to the branch, I'd like to fix up the styling to get this merged!

wowi42:
@Fizzadar sorry for the delay. You got access to the branch and I did some patch. We are good to go I think.

Fizzadar (approved):
Awesome, thank you for adding this @wowi42!

PR #1311, opened 2025-03-12
[IMP] files.file: amend <touch> flag meaning

Pirols:
As it stands, the `<touch>` flag of `files.file` is useless: it only controls whether a `touch <path>` command is issued when the `<path>` file already exists, which results in a no-op, in both cases.

Since `<assume_present>` was removed in dfa69b66fe8ebf9c02accb1d3f6ec4905f11476d, it may make sense to have `<touch>` control whether the operation is allowed to create an empty file, if one is not found. And, raise if the file is not found and the flag is unset: the usecase is that one may want to only change ownership/permissions of an existing file, without accidentally creating an empty placeholder for it.

A small unrelated change to the condition to set the mode is also included.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -1204,7 +1204,7 @@ def file(
     user: str | None = None,
     group: str | None = None,
     mode: int | str | None = None,
-    touch=False,
+    touch=True,
     create_remote_dir=True,
     force=False,
     force_backup=True,
@@ -1268,6 +1268,9 @@ def file(
         return
 
     if info is None:  # create
+        if not touch:
+            raise OperationError("File {0} does not exist and `touch` is unset".format(path))
+
         if create_remote_dir:
             yield from _create_remote_dir(path, user, group)
 
@@ -1281,12 +1284,8 @@ def file(
     else:  # update
         changed = False
 
-        if touch:
-            changed = True
-            yield StringCommand("touch", QuoteString(path))
-
         # Check mode
-        if mode and (not info or info["mode"] != mode):
+        if mode and info["mode"] != mode:
             yield file_utils.chmod(path, mode)
             changed = True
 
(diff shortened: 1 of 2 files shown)


lemmi:
`touch` is not a noop on existing files. It updates `atime` and `mtime`.

Pirols:
> `touch` is not a noop on existing files. It updates `atime` and `mtime`.

Uhm, indeed. Either way, the api could be clearer: unsetting `touch` does not prevent a new file from being created if it didn't exist already, that's surprising imo.

lemmi:
Ignoring the other options, this is the current behaviour:

| `present` | `touch` | file exists | file is missing | |
|---|---|---|---|---|
| `True` | `False` | nothing changes | file is created | default | 
| `True` | `True` | file is touched | file is created | |
| `False` | `False` |  file is removed | no change | |
| `False` | `True` | file is removed | no change | |

Personally this is pretty much how I expected things to be and it makes sense. `present=False` and `touch=True` is a rather odd combination. But I'd argue not having a file at the end of the operation "i.e. not being present" makes sense.

Pirols:
My point is more about the default set of options: `present=True` and `touch=False`. I would expect those to indicate that the user wants a file to be present (and perhaps perform some `chown`/`chmod` ops on it) while preventing `pyinfra` from creating (`touch`ing) an empty file that would act as a placeholder. In that case, IMO it would make sense for `pyinfra` to inform the user that the file is missing instead.

Fizzadar:
I think @lemmi is correct here, `present=True` means "the file should exist", this matches every other (stateful) operation, arguments define the desired state. If present pyinfra should be ensuring the file exists.

Pirols:
> I think @lemmi is correct here, `present=True` means "the file should exist", this matches every other (stateful) operation, arguments define the desired state. If present pyinfra should be ensuring the file exists.

Uhm, I understand the stance. The behavior proposed here can be achieved with preliminary fact checks if one wants to, then!

PR #1312, opened 2025-03-12
files.move: new operation

Pirols:
Proposing a fairly simple operation to use `mv` among remote files. Notice: it's not as smart. `mv t1 t2/t1` will try to create `t2/t1/t1`.

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -1069,6 +1069,34 @@ def template(
     )
 
 
+@operation()
+def move(src: str, dest: str, overwrite=False):
+    """
+    Move remote file/directory/link into remote directory
+
+    + src: remote file/directory to move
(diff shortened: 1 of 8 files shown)


pyinfra/operations/files.py
```diff
@@ -1069,6 +1069,34 @@ def template(
     )
 
 
+@operation()
+def move(src: str, dest: str, overwrite=False):
+    """
+    Move remote file/directory/link into remote directory
+
+    + src: remote file/directory to move
+    + dest: remote directory to move `src` into
```
Fizzadar, line 1078: Might not be smart but the argument is explicit that this is a directory, so I think that's OK.

Fizzadar (approved):
Thank you for the PR, @Pirols, nice addition!

PR #1313, opened 2025-03-18
facts: server.Mounts: fix whitespaces and escaped characters

lemmi:
The `mount` command can not be parsed unambigiously. Whitespaces in paths and `,` in option lists are not escaped. See the new tests for examples the old code can not handle.

This patch switches to reading from `/proc/self/mountinfo` instead (see man 5 proc for documentation) where strings are properly escaped.

While `/proc/self/mountinfo` provides a lot more details, these are ignored to keep compatibility with the old type. If there is a need, the missing fields can be easily added.

@Fizzadar 461d8ec5c8556e6c2db6b9803e7a62085dea628f introduced handling "map devices". I'm not sure what they are. I'd love to see a line from `/proc/self/mountinfo`, so we can test against that.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -206,34 +206,53 @@ class Mounts(FactBase[Dict[str, MountsDict]]):
 
     @override
     def command(self):
-        return "mount"
+        return "cat /proc/self/mountinfo"
 
     @override
     def process(self, output) -> dict[str, MountsDict]:
         devices: dict[str, MountsDict] = {}
 
+        def unescape_octal(match: re.Match) -> str:
+            s = match.group(0)[1:]  # skip the backslash
+            return chr(int(s, base=8))
+
+        def replace_octal(s: str) -> str:
+            """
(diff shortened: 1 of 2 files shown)


Fizzadar (commented):
Oh this is a great change @lemmi!

> @Fizzadar https://github.com/pyinfra-dev/pyinfra/commit/461d8ec5c8556e6c2db6b9803e7a62085dea628f introduced handling "map devices". I'm not sure what they are. I'd love to see a line from /proc/self/mountinfo, so we can test against that.

Heh.. I'm not sure I remember, will see if I can dig up some more info on that..

lemmi:
> Heh.. I'm not sure I remember, will see if I can dig up some more info on that..

I think I saw something related to lvm volumes. But I'm not using them anywhere at the moment.

I had another look into how exactly the kernel escapes the strings to get rid of the `encode().decode(...)` chain and replaced that with a specific conversion.

Fizzadar (approved):
Got there eventually, thank you for putting this together @lemmi!

Still not found the map thing, since I failed to add any test or useful context I'm going ahead and merging this as-is, we can re-add the map thing if needed.

PR #1315, opened 2025-03-18
Fix wrong example operation using forbidden argument 'name'

robertmx:


Fizzadar (approved):
Whoops! Thanks for fixing @robertmx

PR #1316, opened 2025-03-24
Update PACMAN_REGEX to support additional characters and modify packa…

wowi42:
…ge.json to include new package version for lm_sensors


Fizzadar (approved):
Thank you for fixing this @wowi42!

PR #1317, opened 2025-03-24
Remove useless parameter

wowi42:


Fizzadar (approved):
Woa that's some old legacy code there, thanks for removing it @wowi42!

PR #1318, opened 2025-03-24
Enhance PostgreSQL role management by adding ALTER ROLE support

wowi42:
Updated the role function to yield SQL commands for altering existing PostgreSQL roles, allowing modifications to attributes such as LOGIN, SUPERUSER, and CONNECTION LIMIT. Changed the parameter name for sudo_user to _sudo_user for consistency.


Fizzadar (commented):
Hrm, it'd be nice to not make every `postgres.role` operation execute a command every time, could we check the existing role to see if `ALTER` is needed?

I think if a password is set we'll always need to alter because we can't check those, but all the other arguments should be the relevant info on the roles fact.

wowi42:
Correct, I will work on it

wowi42:
@Fizzadar I fixed it but we need to fix the actions

wowi42:
@Fizzadar ready to be merged

Fizzadar (approved):
Awesome, thanks @wowi42!

PR #1321, opened 2025-03-27
Add example of logging to using operations docs

goetzk:
How to log output from pyinfra has come up several times, I thought I'd try to include an example.


Fizzadar (approved):
Nice add, thanks @goetzk!

PR #1322, opened 2025-03-27
Add inventory processing note and reference it

goetzk:
I thought the contents of inventories would be processed automatically but that isn't the case. To clear things up I've added a warning to inventory-data and added a reference to that page from the example docs.


Fizzadar (approved):
Nice - thanks @goetzk!

PR #1323, opened 2025-03-27
Add a section explaining connector flow

goetzk:
Methods are used at different times to affect the configuration process; this tries to explain which methods can be used and when they will be run.


docs/api/connectors.md
```diff
@@ -136,6 +115,71 @@ class LocalConnector(BaseConnector):
...
+appropriate times.
+
+To hook in to to the various steps with the methods outlined below.
+```
+--> Loading config...
+--> Loading inventory...
+```
+`make_names_data` is used to supply inventory data about a host while at 'Loading inventory' stage.
+
+Its worth being aware up front that due to `make_names_data` being a `staticmethod` it has no automatic access to the parent classes attributes.
```
Fizzadar, line 130: This is an annoying decision I made that I'd like to change 🙃

Fizzadar (approved):
Huge improvement, thank you @goetzk!

PR #1331, opened 2025-04-04
[ADD] ops/server.user: allow to add user to more secondary groups 

Pirols:
It is often necessary to add a user to a secondary group, without dropping it from other secondary groups it already belonged to. `usermod` provides a `-a` flag dedicated to this very need. This commit allows to leverage that flag via `pyinfra`.

In fact, in `pyinfra` this is even more necessary as facts are collected and never updated, making so that:

```python
from pyinfra import host
from pyinfra.facts.server import Users

server.group(
    name="create main group",
    group="user",
)

server.group(
    name="create secondary group 1",
    group="extra_group1",
)

server.user(
    name="create user `user`",
    user="user",
    groups=["extra_group1"],
)

server.group(
    name="create secondary group 2",
    group="extra_group2",
)

server.user(
    name="Add user to extra group 2",
    user="user",
    groups=[
        *host.get_fact(Users).get("user", {}).get("groups", []),
        "extra_group2",
    ],
)
```

* yields `usermod -G extra_group1,extra_group2 user` if `user` was member of `extra_group1` before the task started, or
* yields `usermod -G extra_group2 user`, if it was not part of the task

Notice that it does not matter whether the user `user` was added to the `extra_group1` group by `pyinfra`.

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -764,6 +764,7 @@ def user(
     shell: str | None = None,
     group: str | None = None,
     groups: list[str] | None = None,
+    append=False,
     public_keys: str | list[str] | None = None,
     delete_keys=False,
     ensure_home=True,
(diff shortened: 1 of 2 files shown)


Pirols:
The CI linting seems to be failing, while `scripts/dev-lint.sh` does not highlight any issue. And the error message in the CI message is cryptic to me.

Fizzadar (approved):
Nice, thank you for adding this @Pirols!

Re:CI - it's the black formatter failing on a completely unrelated file, so ignoring!

PR #1335, opened 2025-04-10
facts: add server.Port fact for checking which process runs there

missytake:
My first pyinfra PR, I hope I'm doing this right :))) and that this fact makes sense for others as well

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -258,7 +258,7 @@ class Mounts(FactBase[Dict[str, MountsDict]]):
         return devices
 
 
-class Port(FactBase[Tuple[str, int]]):
+class Port(FactBase[Tuple[str, int] | Tuple[None, None]]):
     """
     Returns the process occuping a port and its PID
     """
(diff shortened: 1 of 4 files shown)


missytake:
Hm, how can I make the end-to-end tests happy? I don't have docker installed, and I don't understand why `pytest -m end_to_end_ssh` is unhappy when `mypy` is happy.

One way would be returning `("", 0)` on an unused port... on Linux PID 0 is always unused (it is only the PPID (parent process ID) of the `/sbin/init splash` process. On Windows it seems to be the system idle process: https://stackoverflow.com/a/3232487
The advantage of using `0` as a return value here is that it means `False` in control flows, other than `-1` which means `True`.

What do you think? And what else is missing to merge this?

Fizzadar:
Hi @missytake! Thank you for the PR!

> Hm, how can I make the end-to-end tests happy? I don't have docker installed, and I don't understand why `pytest -m end_to_end_ssh` is unhappy when `mypy` is happy.

I believe this was broken when you branched off 3.x; I just rebased your branch so it should pass now.

> One way would be returning `("", 0)` on an unused port... on Linux PID 0 is always unused (it is only the PPID (parent process ID) of the `/sbin/init splash` process. On Windows it seems to be the system idle process: https://stackoverflow.com/a/3232487 The advantage of using `0` as a return value here is that it means `False` in control flows, other than `-1` which means `True`.
> 
> What do you think? And what else is missing to merge this?

I think the latest `None, None` return seems best for this case, let's stick with that 👍

missytake:
Great, all green :)

PR #1336, opened 2025-04-11
ssh: Add support for ServerAliveInterval

chipot:
I have some tasks running in pyinfra that are particularly long. This is within my expectations, but the problem is that paramiko raises EOFError for all the other hosts that are waiting their turn to run the next step.

After looking a bit I realize paramiko doesn't support `SeverAliveInterval` from ssh/config, but has a call `transport.set_keepalive` doing the same thing.

This PR adds the parsing for `ServerAliveInterval` from ssh/config and sets appropriately the keepalive on the transport.


Fizzadar (approved):
Nice find @chipot, this is great!

PR #1337, opened 2025-04-11
[WIP] refactor: enhance DebPackages and DebPackage facts to include detailed package information

wowi42:
- Updated the DebPackages class to return a structured dictionary for each package, including desired action, status, version, architecture, and description.
- Modified the regex pattern to capture additional package details.
- Adjusted related test cases to reflect the new data structure for package facts.


Fizzadar (commented):
@wowi42 this is neat, way more info than before, but does introduce a breaking change (I believe it'll also break `apt.packages`).

It might be better off to expand the `DebPackage` fact rather than this one - is there a use-case in mind for getting all this info for all packages vs. individual packages?

PR #1349, opened 2025-04-29
ssh: Fix `TypeError` in `SSHClient.connect`

chipot:
Fixed issue introduced by my previous PR: https://github.com/pyinfra-dev/pyinfra/pull/1336

`keep_alive` was passed to `ParamikoClient.connect` unexpectedly, causing a `TypeError` exception


Fizzadar (approved):
Oops! LGTM. Would be good to add a test for this case but can cover that in a separate PR.

PR #1352, opened 2025-04-30
Update Docker command handling to avoid empty results

### LINKED ISSUE #1278: DockerImages fact raises exception when there are no images present
## Describe the bug
The pyinfra.facts.docker.DockerImages fact raises an exception if there is no image.

## To Reproduce
1. Ensure there are no images present with docker image ls
2. Check the fact with host.get_fact(DockerImages)
3. Exception "docker image inspect requires at least on argument"

## Expected behavior
I expect it to return an empty list instead of an exception

## Meta

Pyinfra 3.2

The behavior is rather obvious when looking at the implementation of DockerImages. In case of no images, docker inspect runs with no arguments. Maybe it would be prudent to check for at least one image?

```
class DockerImages(DockerFactBase):
    """
    Returns ``docker inspect`` output for all Docker images.
    """

    def command(self) -> str:
        return "docker image inspect `docker images -q`"
```

wowi42:
It fix #1278

### DIFF
--- a/pyinfra/facts/docker.py
+++ b/pyinfra/facts/docker.py
@@ -49,7 +49,7 @@ class DockerContainers(DockerFactBase):
 
     @override
     def command(self) -> str:
-        return "docker container inspect `docker ps -qa`"
+        return ('container_ids=$(docker ps -qa) && [ -n "$container_ids" ] && docker container inspect $container_ids || echo "[]"')
 
 
 class DockerImages(DockerFactBase):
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
Finally got a chance to merge, looks good to me - thank you @wowi42!

Fizzadar:
Note: test issue unrelated, don't have perms so can't rebase but should be ok once landed.

PR #1354, opened 2025-05-04
Add podman connector

Griffoen:
The Podman CLI is compatible with the Docker CLI, so I refactored the docker connector to create the Podman connector.
This enables users of distros which prefer Podman over Docker to test their deployments more easily. 

The Podman connector works exactly like the docker connector, for example: `pyinfra @podman/fedora:42 dnf.packages packages=uv -y`

I copied the docker tests in `/tests/test_connectors/test_docker.py` for the Podman tests.
However, I omitted the `/test/end-to-end/test_e2e_docker.py` tests since I don't know how they are supposed to be set up, and I don't have docker installed to run the tests.


Griffoen:
The failing test (below) has nothing to do with this MR, and it succeeded after a re-run while developing this MR.
I don't know what is going on here, but it does not affect this MR as far as I can tell.
```
=========================== short test summary info ============================
FAILED tests/test_connectors/test_ssh.py::TestSSHConnector::test_get_file_sudo - AssertionError: expected call not found.
Expected: getfo('/tmp/pyinfra-e9c0d3c8ffca943daa0e75511b0a09c84b59c508', <MagicMock name='open()' id='139776354182096'>)
  Actual: not called.
========== 1 failed, 1169 passed, 12 deselected, 1 warning in 22.55s ===========
```

Griffoen:
Considering how the docker connector testing code works, and how the podman connector is basically a string substituted docker connector, I don't think the tests_docker.py can ever find any errors with the podman connector.

It may just be better to not include/adapt the tests_docker.py, and instead let the end2end tests verify no string substitution was missed or wrong.

Fizzadar (approved):
@Griffoen this is awesome, thank you so much for working on this. I think the testing is fine too even if duplicated it's only a tiny amount of time to execute and means we're covering any code drift. I just pushed up the e2e test as well for coverage there. Good to go!

As mentioned on Matrix this connector is going to be included in main because of it's overlap with the docker connector (and common interchangable use in real world).

Issue #1137, opened 2024-07-01
files.put(mode=True) always reports changed == True


cdleonard:
When using files.put to copy an executable script to the target it always shows up as "changed". This can slow down scripts which check the "changed" bit in order to decide if further changes are required.

This works fine without passing the mode argument.

This also works as expected if the mode argument is an explicit value like "400" or "500".

Happens with both pyinfra `3.0b4` and `v2.9.2`

PR #1363, opened 2025-06-01
Fix files.put(mode=True) (#1137) and update files.put docstring for similar mismatched type comparison issue

vram0gh2:
Addresses the type mismatch issue causing issue #1137 and documents related type mismatch pitfalls for the ``mode`` argument to ``files.put()``.


Fizzadar (approved):
Thank you for the PR, @vram0gh2!

PR #1366, opened 2025-06-02
Allow specifying custom tempdir for files.download

scy:
This introduces a new argument to the `files.download` operation: `temp_dir`, which allows the user to specify a custom temporary directory for downloading into.

I know that `TEMP_DIR` exists as an (undocumented) config option, but since `Host._get_temp_directory` is memoized, it's not feasible to set this option only for certain operations.

My use case is downloading very large files (> 100 GB) into specific volumes on the server, which are backed by different filesystems and have some constraints regarding free space. The default of downloading everything into `/tmp` (or whatever) doesn't work because it would fill up the root partition before the download even finishes, but setting `TEMP_DIR` to different directories throughout a single deploy is also not really possible without resorting to hacks, afaics.

To allow customizing the temporary directory, `Host.get_temp_filename` also gained a (keyword-only, optional) new argument `temp_directory`.

### DIFF
--- a/pyinfra/api/host.py
+++ b/pyinfra/api/host.py
@@ -328,12 +328,18 @@ class Host:
 
         return temp_directory
 
-    def get_temp_filename(self, hash_key: Optional[str] = None, hash_filename: bool = True):
+    def get_temp_filename(
+        self,
+        hash_key: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


scy:
The failing tests are unrelated to my changes.

PR #1371, opened 2025-06-09
Dont show change, if appending subset of existing groups

aaron-riact:
Dont show change, if appending subset of existing groups

Is that a different repo?  Looks like `append` is not mentioned on the website docs, but is in the api already
I get a bunch of errors on files not in this pr

### DIFF
--- a/pyinfra/operations/server.py
+++ b/pyinfra/operations/server.py
@@ -939,10 +939,13 @@ def user(
             args.append("-g {0}".format(group))
 
         # Check secondary groups, if defined
-        if groups and set(existing_user["groups"]) != set(groups):
+        if groups:
             if append:
-                args.append("-a")
-            args.append("-G {0}".format(",".join(groups)))
+                if not set(groups).issubset(existing_user["groups"]):
+                    args.append("-a")
+                    args.append("-G {0}".format(",".join(groups)))
+            elif set(existing_user["groups"]) != set(groups):
+                args.append("-G {0}".format(",".join(groups)))
 
         if comment and existing_user["comment"] != comment:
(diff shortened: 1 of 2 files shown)


Fizzadar:
>     * [ ]  Pull request includes documentation for any new/updated operations/facts
>       Is that a different repo?  Looks like `append` is not mentioned on the website docs, but is in the api already

Ah, this is a good flag. By that I mean the docstring (which is used to generate the docs): https://github.com/pyinfra-dev/pyinfra/pull/1371/files#diff-93ced8bf570709d17b4670036e1757810812c8af258c2740196f8ac2cd82a2e4L796. Made a note to make this clearer.

>     * [ ]  Type checking & code style passes (see `scripts/dev-lint.sh`)
>       I get a bunch of errors on files not in this pr

Mind sharing these? GitHub broke CI, now fixed. Logs would be great to fix this in a new issue if you don't mind!

Fizzadar (approved):
Nice add/fix here, @aaron-riact - thank you for the PR!

aaron-riact:
> Ah, this is a good flag. By that I mean the docstring (which is used to generate the docs): https://github.com/pyinfra-dev/pyinfra/pull/1371/files#diff-93ced8bf570709d17b4670036e1757810812c8af258c2740196f8ac2cd82a2e4L796. Made a note to make this clearer.

Ah, I see.  Thanks for clarifying)


> Mind sharing these? GitHub broke CI, now fixed. Logs would be great to fix this in a new issue if you don't mind!

I thought it was the github tests, but now that I look at them I see they passed first time.
If I run them locally, they are also all good.  Sorry, I may just have got confused with something else I was doing at the time.

PR #1372, opened 2025-06-11
Add atime and mtime support for files ops, start with files.put

vram0gh2:
Add concrete support for ``atime`` and ``mtime`` checking/setting in the ``files.put`` operation and provide infrastructure to add this time metadata support to other operations as well.

Includes a fix in ``facts/files.py`` to accommodate negative timestamps from ``stat`` on the remote host.

Added support for more ``os.stat_result`` file metadata to be provided in the test json files, by updating the mock ``os.stat`` patch used for testing.  This includes ``atime``, ``mtime``, but also the other members of the standard ``stat`` 10-tuple as well.

@Fizzadar: Looking for any feedback/implementation concerns, style changes you'd like, and how best to handle documenting the specifics and caveats.  I added a note in the operation docstring, but if ``atime`` and ``mtime`` support is implemented in other operations (*e.g.*, ``files.file``) these will be true there as well.  Is there a preferred way to break that information out into something outside of the function-specific docstring?

The expanded mock ``os.stat`` added a fair number of LOC, but it's also now equipped for future features/higher fidelity.

Also, I used the American spelling of "normalize" in one of the added support functions, happy to change it to British if preferred.


pyinfra/operations/files.py
```diff
@@ -778,6 +778,56 @@ def get(
             yield FileDownloadCommand(src, dest, remote_temp_filename=host.get_temp_filename(dest))
 
 
+def _canonicalize_timespec(field, local_file, timespec):
+    assert field == "atime" or field == "mtime"
```
Fizzadar: Using an enum for the field argument would be cleaner here + support type-checking.

pyinfra/operations/files.py
```diff
@@ -778,6 +778,56 @@ def get(
...
+            timespec_with_tz = timespec.replace(tzinfo=host.get_fact(Date).tzinfo)
+            return timespec_with_tz
+        else:
+            return timespec
+    elif isinstance(timespec, bool) and timespec:
+        lf_stat = os.stat(local_file)
+        if field == "atime":
+            return datetime.fromtimestamp(lf_stat.st_atime, tz=timezone.utc)
+        else:
+            return datetime.fromtimestamp(lf_stat.st_mtime, tz=timezone.utc)
```
Fizzadar:
```suggestion
        tstamp = if_stat.st_atime if field == "atime" else if_stat.st_mtime
        return datetime.fromtimestamp(tstamp, tz=timezone.utc)
```

Nitpicking, just a bit neater.

pyinfra/operations/files.py
```diff
@@ -802,6 +854,9 @@ def put(
     + create_remote_dir: create the remote directory if it doesn't exist
     + force: always upload the file, even if the remote copy matches
     + assume_exists: whether to assume the local file exists
+    + atime: value of atime the file should have, use ``True`` to match the local file
+    + mtime: value of atime the file should have, use ``True`` to match the local file
+    + timesrc: the source of the time value if atime or mtime are ``True``
```
Fizzadar: Leftover argument?

Fizzadar (commented):
Nice, this is really neat @vram0gh2! Sorry it took so long for me to do a proper review. Replying to your notes specifically:

> I added a note in the operation docstring, but if atime and mtime support is implemented in other operations (e.g., files.file) these will be true there as well. Is there a preferred way to break that information out into something outside of the function-specific docstring?

Currently not, and yeah that could be messy to duplicate everywhere. One option would be to put it into the top level docstring in `operations/files.py` such that it appears at the top of the [operation doc](https://docs.pyinfra.com/en/3.x/operations/files.html) and then reference that within each operation.

I do think this would be a great to expand to the other ops!

> The expanded mock os.stat added a fair number of LOC, but it's also now equipped for future features/higher fidelity.

👍 really nice, good add.

> Also, I used the American spelling of "normalize" in one of the added support functions, happy to change it to British if preferred.

All good, American makes sense as consistent w/Python, though I'm sure pyinfra contains a mix of both already 🙃

vram0gh2:
Thanks for the review, @Fizzadar !

I made changes/fixes as per the feedback .  Docstrings should now be fixed, and we're using an ``enum`` to indicate ``atime`` or ``mtime``.

There could be some stylistic room to argue in that I gave the enum a somewhat? long name of ``MetadataTimeField`` to be clear it wasn't referring to a portion of an ISO date string or components of a ``datetime``.  This results in a few spots where lines are wrapped.  I wouldn't object to shortening this if the ambiguity argument is outweighed by the desire for compactness.

If this is considered merge-ready otherwise, perhaps we can address the documentation issue if/when ``atime`` and ``mtime`` support is added to other ops and leave it where it is for now.

Fizzadar:
> Thanks for the review, @Fizzadar !
> 
> I made changes/fixes as per the feedback . Docstrings should now be fixed, and we're using an `enum` to indicate `atime` or `mtime`.

🙏 thank you!

> There could be some stylistic room to argue in that I gave the enum a somewhat? long name of `MetadataTimeField` to be clear it wasn't referring to a portion of an ISO date string or components of a `datetime`. This results in a few spots where lines are wrapped. I wouldn't object to shortening this if the ambiguity argument is outweighed by the desire for compactness.

I think the name is good as-is, rather a more explicit name than something shorter for styles sake!

> If this is considered merge-ready otherwise, perhaps we can address the documentation issue if/when `atime` and `mtime` support is added to other ops and leave it where it is for now.

👍 sounds good to me

Fizzadar (approved):
Awesome stuff, @vram0gh2! When properly at keyboard I’ll fix the conflicts and merge.

Fizzadar:
Augh damn Python 3.10 and it's useless isoformat parsing, can't believe it took until 3.11 to fix that!

PR #1375, opened 2025-06-18
avoid use of non-POSIX "chown -n" in files.block

bad:
chown(1) always recognises numeric user and group ids.

While here:
- run both versions of the stat(1) invocations in the $() command substitution for chown instead of invoking chown(1) with bad arguments.
- always use numeric ids as file owners are not required to exist in the user and group databases.  This also speeds up the operation as the pointless uid/gid translation doesn't have to be performed multiple times.

Not sure why you're faffing about with the non-portable stat(1) when `ls -n` can reliably produce the required information, but that's a different cat to skin.

- [NA] Pull request includes tests for any new/updated operations/facts
- [NA] Pull request includes documentation for any new/updated operations/facts

### DIFF
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -1548,12 +1548,12 @@ def block(
         "2>/dev/null || stat -f %Lp",
         q_path,
         ") $OUT && ",
-        '(chown $(stat -c "%U:%G"',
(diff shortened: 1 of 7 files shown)


Fizzadar (approved):
Thanks for fixing this, @bad!

PR #1380, opened 2025-06-22
feat: Try multiple hash funcs files.get and files.put

mrkbac:
Support more hash functions for `files.put`, `files.get`, on machine I'm working with does not support the sha1 commands, So with this PR `md5`, `sha1`,`sha264` are tries in this order

Maybe this should be configurable?


pyinfra/operations/files.py
```diff
@@ -770,12 +772,24 @@ def get(
...
     # Remote file exists - check if it matches our local
     else:
-        local_sum = get_file_sha1(dest)
-        remote_sum = host.get_fact(Sha1File, path=src)
-
-        # Check sha1sum, upload if needed
-        if local_sum != remote_sum:
+        sum_match = False
+        for fact, get_sum in [
+            # md5 is most common, so check it first
```
Fizzadar: I think we should use sha1 as the first/default here to match the current implementation.

pyinfra/operations/files.py
```diff
@@ -904,10 +918,20 @@ def put(
 
     # File exists, check sum and check user/group/mode if supplied
     else:
-        remote_sum = host.get_fact(Sha1File, path=dest)
-
-        # Check sha1sum, upload if needed
-        if local_sum != remote_sum:
+        sum_match = False
+        for fact, get_sum in [
```
Fizzadar: Probably makes sense to put this block into a util function and de-dupe with the above.

Fizzadar (commented):
Nice, this is awesome stuff @mrkbac! Two minor comments to addres 🙏

pyinfra/operations/files.py
mrkbac: I could create a fact that compares local and remote files and simply returns 'True' or 'False'. This could be useful for others as well. What do you think?

mrkbac: I guess it's not that easy, because I would call a fact in a fact

Fizzadar (approved):
Thank you for this, @mrkbac, looks great! I made a mess by pushing my own md5/sha256 changes up yesterday so fixed that up ready to merge, apologies!

PR #1382, opened 2025-06-25
facts.server, operations.server: Make Mount fact and operation portable

DtxdF:
Using `platform.system(...)` is not a good idea because it runs on the local machine and we need to take into account the operating system of the remote machine.

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 import os
-import platform
 import re
 import shutil
 from datetime import datetime
@@ -13,6 +12,7 @@ from dateutil.parser import parse as parse_date
 from distro import distro
 from typing_extensions import TypedDict, override
 
+from pyinfra import host
 from pyinfra.api import FactBase, ShortFactBase
 from pyinfra.api.util import try_int
 from pyinfra.facts import crontab
@@ -207,8 +207,10 @@ class Mounts(FactBase[Dict[str, MountsDict]]):
     default = dict
 
     @override
-    def command(self):
-        if platform.system() == "FreeBSD":
+    def command(self) -> str:
+        self._kernel = host.get_fact(Kernel)
+
+        if self._kernel.strip() == "FreeBSD":
             return "mount -p --libxo json"
         else:
             return "cat /proc/self/mountinfo"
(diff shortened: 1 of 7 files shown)


pyinfra/facts/server.py
```diff
@@ -208,7 +208,7 @@ class Mounts(FactBase[Dict[str, MountsDict]]):
 
     @override
     def command(self):
-        if platform.system() == "FreeBSD":
+        if host.get_fact(Kernel).strip() == "FreeBSD":
```
Fizzadar: Unfortunately this won’t work (can’t call facts from facts). Will need to include the check condition within the command itself.

Also noting to self should make this error properly here, as shown by the failing test case.

jgelens: Somehow the check `get_fact` call does work though. It returns what it should on my remote FreeBSD machine:

```
pyinfra  inventories/physical.py fact server.Mounts
--> Fact data for: server.Mounts
{
    "xx.xx.xx.xx": {
        "/": {
            "device": "zroot/ROOT/FreeBSD",
            "type": "zfs",
...
```

WIthout this change it returns an error:
```
[x] Loaded fact server.Os
[x] cat: /proc/self/mountinfo: No such file or directory
[x] Error: could not load fact: server.Mounts
```

Unintentional side effect?

Fizzadar: Interesting! Not intentional (which is why unit tests fail since they don’t mock this path), but this feels like something that would be very useful..

Need to have a proper think about the implications, but this would be a good addition I think.

DtxdF: Yes, I had doubts about that change but I have the same result as @jgelens comments, it works. The other solution is to put a conditional check through `sh(1)` in the `command(...)` function but using pyinfra for this is better.

This is what happens when I run pyinfra (excerpt):

```console
...
--> Gathering facts...
    [control-centralita] Loaded fact server.Kernel
    [control-centralita] Loaded fact server.Kernel
    [control-centralita] Loaded fact server.Mounts
...
```

Fizzadar: Figured it out! So it does _appear_ to work, and indeed it will in nearly all cases. It falls apart when you want to get a fact from a different host that isn't the current one. Something like:

```py
from pyinfra import inventory

other_host_kernel = inventory.get_host("main").get_fact(Kernel)
```

Which breaks if you use the host context within the fact. The fact module is super old and still takes `state` and `host` arguments to each function. I don't see any reason this can't be tweaked to set the state/host context before gathering facts, which would then work in all cases. Working on this today to unblock this/release.

Fizzadar: https://github.com/pyinfra-dev/pyinfra/commit/9bc54b8c38b90a5a88e7eea8671477b44b7efd8c cleans the use of state/host up in the API. Facts should only ever (be definiton) be called with the host context set, so this is perfectly safe, same as code under `operations`.

Fizzadar: And https://github.com/pyinfra-dev/pyinfra/pull/1382/commits/b5196274a7b174b29125f118e73359a1066633f5 to get the tests working.

PR #1383, opened 2025-06-26
Fix user addition/modification operations for FreeBSD.

jgelens:
- The user addition/modification operations for FreeBSD were not correct. This patch fixes all issues I found (see man page https://man.freebsd.org/cgi/man.cgi?pw(8))

- On FreeBSD `/etc/shadow` does not exist, the file is called `/etc/master.passwd`.
The line for that is 6 characters too long to pass the flake8 linter. But adding newlines here will become rather messy in my opinion. Any thoughts?

### DIFF
--- a/pyinfra/facts/server.py
+++ b/pyinfra/facts/server.py
@@ -526,10 +526,10 @@ class Users(FactBase):
         for i in `cat /etc/passwd | cut -d: -f1`; do
             ENTRY=`grep ^$i: /etc/passwd`;
             LASTLOG=`(((lastlog -u $i || lastlogin $i) 2> /dev/null) | grep ^$i | tr -s ' ')`;
-            PASSWORD=`grep ^$i: /etc/shadow | cut -d: -f2`;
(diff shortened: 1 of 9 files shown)


pyinfra/facts/server.py
```diff
@@ -526,7 +526,7 @@ def command(self):
         for i in `cat /etc/passwd | cut -d: -f1`; do
             ENTRY=`grep ^$i: /etc/passwd`;
             LASTLOG=`(((lastlog -u $i || lastlogin $i) 2> /dev/null) | grep ^$i | tr -s ' ')`;
-            PASSWORD=`grep ^$i: /etc/shadow | cut -d: -f2`;
+            PASSWORD=`(grep ^$i: /etc/shadow || grep ^$i: /etc/master.passwd) 2> /dev/null | cut -d: -f2`;
```
jgelens, line 529: Too long for the linter. Any idea how to improve this without making the code unclear? Maybe adding an exception?

Fizzadar (approved):
Awesome stuff, thanks for fixing and testing this @jgelens!

Issue #985, opened 2023-05-26
show diff of files which will be modified with --dry


khimaros:
## Is your feature request related to a problem? Please describe
it would be helpful to see specifically what changes will be made to a file before doing it. the only tool i've seen do this well is `slack`, but i miss it every time.

## Describe the solution you'd like
when running in --dry mode (or perhaps with a new flag --diff), show the files which would be changed by a deploy and the differences between the old and new file.

Fizzadar:
Love this idea. As part of tackling #805 I am going to overhaul the diff output to be more verbose (think terraform).

xvello:
I managed to hack a prototype by wrapping `files.put` and running the diff there. It's inefficient because it needs to always pull the remote file (even if the sha1 matches), implementing it within `files.put` would remove this issue.

![image](https://github.com/user-attachments/assets/a54f9786-a970-4603-b317-ecb7a534ce6c)

<details><summary>Here's my current code, feel free to reuse under MIT (click to unfurl). I'd be happy to try cleaning it up and turning it into a PR after we decide how the output should look and how to turn it on.</summary>
&nbsp;

```python
import difflib
from io import BytesIO
from typing import IO, Any, Generator, Optional

import click
from pyinfra import host, logger
from pyinfra.api.operation import OperationMeta
from pyinfra.api.util import get_file_io
from pyinfra.facts.server import Hostname
from pyinfra.operations import files


def put(
    src: str | IO[Any], dest: str, name: Optional[str] = None, present: bool = True, **kwargs: Any
) -> OperationMeta:
    if not present:
        return files.file(path=dest, present=False, name=(name or f"Delete {dest}"))

    current_contents = BytesIO()
    current_lines: list[str] = []
    try:
        if host.get_file(dest, current_contents):
            current_lines = current_contents.getvalue().decode("utf-8").splitlines(keepends=True)
    except FileNotFoundError:
        pass

    result = files.put(src, dest, name=name, **kwargs)

    if result.changed:
        hostname = host.get_fact(Hostname)
        if current_lines:
            logger.info(f"\n    Will modify {click.style(dest, bold=True)} on {hostname}:")
        else:
            logger.info(f"\n    Will create {click.style(dest, bold=True)} on {hostname}:")

        with get_file_io(src, "r") as f:
            desired_lines = f.readlines()
        for line in generate_color_diff(current_lines, desired_lines):
            logger.info(f"  {line}")
        logger.info("")

    return result


[...]

PR #1385, opened 2025-06-27
Show a diff when a file will be or gets modified

jgelens:
Now shows a diff of files which will be modified, only when its checksum its different.

Thanks @xvello (https://github.com/pyinfra-dev/pyinfra/issues/985#issuecomment-2229224464) for the code!

This solves #985

### DIFF
--- a/pyinfra/api/config.py
+++ b/pyinfra/api/config.py
@@ -53,6 +53,7 @@ class ConfigDefaults:
     IGNORE_ERRORS: bool = False
     # Shell to use to execute commands
     SHELL: str = "sh"
+    DIFF: bool = False
 
 
 config_defaults = {key: value for key, value in ConfigDefaults.__dict__.items() if key.isupper()}
--- a/pyinfra/facts/files.py
+++ b/pyinfra/facts/files.py
@@ -657,3 +657,18 @@ class Block(FactBase):
         if output and (output[0] == f"{MISSING}{self.path}"):
             return None
         return output
+
+
+class FileContents(FactBase):
+    """
+    Returns the contents of a file as a list of lines. Works with both sha1sum and sha1. Returns
+    ``None`` if the file doest not exist.
+    """
+
+    @override
+    def command(self, path):
+        return make_formatted_string_command("cat {0}", QuoteString(path))
+
+    @override
+    def process(self, output):
+        return output
--- a/pyinfra/operations/files.py
+++ b/pyinfra/operations/files.py
@@ -1033,21 +1033,22 @@ def put(
     # File exists, check sum and check user/group/mode/atime/mtime if supplied
     else:
         if not _file_equal(local_sum_path, dest):
-            # Generate diff when contents change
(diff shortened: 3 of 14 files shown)


jgelens:
Example output:

Screenshot 2025-06-27 at 09 27 10

jgelens:
Working on some improvements to not use `host.get_file'

Fizzadar:
This is awesome, just noting I'm not forgetting it. Need to handle when to show/not show the diff (ie when `-v`). Going to get the pending stuff released and get then think on that to land this.

jgelens:
Right, maybe it would be best to add a new flag for that. I always would like to see the diff, even when not using `-v`. So maybe we could add a `--diff` flag for this; similar to the same Ansible feature.

I was also thinking to add a global flag, which also can be used to override single operations. In case people do not want to show it for e.g. editing a file with passwords.

xvello:
I also think it's good to show the diffs without verbose logging, that's how I currently use it. A command line switch could indeed be enough.

@jgelens thanks for pushing this over the finish line. When you're pushing other changes, would you mind adding a `Co-authored-by: Xavier Vello <xavier.vello@gmail.com>` line in the git commit message of the first commit, for attribution?

jgelens:
> @jgelens thanks for pushing this over the finish line. When you're pushing other changes, would you mind adding a `Co-authored-by: Xavier Vello <xavier.vello@gmail.com>` line in the git commit message of the first commit, for attribution?

Of course, added.

Fizzadar:
> Right, maybe it would be best to add a new flag for that. I always would like to see the diff, even when not using `-v`. So maybe we could add a `--diff` flag for this; similar to the same Ansible feature.

👍  for `--diff`, good call.

goetzk:
I've been waiting on this PR, really excited to see it so close to inclusion :) Thanks heaps to everyone involved!

Fizzadar:
@jgelens are you happy for me to finish this up with the `--diff` arg?

jgelens:
Yeah sure, no problem. Otherwise I can do it next week or so.

jgelens:
@Fizzadar I'm not sure if adding it to the Config object is enough. See https://github.com/pyinfra-dev/pyinfra/pull/1385/commits/2381ab93f8b472df8fa3398938b80d4d0a0ce96c

jgelens:
> @Fizzadar I'm not sure if adding it to the Config object is enough. See [2381ab9](https://github.com/pyinfra-dev/pyinfra/commit/2381ab93f8b472df8fa3398938b80d4d0a0ce96c)

Can you confirm whether this is the right way to implemented flag like this?

Fizzadar:
@jgelens sorry for the delay - I think this is correct, thank you!

It's made me think we should (not now) move the -v flags over from state -> config to match, and also that `--diff` should probably set `-v=1` if not set as well (to show the "blah is already installed" etc).

Fizzadar (approved):
Awesome stuff, @jgelens, very excited to land this!

PR #1391, opened 2025-07-04
Add extra optional arguments for wget or curl for files.download

jgelens:
This add an optional dict which is passed to curl and wget as extra argument on its cli.
It's useful to provide for example authentication to the commands (for curl: --user foo:bar). 
Another example is that it can be used to make S3 requests using curl's --aws-sigv4 option.


Fizzadar (approved):
Neat! Super flexible add here, love it. Thanks for the PR @jgelens!

PR #1397, opened 2025-07-12
Add support for enabling OpenRC service on a specific runlevel

sengo4hd:

Hello and thank you for your work on this project :)
I find the design to be very clear and well-thought, making it easy to engage with contributing.

I want to use pyinfra to enable a nftables service on the boot runlevel with OpenRC. I noticed that the openrc.service operation would always trigger a change on every run even though the service was never really enabled on this runlevel. This is because the change was always happening on the default runlevel, ignoring the configuration.

I would like to propose these changes:
- A noop message for when a service is already enabled/disabled to help with troubleshooting.
- Support for enabled/disabling services on a specific runlevel, as expected from the documentation.

I did not find it necessary to update the openrc.service documentation.
I also fixed a small typo in a link in the contributing documentation.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -78,7 +78,7 @@ pytest -m end_to_end_docker
 
 ## Documentation
 
-Documentation changes should be made in the [pyinfra git repository](https://github.com/pyinfra-dev/pyinfa) - the repository `docs.pyinfra.com` should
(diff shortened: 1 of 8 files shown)


Fizzadar (approved):
Very nice, thank you @sengo4hd!

PR #1404, opened 2025-07-21
operations/crontab: fix replace containing newline, add test

Nananas:
This replace is triggered e.g. when replacing an (pre-)existing crontab line without name to one maintained by pyinfra, this time with a pyinfra-name.

The sed replace command contains a newline character that should stay as a '\n' and not an actual newline, as sed does not support a multi-line script.


Fizzadar (approved):
Thank you for the fix and test on this, @Nananas 🙏!

PR #1422, opened 2025-08-10
Allow SCP as file transfer protocol when using SSH connector

DonDebonair:
This PR adds an option to use SCP as the file transfer protocol when using the SSH connector. I took [this library](https://github.com/jbardin/scp.py) as inspiration and wrote a very minimal version of it that is only Python 3 compatible (the original lib is Py2/3 compatible, which makes it overly complex).

I did not implement all of the methods of the original library, and chose to only implement `getfo()` and `putfo()`, because that is what the SSH connector needs. I created a Protocol to make the 2 options compatible in a type-safe way.

2 important caveats:

1. I did not add tests. The only way to test this in a useful way, is to have an OpenSSH server running to test the real upload/download of files. That's also what the original library does. This would not be a unit test, but more of an integration test. It could feasibly be done using Docker, but maybe you have another suggestion @Fizzadar ? I know pyinfra has end2end tests, maybe I should piggyback off those?
2. The SCP protocol is complicated, so most of my code was copy/pasted, cleaned up and properly typed by me. I do not claim to understand the code that well yet in case issues arise. I personally don't think that's a real problem because the SCP protocol is optional to use and pyinfra will still default to SFTP.

Let me know what you think!

### DIFF
--- a/pyinfra/connectors/scp/__init__.py
+++ b/pyinfra/connectors/scp/__init__.py
@@ -1,0 +1,1 @@
+from .client import SCPClient  # noqa: F401
--- a/pyinfra/connectors/scp/client.py
+++ b/pyinfra/connectors/scp/client.py
@@ -1,0 +1,204 @@
+from __future__ import annotations
+
+import ntpath
+import os
+from pathlib import PurePath
+from shlex import quote
+from socket import timeout as SocketTimeoutError
+from typing import IO, AnyStr
+
+from paramiko import Channel
+from paramiko.transport import Transport
+
+SCP_COMMAND = b"scp"
+
+
+# Unicode conversion functions; assume UTF-8
+def asbytes(s: bytes | str | PurePath) -> bytes:
+    """Turns unicode into bytes, if needed.
+
+    Assumes UTF-8.
+    """
+    if isinstance(s, bytes):
+        return s
+    elif isinstance(s, PurePath):
+        return bytes(s)
+    else:
+        return s.encode("utf-8")
+
+
+def asunicode(s: bytes | str) -> str:
+    """Turns bytes into unicode, if needed.
+
+    Uses UTF-8.
+    """
+    if isinstance(s, bytes):
+        return s.decode("utf-8", "replace")
+    else:
+        return s
+
+
+class SCPClient:
+    """
+    An scp1 implementation, compatible with openssh scp.
+    Raises SCPException for all transport related errors. Local filesystem
+    and OS errors pass through.
+
+    Main public methods are .putfo and .getfo
+    """
+
+    def __init__(
+        self,
+        transport: Transport,
+        buff_size: int = 16384,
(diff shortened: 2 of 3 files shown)


pyinfra/connectors/scp.py
```diff
@@ -0,0 +1,228 @@
...
+from paramiko import Channel
+from paramiko.transport import Transport
+
+SCP_COMMAND = b"scp"
+
+# this is quote from the shlex module, added in py3.3
+_find_unsafe = re.compile(rb"[^\w@%+=:,./~-]").search
+
+
+def _sh_quote(bs: bytes) -> bytes:
```
Fizzadar: I think this function can be replaced with `shlex.quote` from the stdlib.

Fizzadar (commented):
This is great, thank you @DonDebonair! One redundant function and also a request: can we move the scp implementation into it's own module similar to ./sshuserclient?

pyinfra/connectors/scp.py
DonDebonair: I didn't use `shlex.quote` because that function expects a string and outputs a string. In this module we're almost exclusively working with byte arrays. Hence the adaptation.

I could probably convert byte arrays to string and then use `shlex.quote`, but then I'd have to take care of the encoding each time, which is different between windows and posix. So it seems easier to just work with bytes here and convert from/to string at the boundaries.

Let me know if you still want me to change it

DonDebonair:
> This is great, thank you @DonDebonair! One redundant function and also a request: can we move the scp implementation into it's own module similar to ./sshuserclient?

I assume you mean moving the `scp.py` module into it's own package? Don't mean to be pedantic, but it helps me to get terminology straight, and in Python a module is a `.py` file and a package is a directory with an `__init__.py` file in it (and confusingly also an installable thing share on PyPI).

I will move the `scp.py` into it's own package.

DonDebonair:
I added the SCP client to it's own package.

Fizzadar:
> I assume you mean moving the `scp.py` module into it's own package? Don't mean to be pedantic, but it helps me to get terminology straight, and in Python a module is a `.py` file and a package is a directory with an `__init__.py` file in it (and confusingly also an installable thing share on PyPI).

Heh, my Python is a little rusty at the moment, my bad.

> I will move the `scp.py` into it's own package.

Awesome thank you.

pyinfra/connectors/scp.py
DonDebonair: Nevermind, I was wrong. I reviewed the PR again myself, and you were right. I can use `shlex.quote`. I have pushed the new code.

DonDebonair:
Checks should pass now

DonDebonair:
Anything still blocking this @Fizzadar?

Fizzadar (approved):
Amazing, thank you for implementing this @DonDebonair (and putting up with my slow reviews)!

Issue #1427, opened 2025-08-12
files.block placement of temporary output file has flaws


morrison12:
## Describe the bug
1. There is a missing colon in what should be a fall-back to `/tmp/` if `TMPDIR` is not set
2. As with #1425, the fallback should be to `config.TEMP_DIR` not a hard-coded `/tmp`

## To Reproduce
Consequence of #1425 

## Expected behavior
Temporary files should be put in the correct place.

## Meta
    System: Darwin
      Platform: macOS-14.7.7-arm64-arm-64bit
      Release: 23.6.0
      Machine: arm64
    pyinfra: v3.4.1
      click: v8.2.1
      click: v8.2.1
      click: v8.2.1
      distro: v1.9.0
      gevent: v25.5.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      typeguard: v4.4.4
      typing-extensions: v4.14.1
    Executable: /Users/someone/.local/bin/pyinfra
    Python: 3.12.11 (CPython, Clang 20.1.4 )

Fizzadar:
Fixed via https://github.com/pyinfra-dev/pyinfra/pull/1428

PR #1428, opened 2025-08-12
fix #1425 and #1427 to get the temp directory in a way that respects config.TEMP_DIR…

### LINKED ISSUE #1425: Location of "askpass" file respects target's TMPDIR but not config.TEMP_DIR
## Describe the bug
In a setup where the target precludes execution of files from `/tmp` and `TMPDIR` is not yet set (on the target), even with `config.TEMP_DIR` set to a useful path (and #1424 fixed), the _askpass_ file is still placed in `/tmp` and thus cannot be executed.

## To Reproduce
Try to  execute an operation  with `_sudo=True` and an already set password against a target that prevents execution of files from `/tmp` (and without `TMPDIR` set).  The login will fail because the _askpass_ file can't be executed.  While this is not a useful long-term setup, it can happen when a machine has not yet been configured.

## Expected behavior
The operation will work flawlessly.

## Meta
    System: Darwin
      Platform: macOS-14.7.7-arm64-arm-64bit
      Release: 23.6.0
      Machine: arm64
    pyinfra: v3.4.1
      click: v8.2.1
      click: v8.2.1
      click: v8.2.1
      distro: v1.9.0
      gevent: v25.5.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      typeguard: v4.4.4
      typing-extensions: v4.14.1
    Executable: /Users/james/.local/bin/pyinfra
    Python: 3.12.11 (CPython, Clang 20.1.4 )

## Possible cause

'/tmp' is hard-coded as the backstop to `TMPDIR` below
https://github.com/pyinfra-dev/pyinfra/blob/c557b997646948e904d758ee44fbbd0baccc7198/pyinfra/connectors/util.py#L24-L35

morrison12:
Use `host.get_temp_directory` instead of hard-coding `/tmp`.

### DIFF
--- a/src/pyinfra/api/host.py
+++ b/src/pyinfra/api/host.py
@@ -313,7 +313,7 @@ class Host:
         )
 
     @memoize
-    def _get_temp_directory(self):
+    def get_temp_directory(self):
         temp_directory = self.state.config.TEMP_DIR
 
         if temp_directory is None:
@@ -339,7 +339,7 @@ class Host:
         Generate a temporary filename for this deploy.
         """
 
-        temp_directory = temp_directory or self._get_temp_directory()
+        temp_directory = temp_directory or self.get_temp_directory()
 
         if not hash_key:
             hash_key = str(uuid4())
--- a/src/pyinfra/api/host.py
+++ b/src/pyinfra/api/host.py
@@ -313,7 +313,7 @@ class Host:
         )
 
     @memoize
-    def get_temp_directory(self):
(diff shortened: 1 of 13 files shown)


Fizzadar (approved):
Thank you for fixing this up, @morrison12! Sorry it took a while but I have finally rebased it on the 3.x rejig work so good to go now :)

PR #1430, opened 2025-08-12
update config from values from config.py (fix #1424)

### LINKED ISSUE #1424: CLI doesn't seem to use some definitions made in config.py
## Describe the bug
While the `pyinfra` CLI appears to execute the `config.py` file if it finds one in the current working directory when the CLI is invoked, it  appears that some definitions in the file are not used, e.g. `config.TEMP_DIR`, despite the [documentation](https://docs.pyinfra.com/en/3.x/using-operations.html#the-config-object) suggesting they would be.

## To Reproduce
WIth `config.py` and `test.py` as follows
```python

# config.py
TEMP_DIR = "foo"

# test.py
import io

from pyinfra import config.

value = io.StringIO()
value.write(config.TEMP_DIR or "<not set>")
files.put(value, "test.txt")

```
then 
```sh
pyinfra @local test.py
cat test.txt
```
gives
```
<not set>
```

## Expected behavior
The output should be `foo`

## Meta
```
    System: Darwin
      Platform: macOS-14.7.7-arm64-arm-64bit
      Release: 23.6.0
      Machine: arm64
    pyinfra: v3.4.1
      click: v8.2.1
      click: v8.2.1
      click: v8.2.1
      distro: v1.9.0
      gevent: v25.5.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      typeguard: v4.4.4
      typing-extensions: v4.14.1
    Executable: /Users/someone/.local/bin/pyinfra
    Python: 3.12.11 (CPython, Clang 20.1.4 )
```

## Possible cause

In the code below, it appears that while the code is executed, the values set in the file are discarded instead of being added into the `config` object (as an aside, it also appears the locking is done before the CLI over-rides are applied) but I may not understand how this is supposed to work

https://github.com/pyinfra-dev/pyinfra/blob/c557b997646948e904d758ee44fbbd0baccc7198/pyinfra_cli/main.py#L553-L563

morrison12:
also lock _after_ applying CLI over-rides.

The current code only takes values defined in `config.py` if they are variables for which we have a default: this as assumed to be the intended limitation. If not, `update` can be simplified accordingly.

Intended to fix #1424 

- [ x] Pull request is based on the default branch (`3.x` at this time)
- [x ] Pull request includes tests for any new/updated operations/facts
- [x ] Tests pass (see `scripts/dev-test.sh`)
- [ x] Type checking & code style passes (see `scripts/dev-lint.sh`)

### DIFF
--- a/pyinfra/api/config.py
+++ b/pyinfra/api/config.py
@@ -232,3 +232,8 @@ class Config(ConfigDefaults):
 
     def copy(self) -> "Config":
         return Config(**dict(self.get_current_state()))
+
+    def update(self, values: dict) -> None:
+        for key, value in values.items():
+            if key in ConfigDefaults.__dict__:
+                setattr(self, key, value)
--- a/pyinfra_cli/main.py
+++ b/pyinfra_cli/main.py
@@ -556,11 +556,8 @@ def _set_config(
     if state.cwd:
         config_filename = path.join(state.cwd, config_filename)
     if path.exists(config_filename):
-        exec_file(config_filename)
-
-    # Lock the current config, this allows us to restore this version after
-    # executing deploy files that may alter them.
-    config.lock_current_state()
+        values = exec_file(config_filename)
+        config.update(values)
 
     # Arg based config overrides
     if sudo:
@@ -582,6 +579,10 @@ def _set_config(
 
     if fail_percent is not None:
         config.FAIL_PERCENT = fail_percent
+
+    # Lock the current config, this allows us to restore this version after
+    # executing deploy files that may alter them.
+    config.lock_current_state()
 
     return config
 


Fizzadar (commented):
I believe this is intentional, setting the config variables as top level was [dropped in v2](https://github.com/pyinfra-dev/pyinfra/blob/2.x/CHANGELOG.md#v20) (_Config variables must now be set on the global config object_), these should now always be set on the context ie:

```py
# config.py
from pyinfra import config
config.SUDO = True
```

morrison12:
Apologies: I somehow completely missed that.
-james

Fizzadar:
Closing per above :)

PR #1437, opened 2025-08-22
Manage deb822 format and apt-key removal in trixie

maisim:
WIP

### DIFF
--- a/src/pyinfra/facts/apt.py
+++ b/src/pyinfra/facts/apt.py
@@ -7,7 +7,6 @@ from typing_extensions import TypedDict, override
 from pyinfra.api import FactBase
 
 from .gpg import GpgFactBase
-from .util import make_cat_files_command
 
 
 def noninteractive_apt(command: str, force=False):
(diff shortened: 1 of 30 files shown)


maisim:
Just to let know you I am working on, comments are welcome

pyinfra/facts/apt.py
```diff
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
 import re
+from typing import List
```
DonDebonair: You don't actually need to import this, as [it's deprecated](https://docs.python.org/3/library/typing.html#typing.List). You should use `list` now as type annotation

pyinfra/facts/apt.py
```diff
@@ -60,44 +60,176 @@ def parse_apt_repo(name):
     }
 
 
-class AptSources(FactBase):
+def parse_deb822_stanza(lines: list[str]):
```
DonDebonair: Add return type annotation.

pyinfra/facts/apt.py
```diff
@@ -60,44 +60,176 @@ def parse_apt_repo(name):
...
+                        "type": _type,
+                        "url": uri,
+                        "distribution": suite,
+                        "components": components,
+                    }
+                )
+    return repos
+
+
+def parse_apt_list_file(lines: List[str]):
```
DonDebonair: Add return type annotation, use `list[str]` for lines

pyinfra/facts/apt.py
```diff
@@ -60,44 +60,176 @@ def parse_apt_repo(name):
...
+        line = raw.strip()
+        if not line or line.startswith("#"):
+            continue
+        repo = parse_apt_repo(line)
+        if repo:
+            repos.append(repo)
+    return repos
+
+
+def parse_deb822_sources_file(lines: List[str]):
```
DonDebonair: Add return type annotation, use `list[str]` for lines

pyinfra/facts/apt.py
```diff
@@ -60,44 +60,176 @@ def parse_apt_repo(name):
...
+    """Parse a full deb822 ``.sources`` file.
+
+    Splits on blank lines into stanzas and uses ``parse_deb822_stanza`` for each
+    stanza. Returns a combined list of repo dicts for all stanzas.
+    """
+    repos = []
+    stanza: List[str] = []
```
DonDebonair:
```suggestion
    stanza: list[str] = []
```

pyinfra/operations/apt.py
```diff
@@ -105,16 +194,22 @@ def key(src: str | None = None, keyserver: str | None = None, keyid: str | list[
...
-            yield "apt-key adv --keyserver {0} --recv-keys {1}".format(
-                keyserver,
-                " ".join(needed_keys),
+            dest_path = _derive_dest_from_src_and_keyids(None, needed_keys, dest)
+            joined = " ".join(needed_keys)
+            # Use a temporary GNUPGHOME so we don't pollute the system/user keyring,
+            # then export and dearmor to the APT keyring destination.
+            yield (
+                "sh -c 'set -e;"
+                " install -d -m 0755 /etc/apt/keyrings;"
```
DonDebonair: Why not use a fact to check if that dir already exists and only emit a `files.directory` command if it doesn't?

DonDebonair:
Hey @maisim, I hope you don't mind me leaving some comments. I agreed with @Fizzadar to help him out a bit with reviewing PRs.

This is great work! Support for the new DEB822-style apt sources and a replacement for `apt-key` are sorely needed. `apt-key` has in fact been removed in the latest stable Debian release (Trixie).

I have a couple of general suggestions, and I'll leave it to @Fizzadar if he agrees:

1. I would split this PR into 2: one for supporting DEB822 apt sources and one for replacing `apt-key`. Makes it easier to review and keeps PRs nicely compact
2. The work related to DEB822 apt sources could really benefit from some unit tests. There is a lot of parsing going on, which is both error-prone and easy to test. Also `AptSources.process(...)` is a complex function, trying to process multiple files at once. The algorithm that you wrote (using file markers, very neat!) looks sound to me, but it took me quite some time to truly understand what is going on. Writing one or more tests for that helps with both understanding what's going on and preventing bugs.
3.  `parse_deb822_stanza` and `parse_apt_list_file` both return a list of dicts. I think there is an opportunity for type-safety here, because the fields that define an apt source [are well defined](https://manpages.debian.org/trixie/apt/sources.list.5.en.html), even down to the [_options_](https://manpages.debian.org/trixie/apt/sources.list.5.en.html#THE_DEB_AND_DEB-SRC_TYPES:_OPTIONS). So my suggestion would be for both functions to return `list[AptRepo]` where `AptRepo` is a dataclass. This dataclass can even be reused across `.sources` and `.list` style definitions. I think the only real difference would be that for the latter, the `components` field would only ever have 1 entry.
Ideally, I would even say that the `AptSources` fact should return a `list[AptRepo]`, but as that would break backwards compatibility, we could add an argument to that fact to optionally return `list[AptRepo]` instead of `list[dict]` (something like `return_type_safe: bool = False`)

What do you think @maisim @Fizzadar ?

maisim:
Hi @DonDebonair !
Thank you very much for your review! I agree with your comments. Regarding the 2-PR split, I need to get back into it; it may indeed be possible to take care of the new format support in a PR and then the changes in the apt part.

DonDebonair:
Let me know if you need any help @maisim !

maisim:
Hi @DonDebonair ,

Big changes here, this is still a WIP. Don't waste your time and let me double check tomorrow before :)
Thanx again for your feedback, it gave me energy to continue the work :+1:

maisim:
It's a big all in one push but i will take care to split all this when I will get this all working ;)

maisim:
I started separating this into several PRs, the first one is here #1460

maisim:
replaced by 3 prs:
https://github.com/pyinfra-dev/pyinfra/pull/1460
https://github.com/pyinfra-dev/pyinfra/pull/1465
https://github.com/pyinfra-dev/pyinfra/pull/1487

PR #1441, opened 2025-08-31
facts/server: support negative value in sysctl

imlonghao:

Some sysctl value can be negative, for example `net.ipv4.tcp_adv_win_scale = -2`

Current facts/server/Sysctl parse -2 as string, but operations/server/Sysctl parse it as int


Fizzadar (approved):
Nice fix - thank you @imlonghao!

Issue #1121, opened 2024-06-07
`files.copy`


yellowhat:
## Is your feature request related to a problem? Please describe

Hi,
I would like to copy a file on another path on the targets.

Ansible provides `ansible.builtin.copy` for this kind of operation:

```yaml
    - name: Kubernetes - .kube/config (user)
      ansible.builtin.copy:
        src: /etc/rancher/k3s/k3s.yaml
        dest: "~{{ ansible_user }}/.kube/config"
        owner: "{{ ansible_user }}"
        group: "{{ ansible_user }}"
        mode: "0600"
        remote_src: true
```

But I cannot find an equivalent one in `pyinfra`.

I have to run a `shell` command:
 
```python
    server.shell(
        name=f"Kubeconfig - .kube/config ({host.data.ssh_user})",
        commands=[
            "cp /etc/rancher/k3s/k3s.yaml .kube/config",
            f"chown {host.data.ssh_user}:{host.data.ssh_user} .kube/config",
        ]
    )
```

## Describe the solution you'd like

Add `files.copy` to copy files on the target itself.

Thanks

Mikadore:
This is https://docs.pyinfra.com/en/3.x/operations/files.html#files-put

yellowhat:
Unfortunately `files.put` is about `Upload a local file, or file-like object, to the remote system.`.

I would like to copy a file from the remote system to the remote system itself.

geerlingguy:
Ditto, it would be convenient to have an idempotent file operation for `copy` instead of having to do a `shell` command each time :)

EricDriussi:
Would be nice if it could handle directories and permissions as well 🤔

yellowhat:
`files.copy` in 3.6 is almost there, as I can only copy a file to another directory but not without renaming.

For example:

```python
    files.copy(
        src="/etc/rancher/k3s/k3s.yaml",
        dest=".kube",
        overwrite=True,
    )
```

Will copy `k3s.yaml` to `.kube` but I would like to also change its name, like:

```python
    files.copy(
        src="/etc/rancher/k3s/k3s.yaml",
        dest=".kube/config",
        overwrite=True,
    )
```

errors:

```
--> Operation error in src/gevent/greenlet.py line 900: dest .kube/config is not an existing directory
```

EricDriussi:
TBH I agree it would be nice if it also handled renames (much like the `cp` command does).

[...]

PR #1446, opened 2025-09-06
Add files.copy operation

EricDriussi:
Adds copy operation suggested in #1121. Any and all feedback is welcome!

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -751,6 +751,15 @@ def _file_equal(local_path: str | IO[Any] | None, remote_path: str) -> bool:
     return False
 
 
+def _remote_file_equal(remote_path_a: str, remote_path_b: str) -> bool:
+    for fact in [Sha1File, Md5File, Sha256File]:
+        sum_a = host.get_fact(fact, path=remote_path_a)
+        sum_b = host.get_fact(fact, path=remote_path_b)
+        if sum_a and sum_b:
+            return sum_a == sum_b
+    return False
+
+
 @operation(
     # We don't (currently) cache the local state, so there's nothing we can
     # update to flag the local file as present.
@@ -1331,7 +1340,7 @@ def copy(src: str, dest: str, overwrite=False):
     dest_file_path = os.path.join(dest, os.path.basename(src))
     dest_file_exists = host.get_fact(File, dest_file_path)
     if dest_file_exists and not overwrite:
-        if _file_equal(src, dest_file_path):
+        if _remote_file_equal(src, dest_file_path):
             host.noop(f"{dest_file_path} already exists")
             return
         else:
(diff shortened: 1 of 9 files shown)


src/pyinfra/operations/files.py
```diff
@@ -1312,6 +1312,35 @@ def move(src: str, dest: str, overwrite=False):
...
+    if not host.get_fact(File, src) and not src_is_dir:
+        raise OperationError(f"src {src} does not exist")
+
+    if not host.get_fact(Directory, dest):
+        raise OperationError(f"dest {dest} is not an existing directory")
+
+    dest_file_path = os.path.join(dest, os.path.basename(src))
+    dest_file_exists = host.get_fact(File, dest_file_path)
+    if dest_file_exists and not overwrite:
+        raise OperationError(f"dest {dest_file_path} already exists and `overwrite` is unset")
```
DonDebonair: Why should this raise an `OperationError`. For `copy()` to be idempotent, the destination file existing, should just be a _noop_ right? Alternatively, you could compare hashes to see if the src file has changed.

EricDriussi: Indeed a noop would be a better fit here, although I still feel that an error should be raised in case `dest_file_path` already exists **and** is different from `src`. So maybe something like this?

```python
    if dest_file_exists and not overwrite:
        if _file_equal(src, dest_file_path):
            host.noop(f"{dest_file_path} already exists")
        else:
            raise OperationError(f"{dest_file_path} already exists and is different than src")
```

Or maybe there's a better way to handle that case?

Also, in the snippet I use the `_file_equal` [function](https://github.com/pyinfra-dev/pyinfra/blob/6bc88997705a34976d8b2c47f1e2d94e0f187c4c/pyinfra/operations/files.py#L742C1-L756C1), but that seems to be intended for local-remote hash comparisons.
Are you aware of any functions that might better fit this use case? Should I just create a new internal function based on this one for remote-remote hash comparison (what could be a nice name for it?)?

DonDebonair: I agree with you that the noop should only happen when the existing file at destination is the same.

I don't know if the `_file_equal` function can be used for remote-remote hash comparisons. @Fizzadar do you know?

Fizzadar: Indeed `_file_equal` is for local <> remote, a new function seems best here, basically fact x2 and compare.

EricDriussi:
Hey! Sorry it took me forever to get back to this...
I implemented the [changes mentioned here](https://github.com/pyinfra-dev/pyinfra/pull/1446#discussion_r2328793654), added/fixed tests and skipped the Windows CI (simply due to me not having a Windows machine to ensure the tests work before pushing).

Also, in the new [`_remote_file_equals` function](https://github.com/pyinfra-dev/pyinfra/pull/1446/files#diff-86e5946f52cfcf78fe5a111af41e331fe33ffc0aba21766b254b4eb1a8e1112aR754), I followed the same fallback logic I saw in `_file_equals` regarding the checksum comparison.
I'm not quite sure if this is correct so please let me know if there's anything that needs to be updated before merging.

Fizzadar (approved):
@EricDriussi no worries re:times, thank you for finishing this up! Looks good to me will leave for Daan to have a double check too.

PR #1447, opened 2025-09-06
operations.util.packaging: Extend PkgInfo for winget

rod7760:
## Background
I'm hoping to utilize the existing `ensure_packages` utility in `pyinfra-windows'. 

The utility currently makes some assumptions that don't work with `winget`. Namely, that all package managers support the following syntax:

```bash
apt install package1=version1 package2=version2

# does not work in winget
winget install package1=version1 package2=version2
```

You must do the following:
```bash
winget install package1 —version version1; winget install package2 —version version2;
```

This PR introduces `Pkginfo.inst_vers_template` to customize the creation `PkgInfo.inst_vers`.
The default template works from programs that support the  `{name}{operator}{version}` syntax. 

winget overides the template to do `'{name} {operator} {version};'.format("winget install Notepad++.Notepad++", "—version", "8.5")`

## Notes

This PR does change the `shlex.quote` behavior slightly, and I have left the failing tests for now.
Do you think this new quoting behavior is ok? 

```
# example failing test
--> COMMANDS OUTPUT:
[
    "pip install pyinfra==1.1 pytask another'>'1"
]
--> TEST WANTS:
[
    "pip install pyinfra==1.1 pytask 'another>1'"
]
```

If you don't want to support this here, I totally understand. I will duplicate some of the code. 

### DIFF
--- a/src/pyinfra/operations/util/packaging.py
+++ b/src/pyinfra/operations/util/packaging.py
@@ -15,12 +15,16 @@ from pyinfra.facts.rpm import RpmPackage
 from pyinfra.operations import files
 
 
+def default_inst_vers_format_fn(name: str, operator: str, version: str):
+    return "{name}{operator}{version}".format(name=name, operator=operator, version=version)
+
+
 class PkgInfo(NamedTuple):
     name: str
     version: str
     operator: str
     url: str
-    inst_vers_template: str = "{name}{operator}{version}"
+    inst_vers_format_fn: Callable = default_inst_vers_format_fn
     """
     The key packaging information needed: version, operator and url are optional.
     """
@@ -38,7 +42,7 @@ class PkgInfo(NamedTuple):
         """String that represents how a program can be installed.
 
         - If self.url exists, then url is always returned.
-        - If self.version exists, then inst_vers_template is used
+        - If self.version exists, then inst_vers_format_fn is used
         to create the string. The default template is '{name}{operator}{version}'.
         - Otherwise, self.name is returned.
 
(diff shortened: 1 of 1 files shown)


pyinfra/operations/util/packaging.py
```diff
@@ -34,13 +35,27 @@ def has_version(self) -> bool:
...
+
+        Note, the result string will be quoted, so input is shell safe.
+        """
+
+        if self.url:
+            return shlex.quote(self.url)
+
+        if self.version:
+            # assumes if version exists, then name and operator do
+            return self.inst_vers_template.format(
```
Fizzadar: Instead of a template, what about using a callable? Thinking:

```py
def inst_vers_format_fn(name, operator, version):
    return "{name}{operator}{version}".format(name=name, operator=operator, version=version)
```

That would (I think!) allow for the custom formatting while retaining the existing quoting.

Fizzadar (commented):
Big fan of using this in other packages, think extending here is a great call.

pyinfra/operations/util/packaging.py
rod7760: Ah, I didn't think about doing this. Thanks!

This should be ready now. The mypy error seems unrelated to my code. Most likely due to the new mypy release.

edit:
I guess I could have just done:

```
return shlex.quote("my_template".format(...)) 
```
I don't quite remember why I quoted the individual parts. 

Regardless, a Callable allows for more control and is probably a better interface.

Fizzadar (approved):
Great add, thank you @rod7760!

Issue #1435, opened 2025-08-16
files.directory doesn't ensure recursive ownership (add op arg)


missytake:
## Describe the bug
In https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra/operations/files.py#L1609 only ownership of the top-level directory is checked, not of the subdirectories and files, also if you specify `recursive=True`.

## To Reproduce
Steps to reproduce the behavior, please include where possible:

Create this directory structure:

```
$ ls -la test/
total 8
drwxrwxr-x  2 user     user 4096 Aug 16 09:38 .
drwxrwxr-x 15 user     user 4096 Aug 16 09:38 ..
-rw-rw-r--  1 www-data root    0 Aug 16 09:38 file
```

And apply the following operation to it:

```
files.directory(
  path="test",
  user="user",
  group="user",
  recursive=True,
)
```

## Expected behavior
If you call `files.directory` with `recursive=True` and the `user` or `group` argument, you expect that it ensures that the directory and its subdirectories and all files in them are ensured to be owned by the specified user/group. 

Recursive ownership could be checked with a simple `find $path ! -user $user -group $group`; for the largest directory we have (500k maildir directories), this check took ~13 minutes, compared to just running a chown, which took ~48 minutes.
This can be a performance nightmare of course; maybe instead of fixing this, the current behavior should simply be documented.
Maybe there are also faster options of checking recursive ownership I don't know about.

Thanks @cliffmccarthy for finding this: https://github.com/chatmail/relay/pull/609#discussion_r2263429306

cliffmccarthy:
Given the performance overhead of merely checking, it seems to me the current behaviour should be documented, explaining that `recursive` only applies to the action taken, and that only the state of the base directory is checked when deciding whether to take action.  I looked through the code history and it seems to have been this way for a long time, so changing existing functionality could be undesirable.

If we want to provide the ability to guarantee the final ownership of the entire tree without calling a potentially-slow `chown` every time, maybe another new argument could be added.  A `check_recursive` argument could enable something like the `find` command above and use that to decide whether to call the recursive `chown` or not.  This new argument would default to false, and could be documented as having potentially high overhead depending on the size of the tree.  That way this would only affect callers that explicitly ask for this.

[...]

PR #1448, opened 2025-09-12
docs: Specify how the recursive argument to directory() works

cliffmccarthy:
- As described in issue #1435, the 'recursive' argument to files.directory() does not ensure recursive ownership of the entire directory tree.  Only ownership of the top-level directory is checked, not that of the subdirectories and files.  Specifying 'recursive=True' will have no effect if the top-level directory already matches the mode, user, and group.
- This is how the operation has behaved for a long time, so changing existing functionality to check the entire tree could be undesirable.  It also takes more time to check the entire tree, so doing so affects performance.  A new option may be provided in the future for callers to explicitly check the whole tree when this is desired.
- Documented the existing behaviour of 'recursive' in the files.directory() docstring.

- [ ] Pull request is based on the default branch (`3.x` at this time)

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -1569,6 +1569,12 @@ def directory(
     + force_backup: set to ``False`` to remove any existing non-file when ``force=True``
     + force_backup_dir: directory to move any backup to when ``force=True``
 
+    ``recursive``:
+        Mode is only applied recursively if the base directory mode does not match
+        the specified value.  User and group are both applied recursively if the
+        base directory does not match either one; otherwise they are unchanged for
+        the whole tree.
+
     **Examples:**
 
     .. code:: python


cliffmccarthy:
The lint failures are in an unrelated file.  At the time of the 3.5 tag, which is the current top-of-branch, the workflow passed; the failures are the result of an unpinned mypy version.  At the time of 3.5, mypy was version 1.17.1; version 1.18.1 of mypy was released yesterday and introduces this failure.

Fizzadar (approved):
Thanks for documenting this, @cliffmccarthy!

Augh mypy, I’m going to pin it since I’m fed up of it breaking CI 🙃

cliffmccarthy:
You're welcome!  Once top-of-branch is updated and passing, I'll rebase this so we can get a clean run.

Fizzadar:
Rebased to fix mypy + update file location (after https://github.com/pyinfra-dev/pyinfra/commit/b14c5751350c1c24dafa592daffc64791d5f6de4 merged).

PR #1450, opened 2025-09-13
feat: use uv for project and dependency management

DonDebonair:
This commit replaces setuptools with uv, standardizing on pyproject.toml to define project metadata

Major changes include:

- All metadata has been moved from `setup.py` to `pyproject.toml`
- `pyinfra` and `pyinfra_cli` packages now reside inside a `src` folder. The big advantage of this is that this separates the code from tests and other files and prevents accidentally referencing files that are not packaged when building the project See: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
- All development commands for testing and linting are now run through `uv run`
- Dependency versions have been redefined using broader version ranges, but are kept to a specific version through a lock file. MyPy is the exception, which is violently pinned to `mypy==1.17.1`.
- Python 3.9 support has been dropped (it will reach EOL 2025-10)
- Python 3.13 support has been added
- Github workflows have been rewritten to use uv
- Package version is now generated from git, using uv-dynamic-versioning


maisim (approved):
I reread the whole thing and it seems very good to me, it's a nice change, thank you!

Fizzadar (approved):
Absolutely awesome, thank you @DonDebonair!

PR #1457, opened 2025-09-23
Support metadata for pyinfra plugins.

rod7760:
## Changes
* Add pyinfra-metadata-schema-1.0.0.json.
* Add pyinfra-metadata.toml with current plugins.
* Add Jinja2 templating to sphinx build.
* Extended build-public-docs.sh to build local docs.
* Added generated index for facts and operations.

## Bench Test
```
scripts/build-public-docs.sh 
```

## Screenshot of Change
image


docs/metadata.py
```diff
@@ -0,0 +1,80 @@
...
+    "service-management",
+    "package-manager",
+    "python",
+    "ruby",
+    "javascript",
+    "configuration-management",
+    "security",
+    "storage",
+    "system",
+    "system-ops",
```
DonDebonair: Why do we need `system-ops` and `system-facts` next to `system`? For the other tags, we don't have separate `-facts` and `-ops` variants. Whether we're dealing with facts or operations should be clear from the plugin type.

docs/metadata.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def parse_plugins(metadata_text: str) -> list[Plugin]:
+    """Given the contents of a pyinfra-metadata.toml parse out the plugins."""
+    pyinfra_metadata = tomllib.loads(metadata_text).get("pyinfra", None)
+    if not pyinfra_metadata:
+        raise ValueError("Missing [pyinfra.plugins] section in pyinfra-metadata.toml")
+
+    plugins = []
+    for p in pyinfra_metadata["plugins"]:
```
DonDebonair: I wonder if this can be even more concise if we'd use Pydantic for defining the `Tag` and `Plugin` models

DonDebonair:
Love this! I would move the `template` folder inside the `docs` folder if possible, because it's only relevant for docs right? And for `metadata.py` I wonder if the other way around makes sense: move it outside of the `docs` folder, because in the future, metadata could be relevant for other things than generating docs.

Maybe we want to create a plugin registry with API (think Ansible Galaxy)

rod7760:
Those both sound reasonable. That template folder is actually from an old implementation. I forgot to delete it. Whoops

pyinfra-metadata.toml
```diff
@@ -0,0 +1,550 @@
+"$schema" = "./pyinfra-metadata-schema-1.0.0.json"
```
Fizzadar, line 1: Might be nice to have a script to regenerate this file in-repo since there's so many entries (probably only relevant to pyinfra core itself).

Fizzadar: Eh, might need to put the tags into machine readable format in the docstrings to do that though 🤔

templates/operations.rst.j2
```diff
@@ -0,0 +1,54 @@
+Operations Index
```
Fizzadar: ~~Is this file or the existing/modified `operations.rst` being used to generate the docs? Both appear to have identical content...~~ nvm, dupe, to be deleted :)

Fizzadar (commented):
This is awesome stuff @rod7760! Couple of bits to cleanup Daan already pointed out and good to go I think.

Community context: this is the precursor work that will enable us to generate other pyinfra projects docs inline on docs.pyinfra.com avoiding a disjointed mess of documentation websites.

rod7760:
Resolved the open comments. I moved `docs.metadata` to `pyinfra.api.metadata`. 

Additionally, I now only render the `operations.rst` and `facts.rst` as `jinja2` templates. 
( `src/pyinfra/operations/files.py` contains `jinja2` templates as examples. They caused undefined variable errors.)

We could also explicitly exclude files, but the jinja2 templating felt niche enough to include files explicitly.

PR #1460, opened 2025-09-24
Pr1/3 : add GPG key management operations and facts

maisim:
Add new gpg.key and gpg.dearmor operations to manage GPG keys and keyrings. These operations provide a modern alternative to apt-key for managing APT repository keys.

Features:
- Install keys from URLs, local files, or keyservers
- Remove keys by ID or entire keyring files
- Convert ASCII armored keys to binary format
- Manage keys in specific keyrings or across all APT keyrings

This is part 1/3 of modernizing APT key management.

### DIFF
--- a/src/pyinfra/facts/gpg.py
+++ b/src/pyinfra/facts/gpg.py
@@ -148,3 +148,71 @@ class GpgSecretKeys(GpgFactBase):
         return ("gpg --list-secret-keys --with-colons --keyring {0} --no-default-keyring").format(
             keyring,
         )
+
+
+class GpgKeyrings(GpgFactBase):
+    """
+    Returns information on all GPG keyrings found in specified directories.
+
+    .. code:: python
+
+        {
+            "/etc/apt/keyrings/docker.gpg": {
+                "format": "gpg",
+                "keys": {...}  # Same format as GpgKeys fact
+            }
+        }
+    """
+
+    @override
+    def command(self, directories):
+        if isinstance(directories, str):
+            directories = [directories]
+
+        search_locations = " ".join(f'"{d}"' for d in directories)
+
+        # Generate a command that finds keyrings and lists their keys
+        # We'll use a shell script that outputs keyring path followed by key info
+        return (
+            f"for keyring in $(find {search_locations} -type f \\( -name '*.gpg' "
+            f"-o -name '*.asc' -o -name '*.kbx' \\) 2>/dev/null); do "
+            f'echo "KEYRING:$keyring"; '
+            f'if [[ "$keyring" == *.asc ]]; then '
+            f'gpg --with-colons "$keyring" 2>/dev/null || true; '
+            f"else "
+            f'gpg --list-keys --with-colons --keyring "$keyring" --no-default-keyring 2>/dev/null || true; '  # noqa: E501
+            f"fi; done"
+        )
+
+    @override
+    def process(self, output):
+        keyrings = {}
+        current_keyring = None
+        current_output: list[str] = []
+
+        for line in output:
+            line = line.strip()
+            if not line:
+                continue
+
+            if line.startswith("KEYRING:"):
+                # Process previous keyring if exists
+                if current_keyring and current_output:
+                    keyring_format = current_keyring.split(".")[-1].lower()
+                    keys = super().process(current_output)
+                    keyrings[current_keyring] = {"format": keyring_format, "keys": keys}
+
+                # Start new keyring
+                current_keyring = line[8:]  # Remove "KEYRING:" prefix
+                current_output = []
+            else:
+                # Accumulate GPG output for current keyring
+                current_output.append(line)
+
+        # Process final keyring
+        if current_keyring and current_output:
+            keyring_format = current_keyring.split(".")[-1].lower()
+            keys = super().process(current_output)
+            keyrings[current_keyring] = {"format": keyring_format, "keys": keys}
+
+        return keyrings
--- a/src/pyinfra/facts/gpg.py
+++ b/src/pyinfra/facts/gpg.py
@@ -171,17 +171,15 @@ class GpgKeyrings(GpgFactBase):
 
         search_locations = " ".join(f'"{d}"' for d in directories)
 
-        # Generate a command that finds keyrings and lists their keys
-        # We'll use a shell script that outputs keyring path followed by key info
+        # Two separate find+exec calls to avoid bash-specific syntax:
+        # - binary keyrings (.gpg, .kbx) use --keyring / --no-default-keyring
+        # - armored keyrings (.asc) are passed directly to gpg
         return (
-            f"for keyring in $(find {search_locations} -type f \\( -name '*.gpg' "
-            f"-o -name '*.asc' -o -name '*.kbx' \\) 2>/dev/null); do "
-            f'echo "KEYRING:$keyring"; '
-            f'if [[ "$keyring" == *.asc ]]; then '
-            f'gpg --with-colons "$keyring" 2>/dev/null || true; '
-            f"else "
-            f'gpg --list-keys --with-colons --keyring "$keyring" --no-default-keyring 2>/dev/null || true; '  # noqa: E501
-            f"fi; done"
+            f"find {search_locations} -type f \\( -name '*.gpg' -o -name '*.kbx' \\) 2>/dev/null"
+            f' -exec sh -c \'echo "KEYRING:$1";'
+            f' gpg --list-keys --with-colons --keyring "$1" --no-default-keyring 2>/dev/null || true\' _ {{}} \\; ;'
+            f" find {search_locations} -type f -name '*.asc' 2>/dev/null"
+            f' -exec sh -c \'echo "KEYRING:$1"; gpg --with-colons "$1" 2>/dev/null || true\' _ {{}} \\;'
         )
 
     @override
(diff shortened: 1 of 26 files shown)


maisim:
Hi @DonDebonair !
Can you take a look at this, please?
It seems to work, but I'm not sure I'm always using gpg correctly. Also, I'm always hesitant to use facts or shell commands within operations. Let me know what you think.
Thanks!

DonDebonair:
Thanks for the PR @maisim . I would not be afraid to use facts inside operations. I actually consider it a best practice to use facts to check things in operations. So instead of doing checks inside the shell commands you yield, there are probably places where you can rely on facts instead. That makes the yielded commands simpler.

If you can change those already, I'll do a more full review this weekend.

Caveat btw: I'm by no means a GPG expert. A short while ago, I wanted to use pyinfra to install Docker on a Debian host. I just created the keyring directory and downloaded the key directly into that, never touching GPG 😅

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,300 @@
...
+                "/etc/apt/trusted.gpg.d/*.gpg",
+                "/etc/apt/keyrings/*.gpg",
+                "/usr/share/keyrings/*.gpg",
+            ]
+
+            for pattern in keyring_patterns:
+                for kid in keyid:
+                    # Remove key from all matching keyrings
+                    yield (
+                        f'for keyring in {pattern}; do [ -e "$keyring" ] && '
```
DonDebonair: Doesn't it make more sense to retrieve keyrings through a fact and then only try to remove the key from keyrings that actually exist?

Or maybe even better, create a fact that lists all known keyrings and then use the `GpgKeys` fact for each of those keyrings to get all the keys. Then you can specifically remove the keyid if it exists

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,300 @@
...
+                        f'gpg --batch --no-default-keyring --keyring "$keyring" '
+                        f"--delete-keys {kid} 2>/dev/null || true; done"
+                    )
+
+                # Clean up empty keyrings
+                yield (
+                    f'for keyring in {pattern}; do [ -e "$keyring" ] && '
```
DonDebonair: As referenced in the other comment, if you can list all keys by keyring, then you know if keyrings are going to be empty after removing `keyid` from those rings and you can remove them based on that

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,300 @@
...
+                    f'--list-keys 2>/dev/null | grep -qi "{clean_key}"'
+                )
+
+            condition = " || ".join(key_checks)
+
+            # If any of the keys exist in the file, remove the entire file
+            yield (f'if [ -f "{dest}" ] && ({condition}); then ' f'rm -f "{dest}"; fi')
+            return
+
+        elif dest and not keyid:
```
DonDebonair, line 309: This could be `else` instead of `elif`. To me, that seems cleaner because it clearly signals there are only 3 possible combinations of conditions: `keyid` is provided and `dest` isn't, `keyid` and `dest` are both provided and lastly, `dest` is provided and `keyid` isn't

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,300 @@
...
+                    f'! gpg --batch --no-default-keyring --keyring "$keyring" '
+                    f'--list-keys 2>/dev/null | grep -q "pub" && rm -f "$keyring" || true; done'
+                )
+
+            return
+
+        elif dest and keyid:
+            # For APT keyring files, use a simpler approach:
+            # Check if keys exist in file, and if so, remove the entire file
+            # This is appropriate for most APT use cases
```
DonDebonair: I'm a bit concerned about "most APT use cases" here. Is there a situation we might be removed a keyfile/keyring that has multiple keys in it?

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,300 @@
...
+from urllib.parse import urlparse
+
+from pyinfra import host
+from pyinfra.api import OperationError, operation
+
+from . import files
+
+
+@operation()
+def key(
```
DonDebonair, line 231: This is a very complex function that covers many different scenarios. I think that from a public API standpoint, that is fine, because we're dealing with 1 resource: keys. But you could break up the function into separate functions that are all called from the `key()` function, based on the scenario.

See the `docker` operations for an example of how this can be done

DonDebonair:
Left some more comments @maisim incl. some ideas on how to leverate facts.

src/pyinfra/operations/gpg.py
maisim: Good idea. Since there’s no global or common location for keyrings, this will allow specifying the directories to search in — for example, /etc/apt/trusted.gpg.d/ and /usr/share/keyrings/ in the APT context.

src/pyinfra/operations/gpg.py
maisim: Yes, but it is more an "documentation" elif. To be more explicit.
What about add an else wich raise an exception if we go into an unhandle case?

DonDebonair: That works for me!

src/pyinfra/facts/gpg.py
```diff
@@ -148,3 +148,71 @@ def command(self, keyring=None):
...
+
+        search_locations = " ".join(f'"{d}"' for d in directories)
+
+        # Generate a command that finds keyrings and lists their keys
+        # We'll use a shell script that outputs keyring path followed by key info
+        return (
+            f"for keyring in $(find {search_locations} -type f \\( -name '*.gpg' "
+            f"-o -name '*.asc' -o -name '*.kbx' \\) 2>/dev/null); do "
+            f'echo "KEYRING:$keyring"; '
+            f'if [[ "$keyring" == *.asc ]]; then '
```
Fizzadar: This check feels risky as it's bash, not sh/posix, compatible. Two options I can think of:

- split the script to run find twice, you could also then just `find -exec` the gpg call after which would be cleaner, just do one for `*.gpg, *.kbx` and a separate for `.asc`
- use `case "$keyring" in *.asc)` or similar

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,414 @@
...
+
+def _install_key_from_keyserver(keyserver: str, keyid: str | list[str], dest: str, mode: str):
+    """Install GPG keys from a keyserver."""
+    if isinstance(keyid, str):
+        keyid = [keyid]
+
+    joined = " ".join(keyid)
+
+    # Create temporary GPG home directory
+    temp_dir = f"/tmp/pyinfra-gpg-{host.get_temp_filename('')[-8:]}"
```
Fizzadar: We should use `host.get_temp_filename` here.

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,414 @@
...
+            for existing_key_id in keys_in_keyring.keys():
+                if (
+                    clean_key == existing_key_id.upper()
+                    or existing_key_id.upper().endswith(clean_key)
+                    or existing_key_id.upper().startswith(clean_key)
+                ):
+                    keys_to_remove.append(existing_key_id)
+
+        if keys_to_remove:
+            # Remove the entire keyring file if any target keys are found
```
Fizzadar: This seems very wrong to me! Why would we want to nuke the entire file? Surly we'd only do that if _all_ keys match for deletion?

`gpg --batch --delete-keys` would probably be more appropriate in all cases here.

src/pyinfra/operations/gpg.py
```diff
@@ -0,0 +1,414 @@
...
+                    or existing_key_id.upper().endswith(clean_key)
+                    or existing_key_id.upper().startswith(clean_key)
+                ):
+                    keys_found = True
+                    break
+            if keys_found:
+                break
+
+        if keys_found:
+            # Remove the entire keyring file - safest approach for keyring management
```
Fizzadar: Same as above!

Fizzadar (changes requested):
Hi @maisim! Apologies for the big delay here, picking this up while Daan is afk, added a few things to fix.

src/pyinfra/operations/gpg.py
maisim: :+1: done in 66580a7b5c6f0cc209c5d14a75ef7498b2089880

src/pyinfra/operations/gpg.py
maisim: At the beginning, 23a22345, I used --delete-keys. In 04349890, I did a change to remove the whole file...
Cannot remember why I did that. Maybe beacause we cannot remove key from armored keyrings... 
Now, on only a key is removed, if there no more key in the keyring, the file is deleted.
Fom armored keyrings, if the key have to be removed but the keyring contains another one, it raise an OperationError

Let me know

src/pyinfra/facts/gpg.py
maisim: You're right
done in a9f02f6b7cfa196b94bfefdf04bbcace48f82773
I used the find exec solution.

src/pyinfra/operations/gpg.py
Fizzadar: I think this is perfect, would much rather error in the armored case than remove the whole file 👍

Fizzadar (approved):
Very nice, @maisim!

PR #1462, opened 2025-09-25
Fix TmpDir to match POSIX

### LINKED ISSUE #1458: server.TmpDir fact is buggy and breaks things
## Describe the bug

`host.get_fact(TmpDir)` simply returns empty on most environments. `server.script` operation fails due to this.

## To Reproduce

+ Operation code & usage

   ```py
   # testtmpdir.py
   from pyinfra.context import host
   from pyinfra.facts.server import TmpDir

   print(host.get_fact(TmpDir))
   ```

   Run with `pyinfra @local testtmpdir.py`

+ Target system information

Tested on two computers

   1. `Linux xxx 6.16.8-arch2-1 #1 SMP PREEMPT_DYNAMIC ... x86_64 GNU/Linux`
   2. `Linux xxx 6.8.0-79-generic #79~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC ... x86_64 x86_64 x86_64 GNU/Linux`

   ```sh
   pyinfra @docker/ubuntu:22.04 testtmpdir.py
   ```

## Expected behavior

I expect it to return `/tmp` from the fact, from what I understand from the [docs](https://docs.pyinfra.com/en/3.x/facts/server.html#server-tmpdir)

## Meta

N/A

The bug is simply due to the `TEMPDIR` env variable not existing on most systems it seems.

https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/facts/server.py#L60

wowi42:
Fix #1458 

 According to POSIX standards, checks environment variables in this order:
    1. TMPDIR (if set and accessible)
    2. TMP (if set and accessible)
    3. TEMP (if set and accessible)
    4. Falls back to /tmp


src/pyinfra/facts/server.py
```diff
@@ -52,12 +52,28 @@ def command(self):
...
-        return "echo $TMPDIR"
+        return """
+if [ -n "$TMPDIR" ] && [ -d "$TMPDIR" ] && [ -w "$TMPDIR" ]; then
+    echo "$TMPDIR"
+elif [ -n "$TMP" ] && [ -d "$TMP" ] && [ -w "$TMP" ]; then
+    echo "$TMP"
+elif [ -n "$TEMP" ] && [ -d "$TEMP" ] && [ -w "$TEMP" ]; then
+    echo "$TEMP"
+else
+    echo "/tmp"
```
Fizzadar: We shouldn't provide a default here, the fallback is specified at the config level currently (https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/api/config.py#L29). 

To be honest I think this way is nicer, but we can't break the config for backwards compatibility so if no matching env vars the fact still needs to return a blank string.

Fizzadar (commented):
Way nicer, slight change needed for backwards compatibiity (sorry) 🙏

src/pyinfra/facts/server.py
teshanshanuka: None of the `TMPDIR|TMP|TEMP` are set in my arch/zsh terminal or in another ubuntu/bash PC I checked.  

I had the [this issue](https://github.com/pyinfra-dev/pyinfra/issues/1458) and the initial problem was `server.script` not working. The underlying issue was that this fact returning empty. I remember seeing something about setting a default in the pyinfra code, but did not check in details. Still the `script` operation was failing (`v3.5`). 

So if this returns a blank, my guess is it still would fail here in my machine, unless the `script` was fixed in the meantime to use the proper default (if that is how it should work).

wowi42: pushed the change, let me know @Fizzadar

Fizzadar (approved):
Awesome stuff, thank you @wowi42!

Fizzadar:
Funny note - was just in go docs and spotted Go doesn't respect POSIX either: https://pkg.go.dev/os#TempDir

PR #1465, opened 2025-09-25
Pr 2/3: facts.apt.AptSources: add deb822 format support

maisim:
facts: add deb822 format support for APT sources
    
    Add support for parsing deb822 format (.sources) files alongside
    traditional .list format APT source files.
    
    Features:
    - Parse deb822 stanzas with Types/URIs/Suites/Components fields
    - Map deb822 specific fields (Architectures, Signed-By) to legacy options
    - Maintain backwards compatibility with existing parse_apt_repo output
    - Support multiple types/URIs/suites combinations per stanza
    
    This is part 2/3 of modernizing APT source management.

Follow #1460 

### DIFF
--- a/src/pyinfra/facts/apt.py
+++ b/src/pyinfra/facts/apt.py
@@ -80,13 +80,7 @@ class AptRepo:
 
     def to_json(self):
         """Convert to dict for JSON serialization"""
-        return {
-            "type": self.type,
-            "url": self.url,
-            "distribution": self.distribution,
-            "components": self.components,
-            "options": self.options,
-        }
+        return dict(self.items())
 
 
 @dataclass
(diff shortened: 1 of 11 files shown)


maisim:
Hi @DonDebonair,
Hope you’re doing well.
I can’t remember why I left it as WIP, so I’m changing it now. I’ll do some functional tests today.

maisim:
~Scoped diff
https://github.com/maisim/pyinfra/compare/pr1-add-gpg-operations...maisim:pyinfra:pr2-add-deb822-support~

Branch rebased on 3.x after #1460  was merged

src/pyinfra/facts/apt.py
```diff
@@ -1,13 +1,230 @@
...
+    full type safety for modern code.
+    """
+
+    type: str  # "deb" or "deb-src"
+    url: str  # Repository URL
+    distribution: str  # Suite/distribution name
+    components: list[str]  # List of components (e.g., ["main", "contrib"])
+    options: dict[str, Union[str, list[str]]]  # Repository options
+
+    # Dict-like interface for backward compatibility
```
Fizzadar, line 31: Nothing to do now but wonder if we should extract this to re-use the pattern elsewhere, the type safety improvement is fantastic.

src/pyinfra/facts/apt.py
```diff
@@ -1,13 +1,230 @@
...
 
 from typing_extensions import TypedDict, override
 
 from pyinfra.api import FactBase
 
 from .gpg import GpgFactBase
-from .util import make_cat_files_command
+
+
+@dataclass
```
Fizzadar: We should set frozen here as is read only

```suggestion
@dataclass(frozen=True)
```

src/pyinfra/facts/apt.py
```diff
@@ -1,13 +1,230 @@
...
+                and self.options == other.options
+            )
+        return False
+
+    def to_json(self):
+        """Convert to dict for JSON serialization"""
+        return dict(self.items())
+
+
+@dataclass
```
Fizzadar:
```suggestion
@dataclass(frozen=True)
```

Fizzadar (approved):
Very nice, @maisim! I'll add frozen flags and merge.

src/pyinfra/facts/apt.py
```diff
@@ -11,7 +11,7 @@
 from .gpg import GpgFactBase
 
 
-@dataclass
+@dataclass(frozen=True)
```
maisim, line 14: Hi @Fizzadar 
Add frozen=True make sense but conflict with __setitem__ that provide backward comatible dict-like write  (e.g. repo['url'] = ...). 

```python
for repo in host.get_fact(AptSources):
    repo['url'] = repo['url'].replace('http://', 'https://')  # raise FrozenInstanceError
```
Two options:

- Keep frozen=True and drop __setitem__ — clean, but a breaking change that should be documented
- Keep @dataclass without frozen — preserves backward compatibility at the cost of mutability guarantees

What do you think?

Fizzadar: Whoops 🤦 I think second, better than breaking functionality. We can add a note to switch it in v4.

PR #1466, opened 2025-09-26
Added a line to clear the memoization cache before setting up the mock

wowi42:
The issue was that the test_get_sftp_fail test was failing because of a memoization cache problem. The get_file_transfer_connection   method in the SSH connector is decorated with @memoize, which caches the result of the method call. When running multiple tests, if   the method was successfully called in a previous test, it would return the cached result instead of executing again and potentially raising the expected SSHException


Fizzadar (approved):
🎉 🎉 🎉 thank you for tracking this down @wowi42! Been a total pain to get rid of all the CI wonkiness caused by this, glad you found what is hopefully the last case!

PR #1492, opened 2025-10-25
feat: make docker.image idempotent

DonDebonair:
`docker.image(...)` will now check if the image exist and only pull when required. It uses the following logic for this:

1. If `force == True` it will always pull the image
2. If an exact digest is provided, it will pull if there is no image with the same registry + repo + name + digest
3. If a tag is provided and the tag is `latest`, it will always pull
4. If a tag is provided and it is not `latest`, it will pull if no image exists with the same registry + repo + name + tag

It offloads most of this logic to Docker itself, by relying on the `DockerImage` fact, which in turn uses `docker image inspect` to figure out if an image with name + tag or digest exists.


src/pyinfra/operations/docker.py
```diff
@@ -127,13 +133,14 @@ def container(
         )
 
 
-@operation(is_idempotent=False)
```
Fizzadar, line 130: 🎉

Fizzadar (approved):
Very nice, @DonDebonair, thank you for the PR!

PR #1504, opened 2025-11-29
facts.apt.update: fix computation of cache_time when host is not UTC

philippemilink:
And reorganize a bit the logic to gather the Date fact only if required.

The timezone of my hosts are set on UTC+1. I noticed that the operation `apt.update` was always executed, even when the `cache_time` parameter was supposed to make it a noop. See the comment in the code for an explanation of the problem.

- [x] Pull request includes documentation for any new/updated operations/facts => not applicable
- [x] Tests pass (see `scripts/dev-test.sh`) => let's see the result of the CI

### DIFF
--- a/src/pyinfra/operations/apt.py
+++ b/src/pyinfra/operations/apt.py
@@ -4,7 +4,7 @@ Manage apt packages and repositories.
 
 from __future__ import annotations
 
-from datetime import timedelta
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Thank you for the PR, @philippemilink! Nice fix and a test too 👌

PR #1505, opened 2025-11-29
cli: add option `--same-sudo-password`

philippemilink:
To tell all hosts in the inventory uses the same password, which will be prompted only once.

In my setup, all SSH hosts of my inventory have the same sudo password. The problem is that when I run an operation which requires sudo on all my inventory, I am prompted to give the sudo password for each host, and all the prompts are somehow mixed on the same line:
pyinfra-sudo-1
pyinfra-sudo-2

I don't want to put the sudo password as a parameter of `pyinfra` command (to avoid leaking it in my shell history), but rather I want to be prompted, but only once.

So this PR adds the option `--same-sudo-password` to prompt for sudo password only once at the beginning of the execution.

- [ ] Pull request includes tests for any new/updated operations/facts => any hint (where to look for inspiration) to add the corresponding test would be welcome 
- [ ] Pull request includes documentation for any new/updated operations/facts => where should I mention this option in the documentation?

### DIFF
--- a/src/pyinfra_cli/cli.py
+++ b/src/pyinfra_cli/cli.py
@@ -84,12 +84,6 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
     show_envvar=True,
 )
 @click.option(
-    "--same-sudo-password",
-    is_flag=True,
-    default=False,
-    help="All hosts have the same sudo password, so ask only once.",
-)
-@click.option(
     "--limit",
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Hi @philippemilink - this looks great! Actually sent me down a rabbit hole... turns out `--use-sudo-password` is completely unused, literally does nothing. Will fix that up too. I pushed a tiny change to put the handling under `_set_config`.

PR #1506, opened 2025-12-02
connectors: close stdin after writing any input

Fizzadar:
Turns out some applications will wait on stdin being closed before going anything (ollama), see: https://github.com/pyinfra-dev/pyinfra/issues/1355


Fizzadar:
I believe this to be safe because this is what happens when you pipe things `echo hi | sleep 500` leaves sleep process w/o any echo.

geerlingguy:
It seems like it'd be safe. I'm sure there's some random edge case that could be broken the other way around... but I can't think of one that I've ever done!

PR #1508, opened 2025-12-07
add import statements on at least one example of every operation; cha…

mkinney:
…nge the docs for Centos; update linux versions; remove Choco facts/operations as it is Windows

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -117,6 +117,7 @@ def download(
 
     .. code:: python
 
+        from pyinfra.operations import files
         files.download(
             name="Download the Docker repo file",
             src="https://download.docker.com/linux/centos/docker-ce.repo",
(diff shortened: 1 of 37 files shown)


src/pyinfra/operations/files.py
```diff
@@ -1789,7 +1790,7 @@ def block(
         # put complex alias into .zshrc
         files.block(
             path="/home/user/.zshrc",
-            content="eval $(thefuck -a)",
```
Fizzadar, line 1814: 🤣

Fizzadar (commented):
Nice, much improved @mkinney!

I think we need to skip/undo the choco removal. It's Windows only but accidentally leftover and removing it would be a breaking change, the ops/facts have not been migrated to pyinfra-windows: https://github.com/pyinfra-dev/pyinfra-windows/tree/main/src/pyinfra_windows/operations/util. I'll do the revert shortly so can get this merged 👍 

cc @rod7760 as pyinfra-windows owner

PR #1512, opened 2025-12-21
Fixed sources urls in Docs "Using the API" section

kelno:
Just fixing 2 urls in doc, those files were moved to a subdir.


Fizzadar (approved):
Thank you for fixing these, @kelno!

PR #1516, opened 2025-12-31
docs: minor errors

romain-dartigues:
I did fix a few errors in the documentation I came through today.

Quick questions:
1. There are a lot of Markdown pages (connectors, support…); would you accept if I converted them in ReST to get the benefits of the Sphinx domains (ie: ref, samp…)?
2. There are a lot of "generated ReST files", (apidocs/, connectors/, facts/, operations/, etc.); why not use [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) and remove a lot of complexity (and avoid errors such as `pyinfra/docs/operations/opkg.rst:22: WARNING: Lexing literal_block 'opkg.packages(\n    packages: str | typing.List[str]="", present: <class \'bool\'>=True,\n    latest: <class \'bool\'>=False, update: <class \'
bool\'>=True, **kwargs,\n)' as "python" resulted in an error at token: "'". Retrying in relaxed mode. [misc.highlighting_failure]`)

### DIFF
--- a/docs/api/connectors.md
+++ b/docs/api/connectors.md
@@ -18,6 +18,8 @@ In the example below `make_names_data` is yielding a hostname, data, groups tupl
 could be stored and processed in any data structure.
 
 ```py
+from pyinfra.connectors.base import BaseConnector
+
 class InventoryConnector(BaseConnector):
     handles_execution = False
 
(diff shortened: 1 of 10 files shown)


Fizzadar:
Thanks for fixing this up, @romain-dartigues!

> Quick questions:
> 
>     1. There are a lot of Markdown pages (connectors, support…); would you accept if I converted them in ReST to get the benefits of the Sphinx domains (ie: ref, samp…)?

Let's hold off on this for now, I'm weighing up moving away from sphinx to mkdocs/similar for simplicity so markdown is probably the future.

>     2. There are a lot of "generated ReST files", (apidocs/, connectors/, facts/, operations/, etc.); why not use [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) and remove a lot of complexity (and avoid errors such as `pyinfra/docs/operations/opkg.rst:22: WARNING: Lexing literal_block 'opkg.packages(\n    packages: str | typing.List[str]="", present: <class \'bool\'>=True,\n    latest: <class \'bool\'>=False, update: <class \' bool\'>=True, **kwargs,\n)' as "python" resulted in an error at token: "'". Retrying in relaxed mode. [misc.highlighting_failure]`)

Autodocs output wasn't right for the pyinfra documentation (this is many years ago though, so I've long forgotten why :)), I know it can be tweaked but it'd be a similar effort to the current custom setup which works fine. I think unless there's a particular benefit to doing so the current system does the job well enough.

Fizzadar (approved):

PR #1518, opened 2026-01-01
docs: replace `python -m` with `uv run -m`

JP-Ellis:
All the instructions recommend that `uv run` be used (with good reason), save for one. So this makes a very minor update to the docs to suggest using `uv run -m ...` for consistency.

This is unlikely to be an issue as the `http.server` module is rather stable, but it won't harm to make sure that the same Python version and venv is used.

- [ ] ~Pull request includes tests for any new/updated operations/facts~
- [ ] ~Pull request includes documentation for any new/updated operations/facts~
- [ ] Tests pass (see `scripts/dev-test.sh`) (2026 bug)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -35,7 +35,7 @@ uv sync
 
 ### Code Style & Type Checking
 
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Good spot! Thank you @JP-Ellis!

PR #1528, opened 2026-01-17
Remove invalid shell from choco

kayoub5:
There is no such thing as a "ps" shell, and even if there one, a fact is not supposed to force it.


Fizzadar:
@kayoub5 this is a leftover from the [`pyinfra-windows`](https://github.com/pyinfra-dev/pyinfra-windows) codebase. The `ps` executable was special cased for that, so needs to remain here for compatibility. Eventually this module will be removed entirely from pyinfra core as it was meant to be.

PR #1540, opened 2026-02-01
Fix CLI flags lost on multiple deploy files

### LINKED ISSUE #1527: CLI: Executing multiple deploy files will only apply cli flags to the first one
## Describe the bug
When using the cli to execute multiple deploy files, such as:  
`pyinfra inventory.py foo.py bar.py --diff`  
The flag(s) will only apply to the first deploy file.  

## Minimal setup to Reproduce

**a.py**
```

files.put(
    src="a.txt",
    dest="/etc/motd",
)
```

**b.py**
```

files.put(
    src="b.txt",
    dest="/etc/motd",
)
```

- a.txt and b.txt can be any text file.
- (I tested this using the SSH connector defined in inventory.py)
- Run `pyinfra inventory.py a.py b.py --diff`.
-> **Only changes from a.py will show.**

I've put the complete example in a repo [here](https://github.com/kelno/pyinfra-cli-bug).

## Code investigation 
When using the CLI one of the first thing happening is that the config is loaded using `_set_config`, which starts by calling `lock_current_state`,  just after config file is loaded but AFTER cli args are loaded.
https://github.com/pyinfra-dev/pyinfra/blob/5291e90eb3700e5b141db4aba82b7d85c248717d/src/pyinfra_cli/cli.py#L600
So at this point, the config from args is not loaded yet so we're basically saving an empty config, to restore it later.  
When the first deploy files has finished running, the ["locked state" is restored](https://github.com/pyinfra-dev/pyinfra/blob/5291e90eb3700e5b141db4aba82b7d85c248717d/src/pyinfra_cli/cli.py#L782), erasing the config object as was loaded by the CLI. 

## Possible fix
Unless I'm missing some deliberate intent, this issue could be fixed by simply moving [`config.lock_current_state()`](https://github.com/pyinfra-dev/pyinfra/blob/5291e90eb3700e5b141db4aba82b7d85c248717d/src/pyinfra_cli/cli.py#L600) to the end of the function, after CLI args are processed.  

## Expected behavior
Config from CLI is kept for all deploy files. 

## Meta
- `pyinfra --support`.
System: Windows
      Platform: Windows-11-10.0.26100-SP0
      Release: 11
      Machine: AMD64
    pyinfra: v3.6
      click: v8.3.1
      distro: v1.9.0
      gevent: v25.9.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      pydantic: v2.12.5
      python-dateutil: v2.9.0.post0
      typeguard: v4.4.4
      typing-extensions: v4.15.0
    Executable: [...]\pyinfra_cli_bug\.venv\Scripts\pyinfra
    Python: 3.12.11 (CPython, MSC v.1944 64 bit (AMD64))

wowi42:
Move config.lock_current_state() to after CLI argument processing so that flags like --diff, --sudo, --parallel, etc. are included in the locked config state. Previously, the lock was created before CLI args were applied, causing them to be lost when the config was restored between deploy file executions.

Fixes #1527 

Summary of the fix:
- src/pyinfra_cli/cli.py: Moved config.lock_current_state() from line 600 (before CLI args) to line 634 (after all CLI args are applied)

#FOSDEM

### DIFF
--- a/src/pyinfra_cli/cli.py
+++ b/src/pyinfra_cli/cli.py
@@ -595,10 +595,6 @@ def _set_config(
     if path.exists(config_filename):
         exec_file(config_filename)
 
-    # Lock the current config, this allows us to restore this version after
-    # executing deploy files that may alter them.
-    config.lock_current_state()
-
     # Arg based config overrides
     if sudo:
         config.SUDO = True
@@ -631,6 +627,11 @@ def _set_config(
 
     if retry_delay is not None:
         config.RETRY_DELAY = retry_delay
+
+    # Lock the current config, this allows us to restore this version after
+    # executing deploy files that may alter them. This must happen after CLI
+    # args are applied so they persist across multiple deploy files.
+    config.lock_current_state()
 
     return config
 


Fizzadar (approved):
Nice catch, @wowi42!

PR #1543, opened 2026-02-01
cli/inventory: support loading host groups directly from module attributes

### LINKED ISSUE #1513: Support loading inventory host group direct from CLI
> This was added in ed8196de7387dd1e5a8d3ab85edc7090bb31158f; there are no docs (fair enough, it's experimental) but I feel like this simplifying use-case of `inventory_module.host_group` would be beneficial. Consider the following `inventory.py` contents:
> 
> ```python
> 
> proxmox_hosts = ["pve01", "pve02", "pve03"]
> 
> access_points = ["unifi01", "unifi02", "unifi03"]
> 
> ```
> 
> 
> 
> It would be nice to use `pyinfra inventory.proxmox_hosts deploy.py` instead of the more verbose and potentially more error prone `pyinfra inventory.py deploy.py --limit proxmox_hosts`.
> 
> 
> 
> Currently the code assumes that this argv syntax _always_ refers to a callable, I think checking if the user parameter is an iterable would be a reasonably simple addition and would enhance UX.
> 
> 
> 
> https://github.com/pyinfra-dev/pyinfra/blob/511e7c3a26af1d83bc4e6ef5113c1d06827dd0cd/pyinfra_cli/inventory.py#L134-L143 

 _Originally posted by @mechanarchy in [#959](https://github.com/pyinfra-dev/pyinfra/issues/959#issuecomment-2257359933)_

wowi42:
Allow `pyinfra inventory.proxmox_hosts deploy.py` syntax where `proxmox_hosts` is a list in inventory.py, not just a callable function.

Previously, the module.attribute syntax only worked with callables that return a dictionary of groups. Now it also supports direct list/tuple attributes containing hosts.

Fixes #1513 

### DIFF
--- a/src/pyinfra_cli/inventory.py
+++ b/src/pyinfra_cli/inventory.py
@@ -177,8 +177,11 @@ def make_inventory(
             "nor refers to a python module"
         )
         return Inventory.empty()
-    else:
+    elif callable(inventory_func):
         return make_inventory_from_func(inventory_func, override_data)
+    else:
+        # The inventory is an iterable (list/tuple) of hosts from a module attribute
+        return make_inventory_from_iterable(inventory_func, override_data)
 
 
 def make_inventory_from_func(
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
Awesome, thank you @wowi42!

PR #1553, opened 2026-02-08
facts.hardware.Memory: use LANG=C while calling vmstat

soltysek:
Looks like reading available memory fails when system locale is not set to English (or similar). Forcing `LANG=C` for vmstat call to avoid this.


Fizzadar (approved):
Good catch, thanks for the PR, @soltysek!

PR #1556, opened 2026-02-11
fix/sudo-single-try

harshang03:
Enforce a single sudo password attempt per command by adding a one-shot askpass sentinel, clearing cached sudo passwords on auth failure, and updating SSH/local connector logging + tests.

ISSUE: https://github.com/pyinfra-dev/pyinfra/issues/1043  

### DIFF
--- a/src/pyinfra/connectors/local.py
+++ b/src/pyinfra/connectors/local.py
@@ -16,6 +16,7 @@ from .util import (
     CommandOutput,
     execute_command_with_sudo_retry,
     make_unix_command_for_host,
+    output_indicates_sudo_password_failure,
     run_local_process,
 )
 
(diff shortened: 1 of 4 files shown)


Fizzadar:
Sorry for the delay on this one @harshang03! The askpass code has moved around a fair bit since Feb - `SUDO_ASKPASS_COMMAND` is now `ASKPASS_COMMAND` and shared by both the sudo and su paths - so this'll need a rebase before I can properly review it. Couple of things to sort out when you do:

- The 3-arg template breaks `su` - `_ensure_askpass_set_for_host` only passes 2 args, so it'll `IndexError` at format time. The sentinel doesn't really mean anything for su either, there's no retry loop there.
- We build a fresh sentinel path per sudo command, so a long deploy leaves one `pyinfra-sudo-askpass-once-*` file per command sat in the temp dir, and leaks the lot if the run gets interrupted. Should be one per host I think?
- `output_indicates_sudo_password_failure` duplicates `SUDO_PASSWORD_REQUIRED_LINES`, which already drives the retry in `execute_command_with_sudo_retry` and covers sudo-rs as well. Let's just reuse that rather than carry two matchers.

I'm going to close this as stale but would be happy to help with review or rebase to land this.

PR #1558, opened 2026-02-18
Fix systemd-state-detection

harshang03:
Short description:

Make systemd.service detect failures immediately by appending systemctl is-active --quiet after start/restart/reload, and improve SystemdStatus parsing with ActiveState/failed state handling.

#issue : https://github.com/pyinfra-dev/pyinfra/issues/1154 

### DIFF
--- a/src/pyinfra/facts/systemd.py
+++ b/src/pyinfra/facts/systemd.py
@@ -41,6 +41,11 @@ def _make_systemctl_cmd(user_mode=False, machine=None, user_name=None):
     return StringCommand(*systemctl_cmd)
 
 
+SYSTEMD_ACTIVE_STATE_KEY = "ActiveState"
+SYSTEMD_ACTIVE_STATES = ("active", "activating", "reloading")
+SYSTEMD_FAILED_STATES = ("failed",)
+
+
 class SystemdStatus(FactBase[Dict[str, bool]]):
     """
     Returns a dictionary map of systemd units to booleans indicating whether they are active.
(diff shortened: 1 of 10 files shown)


Fizzadar (changes requested):
Thanks for digging into this @harshang03, and sorry for the slow reply! The fact side is spot on - `ActiveState=active` with a `SubState` we don't happen to list (`.target` units, `start-post`) reporting not-running is a real bug. Couple of problems with the operation half though:

- `is-active --quiet` will false-fail for `Type=oneshot` units w/o `RemainAfterExit=yes` - those go `inactive` immediately after a successful start, so the `&&` chain fails and we mark the whole operation failed. That's a regression for existing deploys, not just a gap in the new check.
- I'd rather not plumb `status_command` through `handle_service_control`, it's shared by 7 backends and this only ever serves systemd. Can we just yield a second `StringCommand(systemctl_cmd, "is-active", "--quiet", QuoteString(service))` from `systemd.service` and leave `util/service.py` alone? Also dodges the second half of the `&&` running unprivileged under `_sudo` when `_shell_executable` is unset.
- The diff reverts `util/service.py` back to `shlex.quote` string building - that got migrated to `QuoteString` in #1617, so this needs a rebase. Fact fixtures need updating too, `command()` emits `--property ActiveState` now but all 6 still assert the old string.

One I'm not sure on: `ActiveState=activating` → `True` means a crash-looping unit in auto-restart counts as running, so `running=True` will never try to start it. I think we want `False` there? wdyt

PR #1574, opened 2026-03-02
connectors.ssh: fix parsing of SSH config file comments

### LINKED ISSUE #1568: pyinfra not robust for comments at the end of lines in ssh config
## Describe the bug
Pyinfra is not processing comments in the ./ssh/config properly 

## To Reproduce
Put comments behind a line like "    IdentityFile ~/.ssh/pyinfra_user_key   # example comment"

The result of running pyinfra: 
--> Connecting to hosts...
    [127.0.0.1] Could not connect ([Errno 2] No such file or directory: '/home/martijn/.ssh/pyinfra_user_key                # your private key')

## Expected behavior
If I can ssh with a username like pyinfra (in my case via an open ssh CA signed key) then I expect that pyinfra will also work ... but that is not the case (not only comments are an issue, I've seen multiple things going working in the past days working with pyinfra)

## Meta
Well this is simple enough to reproduce I guess. 
I've fixed it by removing the comments. So pyinfra is usuable ... however it could be made more robust

wowi42:
SSH config lines with inline comments (e.g. "IdentityFile ~/.ssh/key # comment") were passed through to paramiko with the comment included, causing issues like "No such file or directory: '/home/user/.ssh/key # comment'".

Fixes pyinfra-dev/pyinfra#1568

### DIFF
--- a/src/pyinfra/connectors/sshuserclient/config.py
+++ b/src/pyinfra/connectors/sshuserclient/config.py
@@ -48,7 +48,7 @@ paramiko.config.invoke = FakeInvoke  # type: ignore
 
 
 def _strip_inline_comment(line):
-    """Strip inline comments from SSH config lines, respecting quoted strings."""
+    """Strip inline comments from SSH config lines, respecting quoted strings"""
     in_quote = False
     quote_char = None
     for i, char in enumerate(line):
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice, thank you @wowi42!

PR #1576, opened 2026-03-02
operations: expand quoting of user inputs to prevent command injection

### LINKED ISSUE #1549: Mitigating command injection - security issue
Hey,

we've been looking at possibly adopting PyInfra for infrastructure management and one aspect is of this security, especially for a tool that will be running numerous shell commands on production servers. I couldn't find anything on how security is handled in this project.

Given the nature of the tool, a main concern is shell/command injection. Looking around the code I see there are a lot of cases of commands being created by simply formatting user input into them in operations and this also not being handled at a lower layer (when the command is actually executed.

Experimenting also seems to indicate that this is not mitigated properly. The snippet below demonstrates the problem. It is a bit of a contrived example of course but you can very much imagine getting the value of an operation argument from a config file or other source. So an attacker could then add a command-injected value in said configuration elsewhere which would cause the command to be executed on the host, without it being obvious from the users pyinfra code that this might happen.

Because this is not just an issue for the `server.shell` operation (where the user might expect these risks to be present and mitigate it themselves) but also for other non-command arguments of operations, there is a high chance of the user not being aware of these risks.

```python
# inventory.py
my_hosts = [
    ("@local", {"requirements_txt_path": "requirements.txt; rm -rf /"}),
]

# deploy.py
from pyinfra.operations import pip
from pyinfra import host

pip.packages(
    name="Install packages",
    virtualenv=".venv",
    requirements=host.data.get("requirements_txt_path"),
)

# running `pyinfra inventory.py deploy.py -y` would now try to delete the root folder on localhost.
```

So I'm wondering is there any global system or development standard that should prevent this, unittests that should catch these kinds of issues or point on a roadmap to address these issues? Or am I maybe just missing something in my research here (very possible 😄).

wowi42:
Add shlex.quote() to user-supplied values (paths, hostnames, usernames, branch names, URLs, passwords, etc.) in pip, npm, ssh, git, and server operations. This prevents command injection when values come from untrusted sources like inventory data or config files.

Phase 1 covers the 5 most critical operation files. The codebase already uses safe patterns (StringCommand + QuoteString) in files.py and FreeBSD operations; this extends the same principle using shlex.quote() for operations that yield plain strings.

Fix #1549 


wowi42:
I would love a second opinion @Fizzadar @mkirkeng-leukeleu

src/pyinfra/operations/ssh.py
```diff
@@ -48,17 +48,19 @@ def keyscan(hostname: str, force=False, port=22):
         pattern=hostname,
     )
 
+    homedir = str(homedir)
+
     keyscan_command = "ssh-keyscan -p {0} {1} >> {2}/.ssh/known_hosts".format(
         port,
```
mkirkeng-leukeleu, line 54: Why not quote the port? That's also user input right, so I think that could also be used for command injection.

mkirkeng-leukeleu:
Looks like a good improvement 👍 :)

There's probably a good reason not to require a higher python version. But if PyInfra ever upgrade to minimum 3.14, this might be a good use case for t-strings. Then you could handle all the shlexing at the point of running the command rather than in each operation separately.

Fizzadar (approved):
Nice, thanks for fixing these up @wowi42!

Long run we should migrate everything over to `StringCommand` I think. This is a great safety improvement as-is still 👍

PR #1578, opened 2026-03-05
facts.server.Sysctl: fix non-zero exit code causing empty result

wowi42:
`sysctl -a` exits with code 1 on Linux when some keys are unreadable (e.g. net.ipv6.conf.*.stable_secret I/O errors), causing pyinfra to skip processing stdout and return an empty dict. Suppress stderr and force a zero exit code so valid output is always parsed.


Fizzadar (approved):
Thanks @wowi42!

PR #1579, opened 2026-03-05
facts: fix requires_command shell operator precedence with compound commands

wowi42:
When a fact uses `requires_command` and its command contains `&&` chains (e.g. runit.RunitStatus), the generated shell command has incorrect operator precedence:

  ! command -v sv >/dev/null || export SVDIR=... && cd "$SVDIR" && ...

Shell parses `||` and `&&` left-to-right, so when `sv` is missing the `||` short-circuits but subsequent `&&` clauses still run and fail, causing the fact to error instead of returning the default value.

Wrap the fact command in a subshell so `||` applies to the entire chain:

  ! command -v sv >/dev/null || ( export SVDIR=... && cd "$SVDIR" && ... )

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -240,7 +240,9 @@ def _get_fact(
             requires_command,
             ">/dev/null",
             "||",
+            "(",
             command,
+            ")",
         )
 
     status = False


Fizzadar (approved):
Oh wow nice catch, surprised we've not hit this before!

PR #1582, opened 2026-03-05
facts.server.SecurityLimits: handle missing limits.conf

wowi42:
When `/etc/security/limits.conf` does not exist (e.g. on Alpine/minimal systems), `cat` fails and pyinfra errors instead of returning the default empty list. Suppress stderr and force exit 0 so a missing file is treated as no limits configured.

### DIFF
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -968,7 +968,7 @@ class SecurityLimits(FactBase):
 
     @override
     def command(self):
-        return "cat /etc/security/limits.conf"
+        return "cat /etc/security/limits.conf 2>/dev/null || true"
 
     default = list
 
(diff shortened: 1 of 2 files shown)


src/pyinfra/facts/server.py
```diff
@@ -968,7 +968,7 @@ class SecurityLimits(FactBase):
 
     @override
     def command(self):
-        return "cat /etc/security/limits.conf"
+        return "cat /etc/security/limits.conf 2>/dev/null || true"
```
Fizzadar:
```suggestion
        return "! test -e /etc/security/limits.conf || cat /etc/security/limits.conf"
```

This way we still catch errors calling `cat` which the `|| true` method silently ignores.

Fizzadar (changes requested):
One minor tweak!

wowi42:
@Fizzadar done

Fizzadar (approved):

PR #1588, opened 2026-03-05
facts.opkg: add requires_command and stop hardcoding /bin/opkg

wowi42:
All five opkg fact classes were missing requires_command, causing confusing errors on hosts without opkg. Also replaced hardcoded /bin/opkg paths with just opkg so the command is resolved via PATH. (maybe a big mistake...)


Fizzadar (approved):

morrison12:
opkg is OpenWrt specific so the probability of multiple opkg binaries seems low.

PR #1597, opened 2026-03-13
feat(operations.apt): add sources_file operation

maisim:
Creates /etc/apt/sources.list.d/<filename>.sources in modern deb822 format. Supports all standard fields (Types, URIs, Suites, Components, Architectures, Signed-By) and is idempotent via AptSources fact check.

Removal (present=False) deletes the file; removal on absent file is a noop.

Tests: create, create_idempotent, remove, remove_missing

### DIFF
--- a/src/pyinfra/operations/apt.py
+++ b/src/pyinfra/operations/apt.py
@@ -4,13 +4,15 @@ Manage apt packages and repositories.
 
 from __future__ import annotations
 
+import io
 import re
 from datetime import datetime, timedelta, timezone
 from urllib.parse import urlparse
 
 from pyinfra import host
-from pyinfra.api import operation
+from pyinfra.api import OperationError, operation
 from pyinfra.facts.apt import (
+    AptSourcesFile,
     AptSources,
     SimulateOperationWillChange,
     noninteractive_apt,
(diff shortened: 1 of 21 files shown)


maisim:
Scoped diff

https://github.com/maisim/pyinfra/compare/pr3-manage-apt-key-removal...maisim:pyinfra:pr4-add-sources-file-operation

src/pyinfra/operations/apt.py
```diff
@@ -243,6 +245,121 @@ def repo(src: str, present=True, filename: str | None = None):
         )
 
 
+@operation()
+def sources_file(
+    filename: str,
+    types: list[str] | str = "deb",
+    uris: list[str] | str = "",
+    suites: list[str] | str = "",
```
Fizzadar: This and above should default to `None` rather than an empty string to avoid ambiguity.

Fizzadar (changes requested):
Finally got here! One small tweak on those default values then good to go 🙏

src/pyinfra/operations/apt.py
maisim: Hi @Fizzadar !

Done, also add few more tests:

| Fixture | What it tests |
|---|---|
| `create_minimal` | Optional fields absent (no components, architectures, signed_by) |
| `create_string_args` | String-to-list normalisation (`types="deb deb-src"`, `suites="bookworm bookworm-updates"`) |
| `error_missing_uris` | `OperationError` when `uris` is not provided |
| `error_missing_suites` | `OperationError` when `suites` is not provided |
| `create_update` | Existing file with stale content → update (different SHA) |


This is a case where we should use a dataclass to "frame" what can contain a repo definition,I will think about that

Fizzadar (approved):
LGTM! Ty for the quick turnaround @maisim 🙏

PR #1599, opened 2026-03-13
ci: upgrade default Python to 3.14

wowi42:
Update lint, type-check, and unit-test matrix to use Python 3.14 as the default version. Add 3.13 as an explicit include entry so it remains tested.


Dexmachi:
this could be really good, once py3.15 comes around, there will be significant performance upgrades, it could make Pyinfra even better

pyproject.toml
```diff
@@ -133,7 +134,6 @@ markers = [
 ]
 
 [tool.coverage.run]
-concurrency = ["gevent"]
```
Fizzadar, line 137: We need this?

uv.lock
```diff
@@ -2,10 +2,9 @@ version = 1
 revision = 3
```
Fizzadar, line 2: I think we should drop this last commit - updating the lock file as part of PRs will never scale I think we have to keep that part of 3.x branch flow only?

Fizzadar (changes requested):
Couple of notes, otherwise good to go 👍

pyproject.toml
wowi42: incompatible with Python 3.14's SysMonitor tracer and not needed

PR #1500, opened 2025-11-17
facts+operations: add uv support

morrison12:
Add initial set of facts and operations for [`uv`](https://docs.astral.sh/uv/).
Comments and Suggestions welcome.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [x] Pull request includes tests for any new/updated operations/facts
- [x] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`)

PR #1600, opened 2026-03-13
docs: fix missing keyword-only args in operations documentation

morrison12:
``./scripts/generate_operations_docs.py`` does not make keyword-only arguments visible in the document (see /1/ below taken from #1500).  It also, as visible at /2/ and /3/, struggles for some reason with the arguments to ``opkg.packages`` (even if they, separately, probably should catch up to modern Python typing).

The proposed fix uses `inspect.signature` instead of `inspect.getfullargspec` as it produces a formatted argument string including the all arguments, regardless of their kind, complete with defaults and annotations..

The one side effect of this change is that once the signature exceeds the maximum line width,  `signature.format` does _not_ pack arguments onto a line the way``build_operations_docs`` currently does (see /3/)  but places them one per line (see /4/)  the way `ruff` does.  

If preferable, I can tweak the output to restore the current (condensed) format.

- [n/a] Pull request includes tests for any new/updated operations/facts
- [n/a] Pull request includes documentation for any new/updated operations/facts
- [n/a] Tests pass (see `scripts/dev-test.sh`)
- [~] Type checking & code style passes (see `scripts/dev-lint.sh`) -- _`ruff format and check run manually`_

/1/

image

/2/
```
/Users/Someone/Projects/pyinfra-dev/docs/operations/opkg.rst:22: WARNING: Lexing literal_block 'opkg.packages(\n    packages: str | typing.List[str]="", present: <class \'bool\'>=True,\n    latest: <class \'bool\'>=False, update: <class \'bool\'>=True, **kwargs,\n)' as "python" resulted in an error at token: "'". Retrying in relaxed mode. [misc.highlighting_failure]
```

/3/

image

/4/

image

/5/

image

### DIFF
--- a/scripts/generate_operations_docs.py
+++ b/scripts/generate_operations_docs.py
@@ -124,15 +124,17 @@ def build_operations_docs():
             args_string = signature(func).format(
                 max_width=MODULE_DEF_LINE_MAX, quote_annotation_strings=False
             )
-            args_string = f"{args_string[1:-2]},\n    **kwargs" if args_string != "()" else "**kwargs"
-            args_string = args_string.replace("   ", "        ")
+            args_string = (
+                f"{args_string[:-2]},\n    **kwargs,\n" if args_string != "()" else "**kwargs,"
+            )
+            args_string = f"{args_string.replace('   ', '        ')}    )"
 
             # Attach the code block
             lines.append(
                 """
 .. code:: python
 
-    {0}.{1}({2})
+    {0}.{1}{2}
 
 """.strip().format(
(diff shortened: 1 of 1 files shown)


Fizzadar:
> The one side effect of this change is that once the signature exceeds the maximum line width, signature.format does not pack arguments onto a line the waybuild_operations_docs currently does (see /3/) but places them one per line (see /4/) the way ruff does.

Honestly this is probably better 🙃

Fizzadar (approved):
Very nice, @morrison12!

PR #1601, opened 2026-03-14
fix: parse packages with a brew "version" correctly.

morrison12:
``brew.BrewPackages`` (and thus ``brew.BrewCasks`` by inheritance  ignores the '@' separator that ``brew`` uses even though the separator is used by ``brew.packages`` (see /1/). This causes the "versions" of the formulae or cask to be reported as separate formulae or casks instead of multiple versions of a single formula or cask (see /2/) and extraneous install attempts (see Detected changes part of /2/).  With the fix applied, there are multiple versions  of a single formula or cask and no install attempt (see /3/).

- [x] Pull request is based on the default branch (`3.x` at this time) ***and started with 3.7 despite the version shown at /3/ !?!***
- [n/a] Pull request includes documentation for any new/updated operations/facts

/1/
https://github.com/pyinfra-dev/pyinfra/blob/812a1499dfb2979848bf4603fe02017c6bf149e7/src/pyinfra/operations/brew.py#L93-L95

/2/
```text
+ pyinfra --version
pyinfra: v3.7
+ brew list --versions
+ grep openssl
openssl@1.1 1.1.1w
openssl@3 3.6.1
+ pyinfra @local fact brew.BrewPackages
+ grep openssl -A 3
        "openssl@1.1": [
            "1.1.1w"
        ],
        "openssl@3": [
            "3.6.1"
        ],
        "opus": [
+ pyinfra --dry @local brew.packages openssl@3.6.1
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation func...
    [@local] Ready: packages

--> Detected changes:
    Operation                       Change       Conditional Change
    brew.packages (openssl@3.6.1)   1 (@local)   -

```

/3/
```text
+ pyinfra --version
pyinfra: v3.6
+ brew list --versions
+ grep openssl
openssl@1.1 1.1.1w
openssl@3 3.6.1
+ pyinfra @local fact brew.BrewPackages
+ grep openssl -A 3
        "openssl": [
            "1.1.1w",
            "3.6.1"
        ],
+ pyinfra --dry @local brew.packages openssl@3.6.1
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation func...
    [@local] Ready: packages

--> Detected changes:
    Operation                       Change   Conditional Change
    brew.packages (openssl@3.6.1)   -        -

```


Fizzadar (changes requested):
I'm not sure this is the right fix since brew uses `@` for distinct formula (eg `python@3.13`, `python@3.14`). I think we'd be better off fixing this on the operation side though I'm unsure how 🤔

morrison12:
> I'm not sure this is the right fix since brew uses `@` for distinct formula (eg `python@3.13`, `python@3.14`). I think we'd be better off fixing this on the operation side though I'm unsure how 🤔

The current change works to add a version of, for example, `python` (see /1/ below) but unfortunately removing doesn't work which (see /2/ below) . As an aside, it also doesn't work without this change (see /3/) although without the error message.

In terms of changing the operation,  removing the `version_join` definition from the call to `ensure_package` might do the trick.   Will try that and see what else, if any, is needed.

/1/
```shell
(venv) machine user$ pyinfra @local fact brew.BrewPackages 2>&1 | grep python -A 3
        "python": [
            "3.12.13",
            "3.13.13",
            "3.14.4"
(venv) machine user$ pyinfra @local brew.packages python@3.10 present=True
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation func...
    [@local] Ready: packages

--> Detected changes:
    Operation                                   Change       Conditional Change
    brew.packages (python@3.10, present=True)   1 (@local)   -

    Detected changes displayed above, skip this step with -y

--> Starting operation: brew.packages (python@3.10, present=True)
    [@local] Success

--> Results:
    Operation                                   Hosts   Success   Error   No Change
    brew.packages (python@3.10, present=True)   1       1         -       -
    Grand total                                 1       1         -       -

(venv) machine user$ pyinfra @local fact brew.BrewPackages 2>&1 | grep python -A 4
        "python": [
            "3.10.20",
            "3.12.13",
            "3.13.13",
            "3.14.4"


/2/
(venv) machine user$ pyinfra  @local brew.packages python@3.10.20 present=False
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation func...
    [@local] Ready: packages

--> Detected changes:
    Operation                                       Change       Conditional Change
    brew.packages (python@3.10.20, present=False)   1 (@local)   -

    Detected changes displayed above, skip this step with -y

--> Starting operation: brew.packages (python@3.10.20, present=False)
    [@local] Error: No available formula with the name "python@3.10.20". Did you mean python@3.12, python@3.10, python@3.14, python@3.13 or python@3.11?
    [@local] Error: executed 0 commands

/3/
(venv) machine user$ pyinfra  @local brew.packages python@3.10.20 present=False
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation func...
    [@local] Ready: packages

--> Detected changes:
    Operation                                       Change   Conditional Change
    brew.packages (python@3.10.20, present=False)   -        -

    Detected changes displayed above, skip this step with -y

--> Starting operation: brew.packages (python@3.10.20, present=False)
    [@local] No changes

--> Results:
    Operation                                       Hosts   Success   Error   No Change
    brew.packages (python@3.10.20, present=False)   1       -         -       1
    Grand total                                     1       -         -       1

(venv) machine user$ pyinfra @local fact brew.BrewPackages 2>&1 | grep python -A 4
        "python@3.10": [
            "3.10.20"
        ],
        "python@3.12": [
            "3.12.13"
        ],
        "python@3.13": [
            "3.13.13"
        ],
        "python@3.14": [
            "3.14.4"
        ],

**Review by Fizzadar — COMMENTED** [review]
Sorry for the slow reply @morrison12, and thanks for digging in with real output!

> In terms of changing the operation, removing the `version_join` definition from the call to `ensure_package` might do the trick.

Yeah I think that's the right direction. What I'd go for: keep the full formula name as the fact key (`python@3.13`, `openssl@3`), then have the brew operations look for a literal match in `current_packages` first and only fall back to `PkgInfo.from_possible_pair` if that misses. Then `brew.packages("python@3.10")` addresses the formula directly and we never have to guess whether the `@` means version or name - which should sort your /2/ case too, since `python@3.10` stays addressable instead of becoming name + version `3.10.20`.

Needs a rebase as well, both `facts/brew.py` and `operations/brew.py` have moved since March.

**Comment by morrison12** [comment]
> Sorry for the slow reply @morrison12, and thanks for digging in with real output!
> 
> > In terms of changing the operation, removing the `version_join` definition from the call to `ensure_package` might do the trick.
> 
> Yeah I think that's the right direction. What I'd go for: keep the full formula name as the fact key (`python@3.13`, `openssl@3`), then have the brew operations look for a literal match in `current_packages` first and only fall back to `PkgInfo.from_possible_pair` if that misses. Then `brew.packages("python@3.10")` addresses the formula directly and we never have to guess whether the `@` means version or name - which should sort your /2/ case too, since `python@3.10` stays addressable instead of becoming name + version `3.10.20`.
> 
> Needs a rebase as well, both `facts/brew.py` and `operations/brew.py` have moved since March.
Indeed.

No worries @Fizzadar.  I looked at it again and now think the best approach is to remove `version_join` from the call to `ensure_package` and add some notes to both the fact and the operation about `brew`'s behaviour and _stop there_... Even given a limited set of examples (see /1/ below), it would seem to require knowledge per-formula rules adjust the version number correctly, i.e. transform `python@3.14.6` into `python@3.14` and `node@22.23.2` into `node@22`.
If we leave things as distinct packages the yes `brew` may complain if the user provides too much version information but they'll get immediate feedback and can correct their target versions.

/1/
```text
(pyinfra-dev)pyinfra-dev$ brew list --versions | rg \@
icu4c@75 75.1_1
icu4c@76 76.1_2
icu4c@77 77.1_1
icu4c@78 78.3
node@22 22.23.2
openssl@1.1 1.1.1w
openssl@3 3.6.3
openssl@4 4.0.1
python@3.10 3.10.20_4
python@3.12 3.12.13_4
python@3.13 3.13.14_1
python@3.14 3.14.6
wxwidgets@3.2 3.2.11
```

morrison12:
> > Sorry for the slow reply @morrison12, and thanks for digging in with real output!
> > > In terms of changing the operation, removing the `version_join` definition from the call to `ensure_package` might do the trick.
> > 
> > 
> > Yeah I think that's the right direction. What I'd go for: keep the full formula name as the fact key (`python@3.13`, `openssl@3`), then have the brew operations look for a literal match in `current_packages` first and only fall back to `PkgInfo.from_possible_pair` if that misses. Then `brew.packages("python@3.10")` addresses the formula directly and we never have to guess whether the `@` means version or name - which should sort your /2/ case too, since `python@3.10` stays addressable instead of becoming name + version `3.10.20`.
> > Needs a rebase as well, both `facts/brew.py` and `operations/brew.py` have moved since March.
> > Indeed.
> 
> No worries @Fizzadar. I looked at it again and now think the best approach is to remove `version_join` from the call to `ensure_package` and add some notes to both the fact and the operation about `brew`'s behaviour and _stop there_... Even given a limited set of examples (see /1/ below), it would seem to require knowledge per-formula rules adjust the version number correctly, i.e. transform `python@3.14.6` into `python@3.14` and `node@22.23.2` into `node@22`. If we leave things as distinct packages the yes `brew` may complain if the user provides too much version information but they'll get immediate feedback and can correct their target versions.
> 
A few more notes:
a) `ensure_packages` does not allow `list[str | PkgInfo]` so I just gave it `packages` (`list[str]`) as it didn't seem to make a difference given the challenges noted above
b) I converted the tests to YAML so I could annotate the ones involving versions
c) there seems to be a low impact problem with `ensure_packages` not always returning a `noop_description`, will chase that done at some point (`tests/operations/brew.packages/remove_packages.yaml` is the repro).

Fizzadar (approved):
Hi @morrison12 I think this looks good now - agree the best/only way to handle this properly is just to not do the version join. Thank you for your persistence with this one also 🙏

PR #1604, opened 2026-03-15
facts.systemd: ensure that user-mode systemd facts do not fail if user manager is not available

martenlienen:
If a script uses systemd facts in user-mode anywhere but will only enable the user's manager via `loginctl enable-linger` during its first run, calls to `systemctl --user` will fail during the preparation stage in the initial run. This change makes it so that such scripts run through anyway immediately. Before, it was required to comment out all systemd commands on the first run until the user manager has been enabled.

### DIFF
--- a/src/pyinfra/facts/systemd.py
+++ b/src/pyinfra/facts/systemd.py
@@ -87,7 +87,7 @@ class SystemdStatus(FactBase[Dict[str, bool]]):
         elif isinstance(services, Iterable):
             service_strs = [QuoteString(s) for s in services]
 
-        return StringCommand(
+        cmd = StringCommand(
             fact_cmd,
             "show",
             "--all",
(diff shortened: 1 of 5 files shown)


src/pyinfra/facts/systemd.py
```diff
@@ -98,6 +98,14 @@ def command(
             *service_strs,
         )
 
+        if user_mode:
+            # In the planning stage, user managers might not be available yet. We
+            # swallow the error `Failed to connect to user scope bus ...`, so that the
+            # preparation stage does not fail.
+            return StringCommand("(", cmd, "2>/dev/null", "||", "true", ")")
```
Fizzadar, line 105: Noting that the `|| true` pattern here will over-catch errors - but parsing the output is incredibly messy as well so this will suffice for now.

Fizzadar (approved):
Good spot, thanks for fixing @martenlienen!

src/pyinfra/facts/systemd.py
martenlienen: Another solution that actually parses the output would have been nicer, but this would have required to either wrap it in a way more complex command or change the facts interface, so that the parsing can happen in python after running the command. I am all for a better solution, but since this `|| true` seems to be a recurring pattern for facts, I decided to go with it here.

Fizzadar: 100%, no issue at all we do it all over the place already. Long run it'd be nice to come up with better parsing here and everywhere it's used but as you say - pretty complicated!

PR #1609, opened 2026-03-16
meta,api: lazy load fact and operation modules

Dexmachi:
# What
This PR introduces lazy loading inside of operations/__init__.py (including freebsd) and facts/__init__.py by leveraging PEP 562 module-level __getattr__.

Instead of eagerly globbing and importing all available facts and operations upfront, modules are now only imported on-demand when explicitly requested. To ensure compatibility with test runners like pytest and avoid ModuleNotFoundError during test discovery, the dynamic imports are properly anchored using __package__.

Note: I originally experimented with lazy loading api/__init__.py as well, but reverted it as it acts as the public contract for the library and making it lazy broke several downstream tests due to circular import complexities.

# Why

The current implementation using glob and from . import * forces a massive I/O and CPU overhead whenever Pyinfra is used in API mode, loading dozens of modules that the user might never use.

By making this lazy:

- The I/O from importing facts/operations only occurs for the specific modules actually utilized in a run.

- Startup time and memory footprint are significantly reduced, especially for users wrapping Pyinfra as an execution engine.

- IDE autocompletion and static type checkers are preserved because __all__ is still dynamically generated without executing the module bodies.
    
# Addition

Now tests/test_operations.py skips execution if not inside a debian based distro (I was having issues in Arch only to discover this was the issue)

Added: ansi_color in api/utils.py, this was made in order to remove the click dependency from inside the API.

It should be noted that this should futurally be refactored in order to separate lib and interface

- [N/A] Pull request includes tests for any new/updated operations/facts -- (N/A - architecture revision)
- [N/A] Pull request includes documentation for any new/updated operations/facts --  (N/A - architecture revision)

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -16,6 +16,7 @@ from inspect import getcallargs
 from socket import error as socket_error, timeout as timeout_error
 from typing import TYPE_CHECKING, Any, Callable, Generic, Optional, Type, TypeVar, cast
 
+import click
 import gevent
 from paramiko import SSHException
 from typing_extensions import override
@@ -25,7 +26,6 @@ from pyinfra.api import StringCommand
 from pyinfra.api.arguments import all_global_arguments, pop_global_arguments
 from pyinfra.api.exceptions import FactProcessError
 from pyinfra.api.util import (
-    ansi_color,
     get_kwargs_str,
     log_error_or_warning,
     log_host_command_error,
@@ -303,7 +303,7 @@ def _get_fact(
         log_message = "{0}{1}".format(
             host.print_prefix,
             "Loaded fact {0}{1}".format(
-                ansi_color(name, bold=True),
+                click.style(name, bold=True),
                 f" ({get_kwargs_str(kwargs)})" if kwargs else "",
             ),
         )
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -16,7 +16,6 @@ from inspect import getcallargs
 from socket import error as socket_error, timeout as timeout_error
 from typing import TYPE_CHECKING, Any, Callable, Generic, Optional, Type, TypeVar, cast
 
-import click
 import gevent
 from paramiko import SSHException
 from typing_extensions import override
@@ -26,6 +25,7 @@ from pyinfra.api import StringCommand
 from pyinfra.api.arguments import all_global_arguments, pop_global_arguments
 from pyinfra.api.exceptions import FactProcessError
 from pyinfra.api.util import (
+    ansi_color,
     get_kwargs_str,
     log_error_or_warning,
     log_host_command_error,
@@ -303,7 +303,7 @@ def _get_fact(
         log_message = "{0}{1}".format(
             host.print_prefix,
             "Loaded fact {0}{1}".format(
-                click.style(name, bold=True),
+                ansi_color(name, bold=True),
                 f" ({get_kwargs_str(kwargs)})" if kwargs else "",
             ),
         )
--- a/src/pyinfra/facts/__init__.py
+++ b/src/pyinfra/facts/__init__.py
@@ -1,12 +1,128 @@
-# This file only exists to support:
-# from pyinfra import facts
-# facts.X.Y
+import importlib
+from typing import TYPE_CHECKING
 
-from glob import glob
-from os import path
+if TYPE_CHECKING:
+    from . import (
+        apk,
+        apt,
+        brew,
+        bsdinit,
+        cargo,
+        choco,
+        crontab,
+        deb,
+        dnf,
+        docker,
+        efibootmgr,
+        files,
+        flatpak,
+        freebsd,
+        gem,
+        git,
+        gpg,
+        hardware,
+        iptables,
+        launchd,
+        lxd,
+        mysql,
+        npm,
+        openrc,
+        opkg,
+        pacman,
+        pip,
+        pipx,
+        pkg,
+        pkgin,
+        podman,
+        postgres,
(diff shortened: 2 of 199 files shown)


Dexmachi:
this does fix this [issue](https://github.com/pyinfra-dev/pyinfra/issues/1605), as seen in the following files:

[pyinfra_fact_import_new.txt](https://github.com/user-attachments/files/26032406/pyinfra_fact_import_new.txt)
[pyinfra_op_import_new.txt](https://github.com/user-attachments/files/26032407/pyinfra_op_import_new.txt)

As it is possible to see, only the bare minimum for each of these was imported.

I would still advise to try and lazify the imports to inside specific functions instead of importing everything in the start of the file, as this would heavily impact import startup.

Dexmachi:
sorry for the late show

either way, it is working perfectly:
image


the main issue is with the syntax
```py
from pyinfra import operations

operations.server.shell
```

image


it does not load them in static evaluation for calling stuff like operations.server.shell, but this is was an issue before, so yeah

src/pyinfra/api/facts.py
```diff
@@ -16,7 +16,6 @@
 from socket import error as socket_error, timeout as timeout_error
 from typing import TYPE_CHECKING, Any, Callable, Generic, Optional, Type, TypeVar, cast
 
-import click
```
Fizzadar, line 19: I don't think we should remove click here, it's used all over the API and it's behavior is needed. In particular `ansi_color` doesn't reset when passed "", click does.

I don't think click itself adds significant load times? I'm not completely against removing it but definitely should be a separate PR.

src/pyinfra/facts/util/__init__.py
```diff
@@ -1,3 +1,5 @@
+# this should be moved to an utility module, _IMO_, but I won't touch it since it's OOB from the scope of this PR.
```
Fizzadar: Needs removing before merge!

(also agree, just out of scope)

src/pyinfra/facts/__init__.py
```diff
@@ -1,12 +1,26 @@
...
+import importlib
 from glob import glob
 from os import path
 
-module_filenames = glob(path.join(path.dirname(__file__), "*.py"))
-module_names = [path.basename(name)[:-3] for name in module_filenames]
-__all__ = [name for name in module_names if name != "__init__"]
+# Lazily discover and build __all__
+# Can be further optimized by using a hardcoded set
+# but this would be the worse decision ever
```
Fizzadar: Please don't add comments like this in here (it would be a terrible decision though!).

I wonder if we could use package entrypoints though, out of scope anyway.

src/pyinfra/operations/freebsd/__init__.py
```diff
@@ -1,12 +1,31 @@
...
 
-from . import *  # noqa
+# # This file only exists to support:
+# # from pyinfra.operations import freebsd
+# # freebsd.X.Y
+#
+# from glob import glob
+# from os import path
+#
+# module_filenames = glob(path.join(path.dirname(__file__), "*.py"))
```
Fizzadar: Let's drop all these please

tests/test_operations.py
```diff
@@ -88,6 +89,11 @@ def assert_commands(commands, wanted_commands):
 def make_operation_tests(arg):
     # Get the operation we're testing against
     module_name, op_name = arg.rsplit(".", 1)
+
+    # Some operations only work on certain distros, so we skip them
+    if module_name == "apt" and distro.id() not in ("debian", "ubuntu", "raspbian"):
```
Fizzadar: Unrelated to PR, remove. But good flag - we should make this part of the operation/fact JSON spec though, we already have this pattern for Windows.

Fizzadar (changes requested):
The import improvements are 🔥, awesome stuff!

Other bits should be broken out or removed though please.

Dexmachi:
> The import improvements are 🔥, awesome stuff!
> 
> Other bits should be broken out or removed though please.

Points taken!

I already fixed what was asked to be fixed (very simple shtuff altogether, so easy fix), however, I'm in an area that blocks port 22, so I'll take a while to push!

all tests (except debian based ones) are passing, and the code is linted

src/pyinfra/facts/__init__.py
Dexmachi: I think entrypoints could be useful, for these types of code

personally, my project deals with this like so

discovers all entry_points -> caches them persistently inside of a json file (O(1) instead of O(N) plugins) -> the name of the role should be the key in the json file, so it isn't required to load the entire package in order to discover it

src/pyinfra/api/facts.py
Dexmachi: I personally would say that the library should be agnostic from any and all interface behavior, this would reduce the overhead due to another library usage and make the API more agnostic, whereas right now it is kind of "made" for CLIs, although ansi_color could reset when passed "" in order to do what it is needed without the need of a library only for simple styling.



I agree that it is out of scope, so I removed it like asked.

src/pyinfra/operations/freebsd/__init__.py
Dexmachi: Allright!

I kept these so I'd have a fallback if this PR wasn't accepted on the first try

tests/test_operations.py
Dexmachi: agreed that it is out of scope

Fizzadar (approved):
This is very awesome @Dexmachi, thank you for working on this 🙏

Dexmachi:
> This is very awesome @Dexmachi, thank you for working on this 🙏

Its an honor to be a part of this!

PR #1614, opened 2026-03-17
docs: fix generate_operations_docs.py and add docs build to CI

wowi42:
## Summary

Fix broken documentation build caused by an invalid `quote_annotation_strings` kwarg passed to `inspect.Signature.format()` (introduced in #1600), and resolve all 372 Sphinx build warnings.

Also adds a `docs-build` job to the CI test workflow so documentation generation is validated on every PR without publishing.

### Fix doc build failure
- Remove invalid `quote_annotation_strings` kwarg from `inspect.Signature.format()` in `generate_operations_docs.py`

### Fix all Sphinx warnings (372 → 0)

**Generated operations docs (~340 warnings):**
- Add blank line before note directive and use proper `.. note::` RST admonition
- Fix stray comma in multiline signatures
- Escape `*` prefixes in `format_doc_line` for `**kwargs`-style args
- Convert `gpg.py` docstrings to standard pyinfra format (`+ arg` style)

**Hand-written docs (~30 warnings):**
- Fix broken `:ref:` cross-references in `faq.rst` and `using-operations.rst` by adding explicit RST labels to `generate_arguments_doc.py` and `deploy-process.rst`
- Deduplicate heading text in `install.md` for `autosectionlabel`
- Fix H1→H3 header level jumps in `api/facts.md` and `api/operations.md`
- Mark orphaned `examples.rst` and `examples/*.md` with `:orphan:`
- Suppress `myst.header` warnings from `CHANGELOG.md` in `conf.py`

### CI: add docs build job
- Add `docs-build` job to `test.yml` that runs after lint checks
- Include it in the `tests-done` gate

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,7 +108,31 @@ jobs:
       - run: uv sync --group test --no-default-groups --python=${{ matrix.python-version }}
       - run: uv run pytest -m end_to_end_${{ matrix.test-name }}
 
-  # Finally, single jon to check if all completed, for use in protected branch
+  docs-build:
+    needs:
+      - lint
+      - type-check
+      - spell-check
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v5
+      - name: Fetch git tags
+        run: git fetch --tags origin
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+      - name: Install
+        run: |
+          uv sync --group docs --no-default-groups
+          uv pip install 'jinja2<3.1'
+      - name: Generate documentation
+        run: scripts/build-public-docs.sh
+        env:
+          PYTHONPATH: '.'
+
+  # Finally, single job to check if all completed, for use in protected branch
 
   tests-done:
     if: ${{ always() }}
@@ -118,6 +142,7 @@ jobs:
       - spell-check
       - unit-test
       - end-to-end-test
+      - docs-build
     runs-on: ubuntu-latest
     steps:
       - uses: matrix-org/done-action@v3
--- a/docs/api/facts.md
+++ b/docs/api/facts.md
@@ -9,7 +9,7 @@ occurs during fact collection. Additionally, a ``requires_command`` variable can
 on the host to collect the fact. If this command is not present on the host, the fact will be set to the default, or empty if no ``default`` function
 is available.
 
-### Importing & Using Facts
+## Importing & Using Facts
 
 Like operations, facts are imported from Python modules and executed by calling `Host.get_fact`. For example:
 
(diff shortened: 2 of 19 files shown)


morrison12:
Quick question on  ``quote_annotation_strings``:

Presumably the problem was that this only is available from python 3.13 on but we need to support both the pyinfra code and the developments tools from 3.10 on ?  

Would that be a useful reminder  in CONTRIBUTING.md ?  A

sking as I noted the useful docs-build CI job uses 3.13 where, ironically, `quote_annotation_strings`` should not cause an issue.

wowi42:
Good question! Actually, `quote_annotation_strings` is **not** available in Python 3.13,
 the 3.13 [`Signature.format()`](https://docs.python.org/3.13/library/inspect.html#inspect.Signature.format) only accepts `max_width`. Python 3.14 added a related parameter called `unquote_annotations` ([What's New in 3.14](https://docs.python.org/3.14/whatsnew/3.14.html#inspect)), but `quote_annotation_strings` as a kwarg doesn't exist in either release's public API.

So the docs build was failing on 3.13 too, it wasn't a 3.10 vs 3.13 compat issue, it was just an invalid keyword argument. Removing it was the right fix since we don't actually need to control annotation quoting for the generated docs.

Regarding a CONTRIBUTING.md note: the doc generation scripts do require Python 3.13+ (because of `Signature.format()` with `max_width`), so that could be worth mentioning. Happy to add that if you think it'd help.

morrison12:
> Good question! Actually, `quote_annotation_strings` is **not** available in Python 3.13, the 3.13 [`Signature.format()`](https://docs.python.org/3.13/library/inspect.html#inspect.Signature.format) only accepts `max_width`. Python 3.14 added a related parameter called `unquote_annotations` ([What's New in 3.14](https://docs.python.org/3.14/whatsnew/3.14.html#inspect)), but `quote_annotation_strings` as a kwarg doesn't exist in either release's public API.

Looks like something curious in the docs as it is shown in the prototype for 3.14 (see /1/) and it works in 3.14 but `unquote_annotations` doesn't !?! (see /2/) .  

> So the docs build was failing on 3.13 too, it wasn't a 3.10 vs 3.13 compat issue, it was just an invalid keyword argument. Removing it was the right fix since we don't actually need to control annotation quoting for the generated docs.

The only/specific  reason `quote_annotation_strings` was added was to get rid of an error I with docs for ``opkg`` but looks like you took care of that in your awesome docs cleanup so it isn't needed.

> 
> Regarding a CONTRIBUTING.md note: the doc generation scripts do require Python 3.13+ (because of `Signature.format()` with `max_width`), so that could be worth mentioning. Happy to add that if you think it'd help.

In general it would probably be good to render expectations explicit.  Thinking about it, perhaps also in configuration files/stanzas but that is separate and in any event likely requires some experimentation.

If `inspect.Signature.format` is the only thing pushing the toolchain from 3.10 to 3.13, I could look at making a back port.


/1/  [source](https://docs.python.org/3/library/inspect.html)
image

/2/
image

Fizzadar (approved):
Amazing, thank you @wowi42! Good call on putting it (back) into CI - keeps burning us there's no reason not to.

Fizzadar:
@morrison12 I think it's reasonable for us to expect contributors to use the latest, or near latest, Python. Can update the doc to explicitly call this out. No need to backport things for dev tooling.

Issue #1607, opened 2026-03-15
Vars not being passed to j2 files in files.template operation


EricDriussi:
## Describe the bug
When using the `files.template` operation, vars are not passed to the source j2 file as expected.

## To Reproduce
Steps to reproduce the behavior, please include where possible:

```python
def dummy():
    secret_key = server.shell(
        name="Generate a secret",
        commands=["openssl rand -hex 16"],
        _shell_executable="/bin/bash",
    ).stdout

    files.template(
        name="Copy settings",
        src="tasks/services/templates/settings.yml.j2",
        dest="/settings.yml",
        mode="644",
    )


python.call(
    name="Dummy",
    function=dummy,
)
```

Where the `settings.yml.j2` file looks like:

```j2
secret_key: "{{ secret_key }}"
```

The error I'm getting is:

```
Error in template: /.../settings.yml.j2 (L11): \'secret_key\' is undefined
```

## Expected behavior
Given the [documentation for this operation](https://docs.pyinfra.com/en/3.x/operations/files.html#files-template), I expected my setup to work and set the var in the file (as it does when getting the var from host).
I might not be understanding the docs. If that is the case, how should I get that var into the template?

## Meta
+ Include output of `pyinfra --support`.
```
System: Linux
      Platform: Linux-6.19.6-arch1-1-x86_64-with-glibc2.43
      Release: 6.19.6-arch1-1
      Machine: x86_64
    pyinfra: v3.7
      click: v8.3.1
      distro: v1.9.0
      gevent: v25.9.1
      jinja2: v3.1.6
      packaging: v26.0
      paramiko: v3.5.1
      pydantic: v2.12.5
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.1
      typing-extensions: v4.15.0
    Executable: /home/.../project/.venv/bin/pyinfra
    Python: 3.13.4 (CPython, Clang 20.1.4 )
```
+ How was pyinfra installed (source/pip)?
Installed using uv

wowi42:
Hi @EricDriussi! The template operation doesn't automatically capture local Python variables — you need to pass them explicitly as keyword arguments. The `**kwargs` in the signature is exactly for this purpose.

Change your call to:

```python
files.template(
    name="Copy settings",
    src="tasks/services/templates/settings.yml.j2",
    dest="/settings.yml",
    mode="644",
    secret_key=secret_key,  # <-- pass it explicitly
)
```

[...]

PR #1615, opened 2026-03-17
docs: operations.files - clarify template vars use

EricDriussi:

Hi there!
I misreported a bug in #1607 after not understanding the documentation on templates.
Specifically, I thought the part about passing custom vars to the template only applied to templates passed as IO-like src, maybe because both concepts are presented together in the examples section (or maybe it was too late in the evening for me to be reading docs...).

This adds a dedicated example for IO-like src and (hopefully) a more explicit explanation on how to get custom data to the template file to make the docs more dummy proof.


Fizzadar (approved):
Much better, thank you @EricDriussi!

PR #1617, opened 2026-03-18
Migrate shlex.quote() to StringCommand + QuoteString across the codebase

### LINKED ISSUE #1549: Mitigating command injection - security issue

wowi42:
## Summary

Follow-up to #1576 as suggested by @Fizzadar: migrate all direct `shlex.quote()` calls to pyinfra's built-in `StringCommand` + `QuoteString` abstractions. This centralizes shell quoting at the command layer instead of scattering `shlex.quote()` throughout individual operations, facts, and connectors.

After this change, the only `import shlex` remaining in `src/` is in `api/command.py` — where `QuoteString` is implemented.

### What changed

**Operations** (8 files):
- `git.py`, `npm.py`, `pip.py`, `server.py`, `ssh.py` — yield `StringCommand` objects with `QuoteString`-wrapped user inputs instead of format strings with `shlex.quote()`
- `systemd.py` — `StringCommand` for enable/disable
- `crontab.py` — `StringCommand` for echo commands
- `util/service.py` — `make_formatted_string_command` with `QuoteString` for all init system service control (systemd, sysvinit, openrc, bsdinit, upstart, runit, launchd)
- `util/packaging.py` — `QuoteString` in `PkgInfo.inst_vers`; widened type hints to accept `StringCommand` command parameters

**Facts** (4 files):
- `rpm.py`, `pacman.py`, `deb.py` — `make_formatted_string_command` with `QuoteString` for fact command construction
- `files.py` — `StringCommand` for tilde path quoting

**Connectors** (2 files):
- `connectors/ssh.py` — `StringCommand` for SSH flag construction
- `connectors/util.py` — `QuoteString` for askpass paths, passwords, and win command quoting

### Tests

Added **30 new test fixtures** proving command injection is mitigated with shell metacharacters (`;`, `$()`, spaces) in user-supplied values:

- **ssh** (4): keyscan hostname, keyscan force, command user@host, upload user@host
- **git** (5): config key, config multivalue, repo checkout, worktree remove, worktree pull remote
- **pip** (2): virtualenv python path, packages virtualenv path
- **server** (8): group delete (Linux + FreeBSD), group add FreeBSD, mount remount, modprobe remove, user edit, user add FreeBSD, user add with groups
- **systemd/init** (4): systemd enable/disable, sysvinit start, openrc start, bsdinit start
- **crontab** (2): add with quoted command, add_named with quoted name
- **facts** (4): rpm.RpmPackage, rpm.RpmPackageProvides, deb.DebPackage, pacman.PacmanUnpackGroup
- **npm** (1): directory with injection characters (from #1576)

Fix #1549

### DIFF
--- a/src/pyinfra/connectors/ssh.py
+++ b/src/pyinfra/connectors/ssh.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import shlex
 from random import uniform
 from shutil import which
 from socket import error as socket_error, gaierror
@@ -671,18 +670,28 @@ class SSHConnector(BaseConnector):
         known_hosts_file = self.data["ssh_known_hosts_file"]
         if known_hosts_file:
             ssh_flags.append(
-                '-o \\"UserKnownHostsFile={0}\\"'.format(shlex.quote(known_hosts_file))
+                StringCommand(
+                    '-o \\"UserKnownHostsFile=',
+                    QuoteString(known_hosts_file),
+                    '\\"',
+                    _separator="",
+                ).get_raw_value()
             )  # never trust users
 
         strict_host_key_checking = self.data["ssh_strict_host_key_checking"]
         if strict_host_key_checking:
             ssh_flags.append(
-                '-o \\"StrictHostKeyChecking={0}\\"'.format(shlex.quote(strict_host_key_checking))
+                StringCommand(
+                    '-o \\"StrictHostKeyChecking=',
+                    QuoteString(strict_host_key_checking),
+                    '\\"',
+                    _separator="",
+                ).get_raw_value()
             )
 
         ssh_config_file = self.data["ssh_config_file"]
         if ssh_config_file:
-            ssh_flags.append("-F {0}".format(shlex.quote(ssh_config_file)))
+            ssh_flags.append(StringCommand("-F", QuoteString(ssh_config_file)).get_raw_value())
 
         port = self.data["ssh_port"]
         if port:
--- a/src/pyinfra/connectors/util.py
+++ b/src/pyinfra/connectors/util.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import shlex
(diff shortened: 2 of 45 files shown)


Fizzadar (approved):
Fantastic 🎉 Thank you as always, @wowi42!

(mind rebasing this as well so I can merge 🙏)

wowi42:
@Fizzadar Done

PR #1628, opened 2026-03-22
Show askpass generation errors

matthijskooijman:
This PR ensures that if the askpass command generation fails, a proper error message is shown, instead of an IndexError with no further hints.

See the commit messages for details.

### DIFF
--- a/src/pyinfra/connectors/util.py
+++ b/src/pyinfra/connectors/util.py
@@ -293,9 +293,7 @@ def _ensure_askpass_set_for_host(host: "Host", key: str, env_var: str):
             )
         )
 
-    host.connector_data[key] = StringCommand(
-        QuoteString(output.stdout_lines[0])
-    ).get_raw_value()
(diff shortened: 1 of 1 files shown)


matthijskooijman:
I just (hopefully) fixed some linter errors, but I was a bit surprised by them. Seems the github workflow runs `ruff format --check`, while the `scripts/dev-lint.sh` script only runs `ruff check` which is apparently something different. Maybe the script should be matched to the workflow?

Also, running the workflow with `--diff` might be useful, so you can directly see what the problem is.

Fizzadar (approved):
Ah so much better, @matthijskooijman, thank you!

Fizzadar:
> I just (hopefully) fixed some linter errors, but I was a bit surprised by them. Seems the github workflow runs `ruff format --check`, while the `scripts/dev-lint.sh` script only runs `ruff check` which is apparently something different. Maybe the script should be matched to the workflow?
> 
> Also, running the workflow with `--diff` might be useful, so you can directly see what the problem is.

Agreed! Updated main branch accordingly, just hit this with the conflict I resolved.

PR #1647, opened 2026-03-31
feat: add requires_command guard sentinel and check_preconditions() hook

maisim:
## Background

Two related reliability gaps existed in the fact-collection layer:

1. The old shell guard for `requires_command` used:

       ! command -v <bin> >/dev/null || (<fact-command>)

   When the binary is absent this produces *no stdout*, making it impossible to distinguish "binary not installed" from "binary installed but returned no data".  Both cases silently return `default()`, masking real errors.

2. There was no way for a fact to express runtime prerequisites beyond "this binary must exist" — e.g. ZFS datasets cannot be listed when the `zfs` kernel module is not loaded, even if the `zfs` binary is present.

## Changes

Replace the old `! … ||` guard with an explicit `if/then/else` that emits a unique marker when the binary is absent:

    if command -v <bin> >/dev/null 2>&1; then
        (<fact-command>);
    else
        echo '##PYINFRA_NOCMD##';
    fi

The marker `_MISSING_COMMAND_MARKER = "##PYINFRA_NOCMD##"` is detected in `_get_fact` and converted into a `MissingCommandError` exception so callers can tell the two cases apart.

    FactError
    └── FactNotCollected          # base: fact skipped, not a data error
        ├── MissingCommandError   # requires_command binary absent
        └── FactPreconditionError # check_preconditions() returned a reason

All three are exported from `pyinfra.api`.

Both exceptions obey the deploy phase in `get_fact()`:

- **Prepare phase**: silently return `default()`.  The binary / module may simply not be installed yet; a later operation will install it.
- **Execute phase**: re-raise.  The binary / module should already be present; its absence indicates incorrect ordering in the deploy.

New optional override on `FactBase`:

```python
def check_preconditions(self, state, host) -> str | None:
```

Return `None` (or no explicit return) to proceed, or a reason string to skip with context.  The framework raises `FactPreconditionError` automatically — fact authors never need to import exception classes.

As an example:
* `ZfsPools` added with `requires_command = "zpool"`.
* `ZfsDatasets.check_preconditions()` uses the existing `server.KernelModules` fact to verify the `zfs` kernel module is loaded before attempting `zfs get -H all` — a real-world scenario where `requires_command` alone is not sufficient.

## Tests

* `tests/test_api/test_api_facts_missing_command.py` — phase-aware behaviour for `MissingCommandError` and the `##PYINFRA_NOCMD##` sentinel.
* `tests/test_api/test_api_facts_preconditions.py` — phase-aware behaviour for `FactPreconditionError` via `check_preconditions()`.

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -24,7 +24,7 @@ from pyinfra import logger
 from pyinfra.api.output import format_text
 from pyinfra.api import StringCommand
 from pyinfra.api.arguments import all_global_arguments, pop_global_arguments
-from pyinfra.api.exceptions import FactProcessError
+from pyinfra.api.exceptions import FactPreconditionError, FactProcessError, MissingCommandError
 from pyinfra.api.util import (
     get_kwargs_str,
     log_error_or_warning,
@@ -36,9 +36,13 @@ from pyinfra.context import ctx_host, ctx_state
 from pyinfra.progress import progress_spinner
 
 from .arguments import CONNECTOR_ARGUMENT_KEYS
+from .state import StateStage
 
 if TYPE_CHECKING:
     from pyinfra.api import Host, State
+
+# Sentinel output line emitted when skip_unless_command binary is absent on the remote host.
+_MISSING_COMMAND_MARKER = "##PYINFRA_NOCMD##"
 
 SUDO_REGEX = r"^sudo: unknown user"
 SU_REGEXES = (
@@ -60,6 +64,22 @@ class FactBase(Generic[T]):
     command: Callable[..., str | StringCommand]
 
     def requires_command(self, *args, **kwargs) -> str | None:
+        """Return the binary name that must exist on the remote host for this fact to run.
+        If the binary is absent the fact returns its ``default()`` value silently.
+        """
+        return None
+
+    def check_preconditions(self, state: "State", host: "Host") -> str | None:
+        """Check that this fact's prerequisites are satisfied before running.
+
+        Override this method to call ``host.get_fact(...)`` and return:
+
+        - ``None`` (or no return) — all prerequisites satisfied, proceed normally
+        - ``"reason message"`` — prerequisite not satisfied with explanation
+
+        The framework handles raising ``FactPreconditionError`` and phase-awareness
+        automatically; fact authors never need to import exception classes.
+        """
         return None
 
     @override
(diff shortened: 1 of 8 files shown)


src/pyinfra/api/facts.py
```diff
@@ -60,8 +64,27 @@ class FactBase(Generic[T]):
     command: Callable[..., str | StringCommand]
 
     def requires_command(self, *args, **kwargs) -> str | None:
+        """Return the binary name that must exist on the remote host for this fact to run.
+        If the binary is absent the fact returns its ``default()`` value silently.
+        """
         return None
 
+    def preconditions(
```
DonDebonair: Given the return value, I would make the function name a bit more descriptive. Something `preconditions_met`?

src/pyinfra/api/facts.py
```diff
@@ -60,8 +64,27 @@ class FactBase(Generic[T]):
...
+        self, state: "State", host: "Host"
+    ) -> bool | tuple[Literal[False], str] | None:
```
DonDebonair: Why allow `None` as a return value? I think the signature is cleaner if it's just `bool | tuple[Literal[False], str]`

src/pyinfra/facts/zfs.py
DonDebonair, line 1: I would move the changes to this file to another PR, or at least a separate commit. It's nice to keep PRs/commits atomic.

DonDebonair (commented):
I love this PR! Left a few small suggestions.

src/pyinfra/api/facts.py
maisim: That seems relevant to me.

src/pyinfra/api/facts.py
maisim: The goal if to not have to return True "manually" if all is ok :
```python
    @override
    def preconditions_met(self, state, host):
        from pyinfra.facts.server import KernelModules

        modules = host.get_fact(KernelModules) or {}
        if "zfs" not in modules:
            return False, "kernel module 'zfs' is not loaded"
        return True # <-- implies to always add return True at the end of the method
```
Explicit is better than impliciit, maybe we can do like that

Another way if to just return the reason, a string, and not False or a tulple `False, "resaon", because if we return anithing, it's because there is a problem

```python
    @override
    def preconditions(self, state, host):
        from pyinfra.facts.server import KernelModules

        modules = host.get_fact(KernelModules) or {}
        if "zfs" not in modules:
            return  "kernel module 'zfs' is not loaded"
```
It might be less intuitive
What do you think ?

src/pyinfra/facts/zfs.py
maisim: It was at an example of usage ;) I put it in an separate commit

DonDebonair:
Looks good to me. @Fizzadar ready to merge. Recommend not squashing the commits because they represent 2 different (but related) features.

DonDebonair (approved):

src/pyinfra/api/facts.py
DonDebonair: That is actually a good point. Suggestion: rename the method to `check_preconditions(...)` and make it return `str | None`. If it returns a string, some preconditions were not met and the return value gives more details. If nothing (`None`) is returned, all is well.

src/pyinfra/api/facts.py
```diff
@@ -186,15 +206,45 @@ def get_fact(
...
+                "Fact %s skipped on %s during prepare: %s",
+                cls.__name__,
+                host.print_prefix,
+                e,
+            )
+            return cls().default()
+        # During the execute phase the binary should already be present.  If it
+        # isn't, the deploy is incorrectly ordered (missing an install step).
+        # Re-raise so the error surface is clear and actionable.
+        raise
```
Fizzadar: Possibly change to warning log to avoid breaking change?

Fizzadar (approved):
I love this PR, this is vastly improved vs. the current behavior. My concern is with the breaking change to raise during execution. I think this is quite likely to break existing working deploys that rely on the defaulting behaviour, something like:

```py
# Executed on both webserver.net + dbserver.net, default (due to missing command) on webserver is silently skipped
if host.get_fact(postgres.Databases):
   ...
```

At the same time, I don't want this stuck waiting for v4.  What about logging, rather than raising, during execution phase (perhaps with config to enable raising?).

@maisim @DonDebonair keen for your thoughts!

src/pyinfra/api/facts.py
maisim: @Fizzadar 
Maybe we should log a warning in v3 and inform it wil raise in v4 ? Or do you think log is the good behavor? (even in v4)

Fizzadar: Yeah I'm thinking warning log (noting that v4 will raise) and we'll make the breaking change in v4.

src/pyinfra/api/facts.py
```diff
@@ -186,15 +206,49 @@ def get_fact(
...
+            )
+            return cls().default()
+        # TODO(v4): remove this compat shim and let the exception propagate.
+        logger.warning(
+            "Fact %s skipped on %s: command not found: %s (this will raise an exception in v4)",
+            cls.__name__,
+            host.print_prefix,
+            e,
+        )
+        return cls().default()
```
maisim, line 238: Made the change here, It does not raise anymore, just log a warning.

Fizzadar (approved):
Awesome PR, huge QOL improvement! Thank you @maisim 🙏

PR #1650, opened 2026-04-02
zfs: return an empty dict if zfs/zpool commands are not available

yacoob:
If the required command is not installed, a default value is returned. `FactsBase.default = None`, which causes `zfs.dataset` to fail [here](https://github.com/pyinfra-dev/pyinfra/blob/cedf47eeab8882f15a54e2a70c8a7ee661319a37/src/pyinfra/operations/zfs.py#L54).

With this PR in, both zfs.Pools and zfs.Datasets will return `{}` when the relevant binaries are missing.

### DIFF
--- a/src/pyinfra/facts/zfs.py
+++ b/src/pyinfra/facts/zfs.py
@@ -26,6 +26,8 @@ class ZfsPools(FactBase):
     def requires_command(self) -> str:
         return "zpool"
 
+    default = dict
+
     @override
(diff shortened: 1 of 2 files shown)


maisim:
Related to

https://github.com/pyinfra-dev/pyinfra/pull/1647

Fizzadar (approved):
Oh great spot @yacoob, thank you for fixing this!

PR #1653, opened 2026-04-03
connectors: use gevent.subprocess in util for macOS + Python 3.13 compatibility (#1653)

### LINKED ISSUE #1652: @local connector fails on macOS with Python 3.13: subprocess._fork_exec is None
## Environment

- pyinfra: 3.7
- gevent: 25.9.1
- Python: 3.13.7 and 3.13.12 (tested both)
- macOS 15.x (arm64)

## Description

The `@local` connector fails with `TypeError: 'NoneType' object is not callable` when executing any operation on macOS with Python 3.13. The issue is that after `gevent.monkey.patch_all()` (called in `pyinfra_cli/__init__.py`), `subprocess._fork_exec` is set to `None`, and pyinfra's local connector triggers a code path that calls it.

## Minimal Reproduction

```bash
pyinfra -y @local server.shell -- "echo hello"
```

## Error

```
File ".../pyinfra/connectors/util.py", line 46, in run_local_process
    process = Popen(command, shell=True, stdout=PIPE, stderr=PIPE, stdin=PIPE)
  File ".../subprocess.py", line 1039, in __init__
    self._execute_child(...)
  File ".../subprocess.py", line 1902, in _execute_child
    self.pid = _fork_exec(...)
TypeError: 'NoneType' object is not callable
```

## Analysis

gevent deliberately sets `subprocess._fork_exec = None` and provides its own `Popen._execute_child` implementation that uses `fork_and_watch()` instead. However, there appears to be a signature mismatch between gevent's `_execute_child` override and CPython 3.13's version:

- **CPython 3.13** removed the `universal_newlines` parameter from `_execute_child`
- **gevent 25.9.1** still includes it in the override signature (25 params vs 24)

This causes the CPython original `_execute_child` (which uses `_fork_exec`) to be called instead of gevent's override.

**Important**: Direct `subprocess.Popen()` calls work fine — even inside greenlets. The issue only manifests through pyinfra's `@local` connector execution path.

## What works

```python
# All of these work fine after monkey.patch_all():
subprocess.Popen("echo hello", shell=True)  # OK
gevent.subprocess.Popen("echo hello", shell=True)  # OK
gevent.spawn(lambda: subprocess.Popen("echo hello", shell=True))  # OK

# Only pyinfra @local fails:
pyinfra -y @local server.shell -- "echo hello"  # TypeError
```

## Workaround

None found. `--dry` mode works for planning but fact gathering (e.g., `host.get_fact(Directory)`) also triggers subprocess execution.

## Related

- gevent/gevent#2169 — Filed initially as a gevent bug, but gevent maintainer confirmed `_fork_exec = None` is intentional and the issue is not reproducible outside pyinfra

Yaminyam:
## Problem

On macOS with Python 3.13, the `@local` connector fails with:

```
File ".../subprocess.py", line 1902, in _execute_child
    self.pid = _fork_exec(...)
TypeError: 'NoneType' object is not callable
```

`_fork_exec` is intentionally set to `None` by gevent (it uses its own `fork_and_watch` instead). The issue is that inside pyinfra's greenlet operation execution context, CPython's original `Popen.__init__` gets called instead of gevent's override, which then calls CPython's `_execute_child` that depends on `_fork_exec`.

## Root Cause

`connectors/util.py` imports `from subprocess import Popen`. After `monkey.patch_all()`, `subprocess.Popen` points to `gevent.subprocess.Popen`, but in the greenlet execution context during operation runs, the resolution falls back to CPython's original implementation.

## Fix

Import `Popen` and `PIPE` directly from `gevent.subprocess` instead of relying on the monkey-patched `subprocess` module. Since pyinfra already imports `gevent` in this same file, there is no additional dependency cost.

## Testing

Before:
\`\`\`bash
$ pyinfra -y @local server.shell -- "echo hello"
# TypeError: 'NoneType' object is not callable
\`\`\`

After:
\`\`\`bash
$ pyinfra -y @local server.shell -- "echo hello"
# [@local] Success
\`\`\`

Tested on:
- Python 3.13.7 and 3.13.12, macOS arm64
- gevent 25.9.1, pyinfra 3.7

Fixes #1652

### DIFF
--- a/src/pyinfra/connectors/util.py
+++ b/src/pyinfra/connectors/util.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 from getpass import getpass
 from queue import Queue
 from socket import timeout as timeout_error
-from subprocess import PIPE, Popen
+from gevent.subprocess import PIPE, Popen
 from typing import TYPE_CHECKING, Callable, Iterable, Optional, Union
 
 import click


Fizzadar (approved):
Explicit always wins! Thank you @Yaminyam this is much much safer 🙏

PR #1654, opened 2026-04-04
selinux.port: fix a bug where the op would not find existing labels if `sepolicy` command was missing from the host

### LINKED ISSUE #1644: `SEPorts.process` stores int keys, `selinux.port` looks up str keys
Affected files:
- `pyinfra/facts/selinux.py`, `SEPorts.process` — stores [port numbers as `int` keys](https://github.com/pyinfra-dev/pyinfra/blob/4c0bc32f32ea2ccc68689829111c7a4fa8b87cc4/src/pyinfra/facts/selinux.py#L103)
- `pyinfra/operations/selinux.py`, `port()` looks up [`str(port_num)`](https://github.com/pyinfra-dev/pyinfra/blob/4c0bc32f32ea2ccc68689829111c7a4fa8b87cc4/src/pyinfra/operations/selinux.py#L172)

As a result port labels are never found When `sepolicy` command is not available, and `selinux.port` always emits a `semanage port -m` command even when the port is already correctly labeled.

yacoob:
I *think* this is the right direction to converge on, as there's probably code out there that uses pyinfra and expects ints. Or not, given this has flown 4y under the radar :D 

I've converted the test fixtures to yaml, as you can't have int keys in json. Once converted, the bug was visible, and those test cases started to fail. 

Fixes #1644.


Fizzadar (approved):
Nice, thank you @yacoob!

PR #1658, opened 2026-04-08
api,ops: config.EXECUTION_ONLY and op._execution_only flags

### LINKED ISSUE #1624: FR: config.SKIP_IDEMPOTENCY flag
By using Pyinfra as an API, one might use it in pipelines to calculate state drifts with facts to have less syscalls for one specific edge case, or to expand the operation (for instance, applying an sudo configuration file per user with a flag). By calculating these drifts by hand and utilizing Pyinfra's operations, the drift gets calculated again internally, which results in an higher execution overhead.

I'd like an `config.SKIP_IDEMPOTENCY` flag paired with an _skip_idempotency in order to... well, skip idempotency
if the user states "I want these packages installed", then Pyinfra should simply run the command with the selected packages, erroring out if the command fails. This would clearly be a more advanced configuration, so I don't think it would be all that bad

This is meant to not rely on server.shell and use Pyinfra's builtin abstractions, which are already made to be safe and easy to use, other than being easier for other tools to use the API.

The CLI could intercept this and pop out a warning about this as well!

Dexmachi:
introduces a flag to direct execution of commands without checks for idempotency, allowing users to rely on Pyinfra's already good abstractions.

# Why
Because, in the API usecase, users often use Facts to implement custom idempotency checks (for instance, to introduce two-way declarativity), and in order to reuse Pyinfra's already existing abstractions without the CLI, there is a use for  a bare execution with no checks. This effectively reduces the necessity of a `server.shell` call for basic operations in API usage by a lot in batch-collecting fact edge cases.

This allows Pyinfra operations to be reused as execution primitives, without duplicating logic outside of the framework.

# How
Just added a new "operation execution path" to Pyinfra. Now there is the Optional direct_execution flag inside of the \@operation() decorator, which only gets used when config.EXECUTION_ONLY is set to true.

It works like so: If EXECUTION_ONLY is enabled, the system attempts to use the operation's `direct_execution` callable. If not provided, it falls back to the standard execution path (including fact checks).

Note: This is a feature only PR in order to make it more atomic, this does not introduce any operation changes and all features in it are completely opt-in.

This closes #1624 

- [N/A] Pull request includes tests for any new/updated operations/facts


Fizzadar (changes requested):
Sorry for the complete lack of review here @Dexmachi! Had a proper look now, some notes below, but:

Bigger question before detail though: is `direct_execution` the shape we want here? It means every operation has to carry a second parallel implementation with no shared code path.

I'm not quite following the use-case/why section here, what situation would the current fact+command setup not suffice? One option would be to make all facts return defaults which might achieve roughly similar without duplicate paths?

Or, alternatively, a separate package of "pyinfra ops without idempotency checks", would still mean duplicated code but I think that would be best as a third party package.




- `_execution_only` never reaches the check. It's in `ExecutionArguments` so `EXECUTION_KWARG_KEYS` picks it up, and `ensure_shared_op_meta` pops it out of `global_arguments` before `command_generator` runs - so we always fall through to `config.EXECUTION_ONLY`. Wants to be `op_meta.global_arguments["_execution_only"]`, same as `_run_once` does.
- It also wants to be a meta argument rather than an execution one. Execution args have to take the same value on every host, and `pop_global_arguments` reads `host.data` too, so anyone setting `_execution_only` in host data gets a hard `OperationValueError`.
- Nothing in the tree actually supplies `direct_execution`, so `config.EXECUTION_ONLY = True` doesn't skip anything today - we log the warning and then call `func()` and gather facts exactly as normal (twice over, as the generator gets iterated for change detection and again for execution). The new docs say operations "will be executed without checking their prior state", which isn't what happens yet.

Issue #1659, opened 2026-04-09
Support for module streams in pyinfra.operations.dnf


yagarea:
DNF module streams aren't supported by `pyinfra.operations.dnf`

Current work around:
```py
server.shell(                                                                                                                                             
    name="Enable PostgreSQL 16 module stream",                                                                                                               
    commands=["dnf module enable -y postgresql:16"],
)
```

wowi42:
Implemented. DNF module streams are now supported natively.

`dnf.module` was added in #1660 (commit `34cd675c`, "feat(operations.dnf): add support for DNF modules"). `pyinfra.operations.dnf.module(module, stream=..., enabled=...)` runs `dnf module enable`/`disable` for a `module:stream`, replacing the previous `server.shell` workaround.

Closing as resolved.

PR #1660, opened 2026-04-09
feat(operations.dnf): add support for DNF modules

yagarea:
Solves #1659 

- [x] Tests pass (see `scripts/dev-test.sh`) (1552 passed, 15 deselected, 3 warnings in 7.93s)


yagarea:
```
Warning: "nstalled" should be "installed".
error: `nstalled` should be `installed`
  --> ./tests/facts/dnf.DnfEnabledModules/multiple_modules.json:9:54
  |
9 |         "Hint: [d]efault, [e]nabled, [x]disabled, [i]nstalled"
  |                                                      ^^^^^^^^
```
What can I do about these spell checks failing? This is real dnf output

wowi42:
Drive-by review, hope it's useful.

### Must-fix

**Shell-injection in `src/pyinfra/operations/dnf.py`**

```python
yield f"dnf module enable -y {module}:{stream}"
```

`module` and `stream` are user-supplied and end up unquoted in a shell string. This regresses the recent `bfb2fafe` ("security: quote untrusted values in command construction"). Suggested:

```python
from pyinfra.api import QuoteString, StringCommand
yield StringCommand("dnf module enable -y", QuoteString(f"{module}:{stream}"))
```

### Should-address

**Scope**: op is named `module` but only enables. Either rename to `module_enable`, or add `present: bool = True` and handle `dnf module disable/reset` symmetrically. As-is, `dnf.module(..., present=False)`-style expectations will be surprising.

**Stream switching**: `dnf module enable` won't switch streams when packages from the old stream are installed, it will error. Worth a docstring note, or auto-issuing a `reset` first.

**Docs checkbox unchecked**: docstrings are present and `scripts/generate_operations_docs.py` picks them up, so probably just an oversight.

### Nice-to-have

**Parse robustness** (`DnfEnabledModules.process`): the `\"[e]\" not in line` filter is fine, but the `startswith(\"Hint:\")` guard is unreachable since the hint line contains no `[e]`. Remove or replace with a real edge case.

yagarea:
@wowi42 Thank you for your review. I have tried to implement your suggestions. I hope now is everything you have listed fixed.

src/pyinfra/operations/dnf.py
```diff
@@ -211,3 +212,68 @@ def packages(
...
+
+@operation()
+def module(module: str, stream: str | None = None, state: str = "enabled"):
+    """
+    Manage a dnf module (application stream).
+
+    + module: name of the dnf module
+    + stream: name of the stream, required when ``state="enabled"``, ignored \
+              when ``state="disabled"``
+    + state: one of ``enabled`` (default) or ``disabled``
```
Fizzadar: Are there any other possible states? If not we should make this a `enabled: bool` flag.

Fizzadar (changes requested):
One nit otherwise this looks good to me @yagarea!

src/pyinfra/operations/dnf.py
yagarea: No, they can not. I will fix it asap.

yagarea:
@Fizzadar Done

Fizzadar (approved):
LGTM, ty @yagarea!

PR #1662, opened 2026-04-13
feat(cli): add --json flag for pure JSON stdout output

### LINKED ISSUE #509: Add `--json` output flag
**Is your feature request related to a problem? Please describe.**
As it stands `pyinfra` is very much a CLI tool executed by users. There's the programmatic API as well for Python folk but there's also no reason the CLI can't be used within scripts and so on. Currently output is written (correctly) to `stderr` but no machine readable output is emitted to `stdout` - this flag would push out JSON during execution, such that the whole output would be a valid JSON object.

This would make it possible to integrate `pyinfra` into any number of systems without writing any Python code.

**Describe the solution you'd like**
Add `--json` CLI flag that makes `pyinfra` emit machine readable JSON output to `stdout` during execution. Possibly in tandem with a flag to hide most/all `stderr` user output.

wowi42:
Emit structured JSON on stdout for debug-inventory, fact, --debug-operations, --dry and deploy runs so the CLI output can be consumed by downstream tools instead of scraping the pretty printer. Forces non-interactive mode and disables the progress spinner to keep stdout unpolluted.

closes #509 


src/pyinfra_cli/cli.py
```diff
@@ -320,8 +334,14 @@ def _main(
...
     debug_facts: bool,
     debug_operations: bool,
+    json_output: bool = False,
     support: bool = False,
 ):
+    # In JSON mode keep the spinner quiet and force non-interactive behaviour
+    # so stdout stays pure JSON and prompts never block the pipe.
+    if json_output:
+        environ.setdefault("PYINFRA_PROGRESS", "off")
+        yes = True
```
DonDebonair: The fact that we switch to non-interactive behaviour should be documented in the docs. Of course, JSON output is pretty useless in interactive mode, but still, it's a change of behaviour.

src/pyinfra_cli/prints.py
```diff
@@ -52,6 +52,132 @@ def jsonify(data, *args, **kwargs):
...
 
+def print_json(payload) -> None:
+    click.echo(jsonify(payload, default=json_encode))
+
+
+def _host_to_dict(host: Host) -> Dict:
+    return {
+        "name": host.name,
+        "groups": list(host.groups),
+        "data": host.data,
```
DonDebonair, line 74: Host data can contain any Python object, which could be unserializable. I'd recommend documenting that.

src/pyinfra_cli/prints.py
```diff
@@ -52,6 +52,132 @@ def jsonify(data, *args, **kwargs):
...
+    }
+
+
+def print_inventory_json(state: "State") -> None:
```
DonDebonair: Quotes are only needed when doing either a self-reference inside a class, or when referencing a class that is defined in the same file, but below the current code. And then only in older versions of Python.


```suggestion
def print_inventory_json(state: State) -> None:
```

src/pyinfra_cli/prints.py
```diff
@@ -52,6 +52,132 @@ def jsonify(data, *args, **kwargs):
...
+    print_json([_host_to_dict(host) for host in state.inventory])
+
+
+def print_facts_json(fact_data: Dict) -> None:
```
DonDebonair: Lowercase type names are preferred.

```suggestion
def print_facts_json(fact_data: dict) -> None:
```

(I realize that this happens all over the code base and is not yet caught be the linter. To be fixed at a later moment)

DonDebonair (changes requested):
Great PR! Left some comments. Basically 2 themes:

1. Recommend adding documentation for the new flag, incl. some caveats (non-interactive behaviour, potentials problems with unserializable host data)
2. Fix type hints to modern Python standards

wowi42:
Thanks for the review @DonDebonair! Addressed everything in `1d43e00a`:

**Type hints** (`src/pyinfra_cli/prints.py`)
- Dropped the quotes around `state: State` on every new helper — `from __future__ import annotations` is already in place so they were redundant.
- Switched the new helpers to PEP 585 lowercase built-ins (`dict`, `list`) for signatures and local annotations. Left the pre-existing `Dict`/`List`/`Tuple` usages alone as they're out of scope (per your note about tackling the codebase-wide sweep later).

**Documentation** (`docs/cli.md`)
- Added a dedicated "JSON output" section covering `--json` with examples (`debug-inventory`, `fact`, `--dry`, deploy).
- "Non-interactive behaviour" sub-section calls out that `--yes` is implied and the progress spinner is disabled (`PYINFRA_PROGRESS=off`), and warns that interactive prompts from deploy code will still block.
- "Caveats" sub-section notes that inventory `data` is serialised as-is, so non-JSON-serialisable values will fail the command, plus a pointer that stderr still carries logs.

Fizzadar (approved):
LGTM!

Conflicts 🙃

src/pyinfra_cli/cli.py
Fizzadar: I'm not sure this is right - it should be possible to diff, but not change, a target host. I think we should just JSON print the proposed changes.

Fizzadar (changes requested):
Spotted a thing!

Fizzadar:
One final question which is probably a future PR - should we consider exporting logs as JSON too newline delimited? Maybe `--json-events` or something.

wowi42:
@Fizzadar all good for me

Fizzadar (approved):
Awesome, LGTM!

PR #1664, opened 2026-04-14
security: quote untrusted values in command construction across connectors, operations, and util

wowi42:
# Summary

Hardening of every place pyinfra builds a shell command by interpolating inventory- or caller-supplied strings without passing them through `QuoteString` / `shlex.quote`. `api.command.StringCommand` only shell-escapes bits wrapped in `QuoteString`; raw string bits are pasted verbatim. Several call sites relied on raw bits, which made them injectable when the surrounding value was not strictly trusted (dynamic inventories, data loaded from external systems, etc.).

## Fixed injection surfaces

**`connectors/util.py` -- `make_unix_command`**
- `_env` was built with `'"{0}={1}"'.format(key, value)`; double-quoted shell context still interprets `"`, `$`, `` ` ``. Now each `key=value` is wrapped in `QuoteString` so values with metacharacters become a single literal token.
- `_chdir`, `_sudo_user`, `_doas_user`, `_dzdo_user`, `_su_user` were pasted as raw bits -- now wrapped in `QuoteString`.
- `_su_shell` used unquoted backtick command substitution (``-s `which {shell}` ``) which executed the user value as a subshell. Replaced with `$(command -v 'shell')` via `StringCommand` + `QuoteString` so injection attempts fail safely inside `command -v`.
- `_sudo_askpass_path` / `_su_askpass_path` were pre-quoted once when stored in `connector_data` and then concatenated back unquoted. Switched to storing the raw path and quoting it at each use site (including `remove_any_sudo_askpass_file`, which previously used `"rm -f {0}".format(path)`).

**`operations/util/files.py` -- `chown()`**
- `user_group` (`"user:group"`) was passed as a raw bit. Wrapped in `QuoteString` and added a `None` assertion so callers that pass neither user nor group fail loudly instead of emitting `chown None <target>`.

**`connectors/docker.py`, `connectors/dockerssh.py`, `connectors/chroot.py`**
- Docker/Podman container IDs and image names were interpolated directly into `f"{self.docker_cmd} container inspect {container_id}"` style `local.shell()` calls -- this is RCE on the *controller* machine, not the target, because `local.shell()` executes via `run_local_process` with `shell=True`. Now `shlex.quote`'d / wrapped in `QuoteString` throughout `connect`, `disconnect`, `run_shell_command`, `put_file`, and `get_file`.
- Same pattern fixed for `dockerssh` (`ssh.run_shell_command`) and `chroot` (container `chroot` directory and `cp` target paths).

**`operations/selinux.py` -- `selinux.boolean()`**
- `yield StringCommand(""setsebool"", f""{persist}{bool_name}"", value_str)` concatenated the `-P` flag with the boolean name into a single raw bit. Split the flag out into its own argument and wrapped `bool_name` in `QuoteString`.

## Tests

- `tests/test_connectors/test_util.py` -- regression/injection-attempt tests for `_doas_user`, `_dzdo_user`, `_sudo_user`, `_su_user`, `_su_shell`, `_chdir`, `_env`, the `SUDO_ASKPASS=` env assignment, and `remove_any_sudo_askpass_file`, plus a mixed scenario. Expected outputs for the existing `test_su_shell_command`, `test_command_env`, and `test_mixed_command` were updated for the new, stricter quoting.
- `tests/operations/files.file/add_user_group_with_metachars.json` :  fixture proving `chown` quotes `user:group` when it contains `;` and `$()`.
- `tests/operations/selinux.boolean/change_bool_name_with_metachars.json` : fixture proving `setsebool -P` quotes a bool name containing backticks.
- All existing chown/docker/chroot fixtures remained valid because their test inputs use characters that are identity under `shlex.quote`.

## Threat model note

Most of these are only exploitable when the injected argument comes from untrusted inventory / dynamic data (e.g. a `@terraform` output used as a host name, a fact result fed into a subsequent operation). For fully hand-authored static inventory the risk is lower -- but the fixes are purely defensive and the audit trail through `StringCommand` is now consistent: **every bit that reaches the shell is either a literal pyinfra-owned string or `QuoteString`-wrapped.**

Related second-pass audit (not included in this PR) also found injection paths in several facts (`facts/git.py`, `facts/iptables.py`, `facts/crontab.py`, `facts/selinux.py`, `facts/docker.py`, `facts/npm.py`, `facts/postgres.py`, `facts/mysql.py`, `facts/snap.py`, `facts/flatpak.py`, `facts/openrc.py`, `facts/systemd.py`, `facts/runit.py`, `facts/gpg.py`, `facts/server.py`) where a fact's `command()` method returns a plain string with raw-formatted parameters. Those are best fixed together as a follow-up by either wrapping each parameter with `shlex.quote` at the call site or converting facts to return `StringCommand` , happy to do that in a separate PR @Fizzadar .

### DIFF
--- a/src/pyinfra/operations/util/files.py
+++ b/src/pyinfra/operations/util/files.py
@@ -6,7 +6,7 @@ from datetime import datetime, timezone
 from enum import Enum
 from typing import Callable, Generator
 
-from pyinfra.api import QuoteString, StringCommand
+from pyinfra.api import OperationError, QuoteString, StringCommand
 from pyinfra.api.output import format_text
 
 
@@ -158,7 +158,8 @@ def chown(
         command = "chgrp"
         user_group = group
 
-    assert user_group is not None, "chown() requires at least one of user or group"
+    if user_group is None:
+        raise OperationError("chown() requires at least one of user or group")
 
     args = [command]
     if recursive:
--- a/src/pyinfra/operations/util/files.py
+++ b/src/pyinfra/operations/util/files.py
@@ -146,7 +146,7 @@ def chown(
     dereference=True,
 ) -> StringCommand:
     command = "chown"
-    user_group = None
+    user_group: str | None = None
 
     if user and group:
         user_group = "{0}:{1}".format(user, group)
@@ -158,6 +158,8 @@ def chown(
         command = "chgrp"
         user_group = group
 
+    assert user_group is not None, "chown() requires at least one of user or group"
+
     args = [command]
     if recursive:
         args.append("-R")
@@ -165,7 +167,7 @@ def chown(
     if not dereference:
         args.append("-h")
 
-    return StringCommand(" ".join(args), user_group, QuoteString(target))
+    return StringCommand(" ".join(args), QuoteString(user_group), QuoteString(target))
 
 
 # like the touch command, but only supports setting one field at a time, and expects any
--- a/src/pyinfra/connectors/chroot.py
+++ b/src/pyinfra/connectors/chroot.py
@@ -1,4 +1,5 @@
 import os
+import shlex
 from tempfile import mkstemp
 from typing import TYPE_CHECKING, Optional
 
@@ -64,7 +65,7 @@ class ChrootConnector(BaseConnector):
         try:
             with progress_spinner({"chroot run"}):
                 local.shell(
-                    "chroot {0} ls".format(chroot_directory),
+                    "chroot {0} ls".format(shlex.quote(chroot_directory)),
                     splitlines=True,
                 )
         except PyinfraError as e:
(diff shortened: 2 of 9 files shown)


src/pyinfra/operations/util/files.py
```diff
@@ -158,14 +158,16 @@ def chown(
         command = "chgrp"
         user_group = group
 
+    assert user_group is not None, "chown() requires at least one of user or group"
```
DonDebonair: `assert` shouldn't be used in non-test code. Asserts can be disabled easily which would cause errors to be silently ignored.

In this case, it's probably better to thrown an exception.

I realize that this mistake is made in other places in the code as well. We should fix that at some point

DonDebonair:
Awesome PR! Left 1 comment which should be addressed. Otherwise looking great.

PR #1666, opened 2026-04-14
fix(connectors.util): honour _temp_dir in askpass helpers

### LINKED ISSUE #1623: Per-host TEMP_DIR override
Hello!
First of all, thank you for this amazing project! Thank you for taking some of your time to read this issue :) !

## Summary

Allow defining a per-host temporary directory that overrides `config.TEMP_DIR`, enabling safe and ergonomic handling of environments where `/tmp` is not usable (e.g. `noexec`, restricted mounts).

## Background

On some systems, such as TrueNAS, the default `/tmp` directory may be:

* mounted with restrictive options (e.g. `noexec`)
* otherwise unsuitable for executing temporary scripts

This breaks internal mechanisms in pyinfra that rely on temporary files, notably `SUDO_ASKPASS`.

### Global override via `config.TEMP_DIR`

```python
from pyinfra.context import config

config.TEMP_DIR = "/run/user/950"
```

**Limitations:**

- Applies to all hosts
- Not suitable for heterogeneous environments
- AFAIK, not thread-safe under parallel execution (if someone was to set it multiple times based on something stored in `host.data`)

### Per-operation `_temp_dir` (introduced in v3.7)

```python
server.shell(..., _temp_dir=host.data.temp_dir)
```

**Limitations:**

- Must be repeated on every operation
- Does not affect internal operations (e.g. askpass helpers)
- Poor ergonomics and maintainability

## Proposed Solution

Introduce a per-host `temp_dir` attribute, defined in inventory:

```python
app_servers = [
    ("truenas.tld", {"ssh_user": "truenas-admin", "temp_dir": "/run/user/950"}),
    ("otherserver.tld", {"install_postgres": True}),
]
```

This value would be automatically used by pyinfra when resolving temporary directories.

### Resolution Order (precedence)

1. `_temp_dir` (operation-level override)
2. `host.data.temp_dir` (new)
3. `config.TEMP_DIR`
4. `config.DEFAULT_TEMP_DIR`

> I don't remember where the host environment variable `TMPDIR` fits in this...

### Expected Behavior

* All operations (including internal ones like `SUDO_ASKPASS`) respect the resolved temp directory
* No need to modify individual operations
* Safe under parallel execution (no shared global mutation)

### Benefits

* Enables correct behavior on systems with restricted `/tmp`
* Eliminates need for global state mutation
* Improves ergonomics and reduces boilerplate
* Aligns with existing per-host configuration patterns in pyinfra

---

Thank you!

wowi42:
Closes #1623.

## Problem

`server.reboot` and other ops respected `_temp_dir` (operation kwarg, or the per-host default cascaded from `host.data._temp_dir` via the global-argument system) for file placement, but the internal `SUDO_ASKPASS` / `SU_ASKPASS` helper script was still dropped into `config.TEMP_DIR` (usually `/tmp`). On hosts where `/tmp` is unusable (`noexec`, restricted tmpfs, TrueNAS, etc.) the askpass helper would fail even when the user had configured a working temp directory.

## Fix

- `_ensure_askpass_set_for_host` now takes a `temp_dir` override and uses it for the mkstemp template.
- `make_unix_command_for_host` threads the operation-level `_temp_dir` through to the askpass helpers.
- The resolved `temp_dir` is stored alongside the cached path (`<key>_temp_dir`); if a later call resolves a different `temp_dir` the cache is invalidated so the script gets regenerated under the correct directory. Callers that read `host.connector_data["sudo_askpass_path"]` directly don't need to change.

Per-host defaults don't need any new code path: setting `host.data._temp_dir` already cascades as the default for the `_temp_dir` global argument on every op, which is what gets passed down.

## Test plan

- [x] Askpass path for default, `config.TEMP_DIR`, and op-level `_temp_dir`.
- [x] Cache invalidation when `temp_dir` changes between calls.
- [x] Full path through `make_unix_command_for_host` with `_sudo=True` + `_temp_dir`.
- [x] `uv run pytest` (1613 passed), ruff / ruff format / mypy / `scripts/lint_arguments_sync.py` clean.

---

- [x] Pull request includes documentation for any new/updated operations/facts (behaviour fix, no public API / docs change)

### DIFF
--- a/src/pyinfra/connectors/util.py
+++ b/src/pyinfra/connectors/util.py
@@ -229,26 +229,43 @@ def write_stdin(stdin, buffer):
     buffer.close()
 
 
+ASKPASS_PATH_KEYS = ("sudo_askpass_path", "su_askpass_path")
+
+
+def _iter_askpass_cache_keys(host) -> list[str]:
+    # Cache keys are either the bare base ("sudo_askpass_path") or the base
+    # joined with the resolved temp_dir ("sudo_askpass_path__/tmp"). Match
+    # both so cleanup covers every askpass file ever generated for the host.
+    return [
+        cache_key
+        for cache_key in list(host.connector_data.keys())
+        for base in ASKPASS_PATH_KEYS
+        if cache_key == base or cache_key.startswith(base + "__")
+    ]
+
+
 def remove_any_sudo_askpass_file(host) -> None:
     # Best-effort cleanup: this is called from host.disconnect(), and the
     # connection may already be broken (e.g. after `server.reboot`). Swallow
     # any errors from the remote ``rm`` and still clear the local state so a
     # reconnect will regenerate a fresh askpass file.
-    sudo_askpass_path = host.connector_data.get("sudo_askpass_path")
-    if sudo_askpass_path:
+    for cache_key in _iter_askpass_cache_keys(host):
+        path = host.connector_data.get(cache_key)
+        if not path:
+            continue
         try:
-            host.run_shell_command(StringCommand("rm", "-f", QuoteString(sudo_askpass_path)))
+            host.run_shell_command(StringCommand("rm", "-f", QuoteString(path)))
         except Exception as e:
-            logger.debug("Could not remove sudo askpass file %s: %s", sudo_askpass_path, e)
-        host.connector_data["sudo_askpass_path"] = None
+            logger.debug("Could not remove askpass file %s: %s", path, e)
+        host.connector_data[cache_key] = None
 
-    su_askpass_path = host.connector_data.get("su_askpass_path")
-    if su_askpass_path:
-        try:
-            host.run_shell_command(StringCommand("rm", "-f", QuoteString(su_askpass_path)))
-        except Exception as e:
-            logger.debug("Could not remove su askpass file %s: %s", su_askpass_path, e)
-        host.connector_data["su_askpass_path"] = None
+
+def clear_askpass_cache(host) -> None:
+    # Drop every cached askpass path without touching the remote, used after
+    # ``server.reboot`` where the previous connection (and therefore any
+    # askpass scripts under its temp dir) is gone.
+    for cache_key in _iter_askpass_cache_keys(host):
+        host.connector_data[cache_key] = None
 
 
 @memoize
@@ -277,13 +294,13 @@ def extract_control_arguments(arguments: ConnectorArguments) -> ConnectorArgumen
     return control_arguments
 
 
-def _ensure_sudo_askpass_set_for_host(host: Host, temp_dir: str | None = None):
+def _ensure_sudo_askpass_set_for_host(host: Host, temp_dir: str | None = None) -> str:
     return _ensure_askpass_set_for_host(
         host, "sudo_askpass_path", SUDO_ASKPASS_ENV_VAR, temp_dir=temp_dir
     )
 
 
-def _ensure_su_askpass_set_for_host(host: Host, temp_dir: str | None = None):
+def _ensure_su_askpass_set_for_host(host: Host, temp_dir: str | None = None) -> str:
     return _ensure_askpass_set_for_host(
         host, "su_askpass_path", SU_ASKPASS_ENV_VAR, temp_dir=temp_dir
     )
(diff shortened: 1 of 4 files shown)


src/pyinfra/api/host.py
```diff
@@ -312,9 +312,28 @@ def deploy(
...
+        Resolve a per-host temporary directory from ``host.data``. A host data
+        entry of ``temp_dir`` (or ``_temp_dir`` as an alias, matching the
+        operation-level global argument name) wins over any global config
+        value so heterogeneous inventories can pin a working path for hosts
+        where the default ``/tmp`` is unusable (``noexec`` mounts, etc.).
+        """
+        value = self.data.get("temp_dir")
+        if value:
+            return value
+        value = self.data.get("_temp_dir")
```
Fizzadar, line 326: This shouldn't be required - we already apply global arguments in host data (so `_temp_dir` in host data will apply to all operations as an argument): https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/api/arguments.py#L405

src/pyinfra/connectors/util.py
```diff
@@ -270,18 +270,30 @@ def extract_control_arguments(arguments: "ConnectorArguments") -> "ConnectorArgu
     return control_arguments
 
 
-def _ensure_sudo_askpass_set_for_host(host: "Host"):
-    return _ensure_askpass_set_for_host(host, "sudo_askpass_path", SUDO_ASKPASS_ENV_VAR)
+def _ensure_sudo_askpass_set_for_host(host: "Host", temp_dir: Optional[str] = None):
+    return _ensure_askpass_set_for_host(
+        host, "sudo_askpass_path", SUDO_ASKPASS_ENV_VAR, temp_dir=temp_dir
```
Fizzadar, line 275: We should add the path to the key here or we'll incorrectly cache the wrong askpass path if the value is changed (very unlikely, but worth guarding against).

```suggestion
        host, f"sudo_askpass_path:{temp_dir}", SUDO_ASKPASS_ENV_VAR, temp_dir=temp_dir
```

Fizzadar (changes requested):
Looks good, two things to fix.

wowi42:
Thanks for the review, both addressed and rebased on `3.x`.

Dropped the `host.py` bits entirely, you were right, `_temp_dir` on `host.data` already cascades via global arguments. The askpass helpers just pick it up from the op-level kwarg now.

For the cache key, went with storing the resolved `temp_dir` next to the path and invalidating when it changes, so readers like `make_unix_command_for_host` keep working unchanged. New test covers the invalidation.

PR description is now a bit stale, I'll clean it up next.

DonDebonair:
~@wowi42 there is one open comment you didn't address yet :)~

Maybe @Fizzadar should comment on what he wants with the cache key

Fizzadar (changes requested):
This seems overly complicated, could we not just add the tmp path to the key at call time if set? so we end up with keys like:

- `sudo_askpass_path__/tmp`
- `sudo_askpass_path__/x`

and so no? No need for the extra field or checks then (I think!)

wowi42:
Good call, that is much cleaner. Pushed efc81583 with the simpler scheme:

- `_ensure_askpass_set_for_host` now caches under `f"{key}__{effective_temp_dir}"` (e.g. `sudo_askpass_path__/tmp`, `sudo_askpass_path__/dev/shm`) and returns the resolved path. The separate tracker field and the invalidation block are gone, switching `_temp_dir` between calls just misses the cache and creates a fresh script under the right directory.
- `make_unix_command_for_host` uses the returned path directly instead of round-tripping through `connector_data`.
- `remove_any_sudo_askpass_file` now walks every `{sudo,su}_askpass_path*` variant so `host.disconnect()` cleans up scripts under any temp dir the host saw during the run. Added `clear_askpass_cache` for `server.reboot`, which clears the cache without touching the (gone) remote.

Tests updated to seed the temp_dir-suffixed keys and cover both the multi-variant cleanup path and same-temp_dir cache reuse.

Fizzadar:
@wowi42 conflicts after style merge unfortunately, think this is good to go after!

Fizzadar (approved):

etiennecollin:
Hey! I'm the author of #1623.
Just want to thank @wowi42 and @Fizzadar for working on this so quickly 😊
Have a great day and thank you for your work!

wowi42:
Force-pushed ff87e088 after rebasing onto current 3.x. The branch was based on v3.8.0 and the existing "Update branch" merge predated #1707 (server.user shadow precheck) and #1721 (modern syntax linting), so the PR was conflicting again.

Conflicts resolved in `src/pyinfra/connectors/util.py` only: kept this PR's `temp_dir` parameter and per-(host, temp_dir) cache key, applied the #1721 modernizations on top (unquoted `Host` forward refs, `str | None` over `Optional[str]`, f-strings over `"{0}".format()`).

@Fizzadar your earlier suggestion to encode the path in the cache key is already in place. The current code is `cache_key = f"{key}__{effective_temp_dir}"`, so each (host, temp_dir) pair gets its own slot.

Tests + lint clean locally (`scripts/dev-test.sh`, `scripts/dev-lint.sh`).

Fizzadar (approved):
Nice! LGTM

PR #1667, opened 2026-04-14
ci: upgrade default Python to 3.14

wowi42:
Update lint, type-check and unit-test matrix to use Python 3.14 as the default version. Add 3.13 as an explicit include entry so it remains tested on ubuntu-24.04, alongside the already-present 3.12, 3.11 and 3.10 matrix entries. Flip the codecov upload condition to the new default. Leave end-to-end-test and docs-build on 3.13 for now, matches the narrow scope of the original change.

Also add the Python 3.14 trove classifier to pyproject.toml and drop ``concurrency = ["gevent"]`` from ``[tool.coverage.run]``; gevent's coverage concurrency integration is incompatible with 3.14 and the pytest-cov run does not need it.

This is PR #1599 re-applied on 3.x without the uv.lock bump.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v5
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.13"
+          python-version: "3.14"
       - name: Install uv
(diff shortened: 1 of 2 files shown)


DonDebonair:
Looks good to me. Ready to merge @Fizzadar

PR #1669, opened 2026-04-17
feat(facts.server): add EtcHosts, Last, Lastb, LoadAverage facts + server.etc_hosts op

wowi42:
Adds four `server` facts inspired by osquery's `etc_hosts`, `last` and `load_average` tables, plus an `etc_hosts` operation that builds on the new fact for idempotency.

### `server.EtcHosts` (fact)

Parses `/etc/hosts` (or a custom `path`) into a mapping of IP to the list of hostnames declared on each line. Comments and empty lines are ignored; duplicate IPs are merged in file order.

```python
{
    "127.0.0.1": ["localhost", "localhost.localdomain"],
    "::1": ["localhost"],
}
```

### `server.etc_hosts` (operation)

Idempotent add / update / remove of entries in `/etc/hosts` (or a custom `path`) keyed by IP. The idempotency decision is made against the parsed `EtcHosts` fact, so a correctly-configured host is a single fact lookup + noop. When a mutation is needed, the operation falls through to `files.FileContents` + `files.put` so comments and unrelated entries survive the rewrite.

```python
server.etc_hosts(
    name="Register db.internal",
    ip="192.168.1.10",
    hostnames=["db.internal", "db"],
)

server.etc_hosts(
    name="Drop the legacy alias",
    ip="192.168.1.10",
    hostnames="db",
    present=False,
)

server.etc_hosts(
    name="Remove 10.0.0.1 entirely",
    ip="10.0.0.1",
    present=False,
)
```

Behavior:

- `present=True` + `hostnames`: the IP line becomes exactly `<ip> <hostnames...>`, adding or replacing it.
- `present=False`, no `hostnames`: the whole line for `ip` is removed.
- `present=False` + `hostnames`: only those names are dropped from the IP's line; the line goes away if nothing remains.

### `server.Last` / `server.Lastb` (facts)

Return login records (`last`) and failed-login records (`lastb`) as a list of `{user, tty, host, time}` dicts. `Lastb` subclasses `Last` to share the parser; only the command and `requires_command` differ. The `time` field is kept as the raw trailing string so callers can re-parse the date format that matches their system.

```python
[
    {"user": "alice", "tty": "pts/0", "host": "192.168.1.5",
     "time": "Thu Apr 17 14:00:00 2026 still logged in"},
    {"user": "reboot", "tty": "system boot", "host": "6.19.10-arch1-1",
     "time": "Thu Apr 17 11:00:00 2026 - Thu Apr 17 12:00:00 2026 (01:00)"},
]
```

Commands (chosen after live-testing on Alpine, Arch, Ubuntu, FreeBSD):

- `Last`: `last -Fwi 2>/dev/null || last -w 2>/dev/null || last 2>/dev/null || true`
- `Lastb`: `lastb -Fwi 2>/dev/null || lastb 2>/dev/null || true`

Flag portability:

| OS | `last` | `lastb` | `-Fwi` | `-T` |
|---|---|---|---|---|
| Alpine (busybox 1.37) | flagless only | not installed | no | no |
| Arch (util-linux 2.41) | yes | yes | yes | yes |
| Ubuntu (util-linux 2.39) | yes | yes | yes | no (new in 2.41) |
| FreeBSD | `-w` only | not installed | no | no |

`-T` was dropped because it only exists in util-linux 2.41+. Fixtures cover the util-linux, FreeBSD and busybox-empty cases.

### `server.LoadAverage` (fact)

Reads `/proc/loadavg` on Linux with an `uptime` fallback for systems without procfs (e.g. FreeBSD, which also uses "load averages:" plural). Returns a dict keyed by window in minutes.

```python
{"1": 0.12, "5": 0.21, "15": 0.22}
```

Command: `cat /proc/loadavg 2>/dev/null || uptime`.

### Tests

- Facts: `tests/facts/server.EtcHosts/`, `server.Last/` (util-linux, FreeBSD, busybox-empty), `server.Lastb/`, `server.LoadAverage/` (proc, FreeBSD uptime, Linux uptime).
- Operation: `tests/operations/server.etc_hosts/` covers add, update, noop-already-present, create-file, full removal, partial hostname removal, absent-IP noop, already-missing hostname noop and missing-file absent noop.


src/pyinfra/facts/server.py
```diff
@@ -1340,3 +1340,211 @@ def process(self, output: Iterable[str]) -> dict[int, ProcessDict]:
...
+            "::1": ["localhost"],
+            "192.168.1.10": ["db.internal"],
+        }
+    """
+
+    default = dict
+
+    @override
+    def command(self, path: str = "/etc/hosts") -> str:
+        return f"cat {path} 2>/dev/null || true"
```
Fizzadar: We should `QuoteString` here to avoid shell escapes.

Fizzadar (changes requested):
One nit on untrusted user input

src/pyinfra/facts/server.py
wowi42: yep...

wowi42: Done. `EtcHosts.command` now returns `make_formatted_string_command("cat {0} 2>/dev/null || true", QuoteString(path))`, same pattern as `files.Sha1File`. The other new facts take no user input. Fact and op fixtures still pass.

Fizzadar (approved):
LGTM!

PR #1670, opened 2026-04-17
facts.server: add AuthorizedKeys, make user_authorized_keys idempotent

wowi42:
Adds a `server.AuthorizedKeys` fact and refactors the `server.user_authorized_keys` operation to drive idempotency from that single fact lookup instead of one `FindInFile` round-trip per key.

### `server.AuthorizedKeys` (fact)

Reads a user's `~/.ssh/authorized_keys` (or a custom `path`) and returns the list of key strings with comments and blank lines stripped; tilde expansion resolves the home directory so there is no extra fact call for `Home`. Order is preserved.

```python
host.get_fact(AuthorizedKeys, user="alice")
# ["ssh-ed25519 AAAAC3Nz... user@host",
#  "ssh-rsa AAAAB3Nz... other@host"]

host.get_fact(AuthorizedKeys, user="alice", path="/etc/ssh/keys/alice")
```

Command: `cat ~<user>/.ssh/authorized_keys 2>/dev/null || true`, or `cat <path> ...` when `path` is given. Missing files resolve to an empty list. Verified live against Arch (`util-linux`) and FreeBSD.

### `server.user_authorized_keys` (refactor)

Previously: one `files.line` per requested key, which internally issued one `FindInFile` fact call per key. For a user with many keys, that dominated runtime.

Now: a single `AuthorizedKeys` call feeds the decision.

- `delete_keys=True`
  - **match**: file already lists exactly the requested keys in order, so the operation only ensures owner/mode via `files.file` (noop for content).
  - **mismatch**: rewrites through `files.put` exactly as before.
- Default mode (`delete_keys=False`)
  - `files.file` ensures the `authorized_keys` file with the right owner/mode.
  - `files.line` is only emitted for keys that the fact says are missing; keys already present are a noop.

### Tests

- Facts: `tests/facts/server.AuthorizedKeys/` covers the default-path tilde-expansion case and a custom-path empty-file case.
- Operation: `tests/operations/server.user_authorized_keys/` covers all-present noop, append-missing-only, `delete_keys=True` match noop, and `delete_keys=True` mismatch rewrite.
- Existing `server.user` key fixtures extended with the new `server.AuthorizedKeys` fact stub (behaviour unchanged).

Full suite passes (1551 tests), lint clean.

### DIFF
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -810,6 +810,39 @@ class Users(FactBase):
         return users
 
 
+class AuthorizedKeys(FactBase[List[str]]):
+    """
+    Returns the SSH public keys listed in a user's ``~/.ssh/authorized_keys`` file as a
+    list of full key strings. Empty lines and lines starting with ``#`` are skipped; the
+    file's order is preserved.
+
+    .. code:: python
+
+        [
+            "ssh-ed25519 AAAAC3Nz... user@host",
+            "ssh-rsa AAAAB3Nz... other@host",
+        ]
+    """
+
+    default = list
+
+    @override
+    def command(self, user: str, path: Optional[str] = None) -> str:
+        # Tilde expansion resolves the user's home without another fact round-trip.
+        target = path if path is not None else "~{0}/.ssh/authorized_keys".format(user)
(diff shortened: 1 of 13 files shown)


Fizzadar (approved):
Looks good to me!

PR #1672, opened 2026-04-19
Add support for using AI coding agents

DonDebonair:
This PR introduces basic support for leveraging coding agents - like Claude Code - when working on pyinfra. This includes the following:

1. `AI_POLICY.md` is added to explain how we (maintainers) think about AI and what we expect when people contribute
2. `AGENTS.md` is added to educate agents on how pyinfra is organised, how to run tests etc.
3. `CLAUDE.md` is added with a reference to `AGENTS.md`

Note: I made some small tweaks to `.editorconfig` to appease PyCharm on line length.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,12 @@
+[*]
+end_of_line = lf
+insert_final_newline = true
+max_line_length = 100
+
+[*.py]
+indent_style = space
+indent_size = 4
+
 [*.json]
 indent_style = space
 indent_size = 4
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ docs/_deploy_globals.rst
 
 .idea/
 .vscode/
+.claude/settings.local.json
 
 pyinfra-debug.log
 Makefile
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,155 @@
+# AGENTS.md
+
+This file provides guidance to coding agents, like Claude Code (claude.ai/code), when working with 
+code in this repository.
+
+## About pyinfra
+
+pyinfra turns Python code into shell commands and runs them on servers. Think Ansible, but Python
+instead of YAML, and much faster. It supports SSH, local machine, Docker, and more via connectors.
+
+## Development Setup
+
+```bash
+uv sync  # Install all dependencies into managed venv
+```
+
+## Common Commands
+
+```bash
+# Run tests
+scripts/dev-test.sh
+# or directly:
+uv run pytest --cov --disable-warnings -m 'not end_to_end'
+
+# Run fixtures for a single operation or fact
+uv run pytest tests/test_operations.py -k "apt.packages"
+uv run pytest tests/test_facts.py -k "LinuxHardware"
+
+# End-to-end tests (require Docker/SSH/local targets)
+uv run pytest -m end_to_end_local
+uv run pytest -m end_to_end_docker
+
+# Lint and type check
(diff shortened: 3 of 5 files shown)


wowi42:
A few concrete suggestions from someone who's been using OpenCode against this repo for the last week or so. I'll list them in roughly descending order of "an agent will stumble on this."

### 1. Wrong test paths in "Adding Operations / Facts"

> corresponding tests in `tests/test_operations/` and `tests/test_facts/`

Those directories don't exist. Actual layout:

- `tests/test_operations.py` + `tests/test_facts.py` are the *runners*.
- Fixtures live under `tests/operations/<module>.<op>/*.json` and `tests/facts/<module>.<Fact>/*.json`, auto-discovered by the `testgen` metaclass.

An agent reading this line will create a new Python test class when they should just drop a JSON fixture, so it's worth fixing before merge.

### 2. JSON fixture convention is worth naming explicitly

This is the single most non-obvious thing about the repo. Suggested blurb:

> **Operation / fact tests are JSON fixtures, not Python tests.** Each file under `tests/operations/<module>.<op>/*.json` or `tests/facts/<module>.<Fact>/*.json` is of the form:
> ```json
> {"args": [...], "kwargs": {...}, "facts": {"module.Fact": {"object_id=X": <mock>}}, "commands": [...]}
> ```
> Additional keys: `exception: {"name": "OperationError", "message": "..."}`, `noop_description`. To cover a new code path, add a fixture, don't write a new Python test.

### 3. Invariants worth calling out, not just listing

The architecture section lists `arguments.py` / `arguments_typed.py`, but the load-bearing fact is that they **must stay in sync** (CI enforces via `scripts/lint_arguments_sync.py`). Touching one requires touching the other.

Similarly, `pyinfra-metadata.toml` is not mentioned at all. Every operation/fact module is registered there as a plugin with tags; missing this doesn't fail tests but breaks the docs generation.

### 4. Docstring format

pyinfra uses `+ param: description` bullets, parsed by `scripts/generate_operations_docs.py`. An agent left to its own devices will switch to Google/NumPy/Sphinx style and silently break docs. Worth a sentence.

### 5. Style rule: optional params default to `None`, not `""`

This is explicit recent maintainer feedback, I don't know who but I got it on two of my PRs :-) . Plenty of older operations still have `param: str = ""` defaults, so an agent pattern-matching from existing files will replicate the mistake. A one-line rule in AGENTS.md heads that off.

### 6. Shell safety primitives

`StringCommand` + `QuoteString` / `MaskString` (from `pyinfra.api`) are how user-supplied values should be composed into shell commands. The security migration in #1664 recently moved several ops to this. An agent writing `"rm -f {0}".format(path)` unaware of the convention will regress that work.

### 7. The single-op pytest invocation

The `-k "apt."` example works but is broad. The most useful shape for agents is:

```bash
uv run pytest tests/test_operations.py -k "docker_compose"   # single op's fixtures
uv run pytest tests/test_facts.py -k "LinuxHardware"
```

### Nit

`## Aattribution` at the bottom of `AI_POLICY.md` — double-A.

Fizzadar:
> 2. JSON fixture convention is worth naming explicitly

We should also mention the YAML format here which IMO is much better (perhaps we should recommend it be used).

DonDebonair:
It's kind of ironic that I let Claude generate the AGENTS.md and that I didn't check the output well enough ;) In any case, I made the required changes @wowi42 @Fizzadar

Fizzadar (approved):
LGTM!

PR #1676, opened 2026-04-20
pyinfra_cli: exclude imports from group data

### LINKED ISSUE #1297: imports in group_data files can cause "pyinfra <inventory> debug-inventory" to fail
## Describe the bug
Certain imports can cause the `pyinfra <inventory> debug-inventory` cli command to fail in `pyinfra_cli.util.json_encode` while trying to dump the inventory with `pyinfra_cli.prints.jsonify()`

This doesn't seem to affect anything that i've seen other than the specific command `pyinfra <inventory> debug-inventory`

## To Reproduce
I initially stumbled across this when following the [dynamic inventory example from the docs.](https://docs.pyinfra.com/en/3.x/examples/dynamic_inventories_data.html)

One way to reproduce:
1. Add `from pyinfra import inventory` to any group data file used in the target inventory
2. Run the following cli command `pyinfra <inventory> debug-inventory`

This produces the following error:
```
  File "<...>/site-packages/pyinfra_cli/util.py", line 159, in json_encode
    raise TypeError("Cannot serialize: {0} ({1})".format(type(obj), obj))
TypeError: Cannot serialize: <class 'pyinfra.context.ContextObject'> (<pyinfra.api.inventory.Inventory object at 0x7f5409b6c550>)
```

The same issue can be reproduced by importing any class as a type. However the error becomes the following:
```
  File "<...>/site-packages/pyinfra_cli/util.py", line 157, in json_encode
    return obj.to_json()
           ~~~~~~~~~~~^^
TypeError: <Class>.to_json() missing 1 required positional argument: 'self'
```

## Expected behavior
The core issue seems to stem from pyinfra_cli treating imports as part of inventory data when parsing python files. I'm not sure what the expected behaviour should be because it depends on whether or not this is by design. 

I think there are a few possibilities:
1. pyinfra should not include imports as a part of inventory data.
2. the command  `pyinfra <inventory> debug-inventory` should avoid calling `obj.to_json()` on classes, and should have a default case for objects that don't have a `to_json()` function rather throwing an error.
3. if this is intended behaviour, it should be pointed out in the docs, as well as ways to work around the behaviour when you don't want an import to be included as inventory data for this reason.

## Meta
```
    System: Linux
      Platform: Linux-6.12.13-200.fc41.x86_64-x86_64-with-glibc2.40
      Release: 6.12.13-200.fc41.x86_64
      Machine: x86_64
    pyinfra: v3.2
      click: v8.1.8
      distro: v1.9.0
      gevent: v24.11.1
      jinja2: v3.1.5
      packaging: v24.2
      paramiko: v3.5.1
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      setuptools: v75.8.0
      typeguard: v4.4.1
      typing-extensions: v4.12.2
    Executable: <...>/.venv/bin/pyinfra
    Python: 3.13.2 (CPython, GCC 14.2.1 20250110 (Red Hat 14.2.1-7))
```

Installed locally to project using python poetry (installs to venv with pip).

```
  File "<...>/.venv/lib/python3.13/site-packages/pyinfra_cli/main.py", line 222, in cli
    _main(*args, **kwargs)
    ~~~~~^^^^^^^^^^^^^^^^^
  File "<...>/.venv/lib/python3.13/site-packages/pyinfra_cli/main.py", line 341, in _main
    print_inventory(state)
    ~~~~~~~~~~~~~~~^^^^^^^
  File "<...>/.venv/lib/python3.13/site-packages/pyinfra_cli/prints.py", line 118, in print_inventory
    click.echo(jsonify(host.data, indent=4, default=json_encode), err=True)
               ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<...>/.venv/lib/python3.13/site-packages/pyinfra_cli/prints.py", line 52, in jsonify
    return json.dumps(data, *args, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
          ~~~~~~^^^^^
  File "/usr/lib64/python3.13/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.13/json/encoder.py", line 261, in iterencode
    return _iterencode(o, 0)
  File "<...>/.venv/lib/python3.13/site-packages/pyinfra_cli/util.py", line 130, in json_encode
    raise TypeError("Cannot serialize: {0} ({1})".format(type(obj), obj))
TypeError: Cannot serialize: <class 'pyinfra.context.ContextObject'> (<pyinfra.api.inventory.Inventory object at 0x7f56f865c550>)
```

wowi42:
Closes #1297.

## Problem

`pyinfra <inventory> debug-inventory` crashes when a group data file imports anything:

```python
# group_data/all.py
from pyinfra import inventory
```

`_get_group_data` exec's the file and treats every module-level name (apart from those starting with `__`) as a piece of group data. Imports therefore end up as data: a module, class, or function. `debug-inventory` then dies in `json_encode` with:

```
TypeError: Cannot serialize: <class 'pyinfra.context.ContextObject'> (<pyinfra.api.inventory.Inventory object ...>)
```

## Fix

Parse the file's AST and collect the names bound by top-level `import` / `from ... import` statements (respecting `as` aliases). Filter those names out of the data dict, unless the file explicitly defines `__all__`, in which case the user is overriding exports deliberately and we honor that.

`from module import *` can't be resolved statically, so it's left alone. Users hitting that rare case can still alias names with a leading double underscore (the existing workaround).

## Test plan

- [x] New regression test `test_imports_do_not_leak_into_group_data`: a fixture imports three things and declares one real `exported_value`; the test asserts the imports are absent from `inventory.group_data` and the real value survives.
- [x] `uv run pytest` (1568 passed)
- [x] `uv run ruff check` / `ruff format --diff` / `mypy` / `scripts/lint_arguments_sync.py` all clean

---

- [x] Pull request is based on the default branch (\`3.x\` at this time)
- [x] Pull request includes documentation for any new/updated operations/facts (behavior fix, no public API / docs change)
- [x] Tests pass (see \`scripts/dev-test.sh\`)
- [x] Type checking & code style passes (see \`scripts/dev-lint.sh\`)


src/pyinfra_cli/inventory.py
```diff
@@ -55,6 +56,31 @@ def _is_inventory_group(key: str, value: Any):
...
+    Return the set of names bound by ``import`` / ``from ... import`` statements
+    in ``filename``. Used to keep those names out of the resulting group data dict
+    (issue #1297) so that e.g. ``from pyinfra import inventory`` in a group data
+    file does not end up as a piece of group data and break ``debug-inventory``.
+    """
+    with open(filename, "r", encoding="utf-8") as f:
+        tree = ast.parse(f.read(), filename=filename)
+
+    names: Set[str] = set()
+    for node in ast.walk(tree):
```
Fizzadar, line 70:
```suggestion
    for node in tree.body:
```

Imports are all top level so avoids walking nested nodes.

Fizzadar (changes requested):
One change to prevent walking the entire ast, looks good otherwise!

PR #1193, opened 2024-09-01
Improve crontab

JakkuSakura:
This should be able to fix #1192, fix #1191, fix #1189

PR #1678, opened 2026-04-20
facts.crontab: match full crontab(5) env var syntax

### LINKED ISSUE #979: fact.Crontab cannot handle variables resulting in crash
## Describe the bug

setting variables in crontab files is not supported and results in a crash

from `man 5 crontab`:

> Blank lines and leading spaces and tabs are ignored.  Lines whose first non-space character
> is a pound-sign (#) are comments, and are ignored.  Note that comments are not allowed
> on the same line as cron commands, since they will be taken to be part of the command.
>  Similarly, comments are not allowed on the same line as environment variable settings.
>
> An active line in a crontab will be either an environment setting or a cron command.
> An environment setting is of the form,
>
>         name = value
>
> where the spaces around the equal-sign (=) are optional, and any subsequent non-leading
> spaces in value will be part of the value assigned to name.  The value string may be placed
> in quotes (single or double, but matching) to preserve leading or trailing blanks.  The name
> string may also be placed in quote (single or double, but matching) to preserve leading,
> trailing or inner blanks.

## To Reproduce

crontab:
```
MAILTO=me@example.com
CRON_TZ=Europe/Amsterdam

@reboot echo "rebooted"
```

```
$ pyinfra server fact server.Crontab
...

...venv/lib/python3.10/site-packages/pyinfra/facts/server.py", line 387, in process
    minute, hour, day_of_month, month, day_of_week, command = line.split(None, 5)
ValueError: not enough values to unpack (expected 6, got 1)
```

## Meta

```
    System: Darwin
      Platform: macOS-13.2.1-arm64-arm-64bit
      Release: 22.3.0
      Machine: arm64
    pyinfra: v2.6.2
    Executable: ./venv/bin/pyinfra
    Python: 3.10.11 (CPython, Clang 14.0.0 (clang-1400.0.29.202))
```

wowi42:
Closes #979.

## Problem

`server.Crontab` crashes when the crontab contains an environment variable declaration whose form doesn't exactly match `[A-Z_]+=value`:

```
$ pyinfra server fact server.Crontab
...
facts/crontab.py: line.split(None, 5)
ValueError: not enough values to unpack (expected 6, got 1)
```

The original fix in #1193 added `_crontab_env_re = r"^\s*([A-Z_]+)=(.*)$"`, which handles `MAILTO=me@example.com` but not valid crontab(5) forms that appear in the wild:

- `MAILTO = me@example.com` (spaces around `=`, explicitly allowed by `man 5 crontab`)
- `cron_tz = Europe/Amsterdam` (lowercase names)
- `"name with spaces"=value` (quoted names, allowed by the spec)

Any of these falls through to the 5-field split and crashes with the same `ValueError`.

## Fix

Relax `_crontab_env_re` to match the `man 5 crontab` spec: any-case identifier OR matching-quoted name, with optional whitespace around `=`. The regex now only needs to recognise the line as an env declaration; the full raw line is still stored in the `env` field so `CrontabFile.format_item()` preserves the original text on round-trip.

## Test plan

- [x] Two new fact fixtures under `tests/facts/crontab.Crontab/`:
  - `env_variables.json` covers the exact repro from #979 (`MAILTO=...`, `CRON_TZ=...`, mixed with `@reboot` and a standard cron line).
  - `env_variables_spaced_and_lowercase.json` covers spaced-around-`=` and lowercase-name forms that used to crash.
- [x] `uv run pytest` (1569 passed)
- [x] `uv run ruff check` / `ruff format --diff` / `mypy` / `lint_arguments_sync.py` clean

---

- [x] Pull request is based on the default branch (\`3.x\` at this time)
- [x] Pull request includes documentation for any new/updated operations/facts (behavior fix, no public API change)
- [x] Tests pass (see \`scripts/dev-test.sh\`)
- [x] Type checking & code style passes (see \`scripts/dev-lint.sh\`)

### DIFF
--- a/src/pyinfra/facts/crontab.py
+++ b/src/pyinfra/facts/crontab.py
@@ -102,7 +102,21 @@ class CrontabFile:
         return self.commands
 
 
-_crontab_env_re = re.compile(r"^\s*([A-Z_]+)=(.*)$")
+# crontab(5) env line: ``name = value`` with optional spaces around ``=``. The
+# name may be a bare identifier or placed in matching single/double quotes; the
+# value runs to end-of-line (the caller handles any quoting in the value).
+_crontab_env_re = re.compile(
+    r"""
+    ^\s*
+    (?:
+        "[^"]*"                         # "quoted name"
+      | '[^']*'                         # 'quoted name'
+      | [A-Za-z_][A-Za-z0-9_]*          # bare identifier
+    )
+    \s*=
+    """,
+    re.VERBOSE,
+)
 
 
 class Crontab(FactBase[CrontabFile]):
--- a/tests/facts/crontab.Crontab/env_variables.json
+++ b/tests/facts/crontab.Crontab/env_variables.json
@@ -1,0 +1,34 @@
+{
+    "command": "crontab -l || true",
+    "requires_command": "crontab",
+    "output": [
+        "MAILTO=me@example.com",
+        "CRON_TZ=Europe/Amsterdam",
+        "@reboot echo rebooted",
+        "0 0 * * * apt update"
+    ],
+    "fact": [
+        {
+            "env": "MAILTO=me@example.com",
+            "comments": []
+        },
+        {
+            "env": "CRON_TZ=Europe/Amsterdam",
+            "comments": []
+        },
+        {
+            "command": "echo rebooted",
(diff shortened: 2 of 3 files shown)


Fizzadar (approved):
Nice fix @wowi42! LGTM

PR #1680, opened 2026-04-20
Feature/sops

maisim:
Depends on https://github.com/pyinfra-dev/pyinfra/pull/1677

Add SOPS facts and operations, improve secrets docs
- SopsKey fact: reads a single key from a SOPS-encrypted file
- sops.set_key: idempotent set/remove of keys in SOPS files
- value=str, check fact before calling for bootstrap pattern
- improve secret_storage.md, add SOPS instructions

### DIFF
--- a/docs/examples/secret_storage.md
+++ b/docs/examples/secret_storage.md
@@ -3,11 +3,107 @@ orphan: true
 ---
 # Using Secrets in pyinfra
 
-Encrypting sensitive information can be achieved using Python packages such as [privy](https://pypi.org/project/privy/).
+## SOPS
+
+[SOPS](https://github.com/getsops/sops) encrypts individual values in JSON/YAML files using age or PGP keys. pyinfra includes built-in support via `facts.sops` and `operations.sops`.
+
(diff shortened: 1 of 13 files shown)


Fizzadar (commented):
Discussed briefly in the Matrix room - I think this would be best suited as a separate operation package `pyinfra-sops`/similar, the scope (writing encrypted secrets to remote host using sops) is very niche and doesn't align with other builtin operations.

Also noting that for most cases any secrets should be read in during inventory generation, rather than at execution.

Fizzadar:
@maisim if you do make this a package and added a pyinfra-metadata file it'd be possible to list on the main docs as well. (ex: https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra-metadata.toml)

maisim:
> @maisim if you do make this a package and added a pyinfra-metadata file it'd be possible to list on the main docs as well. (ex: https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra-metadata.toml)

I will do it asap, thanx

PR #1683, opened 2026-04-21
meta: add PR review skill

Fizzadar:
Adds a Claude Code skill for reviewing open PRs, with a sync script that tracks review state and skips unchanged PRs.

---

*Written with assistance from Claude.*

### DIFF
--- a/.claude/skills/review-prs/SKILL.md
+++ b/.claude/skills/review-prs/SKILL.md
@@ -9,7 +9,7 @@ Review all open PRs on pyinfra-dev/pyinfra and maintain review files in `.prs/`:
 
 1. **Run sync script**: Execute `bash .claude/skills/review-prs/sync.sh` to clean up stale reviews and get the list of PRs to update/create.
 
-2. **For each PR** in the sync output (Update + New only, NOT Unchanged), launch agents in parallel (batches of 3-5) to review:
+2. **For each PR** in the sync output (Update + New only, NOT Unchanged), launch agents using rolling concurrency (see Parallelism below) to review:
    - Each agent gets: repo name, PR number, whether it's an update or new review.
    - Agent fetches the PR diff via `gh pr diff --repo pyinfra-dev/pyinfra {number}`.
    - Agent also fetches `updatedAt` via `gh pr view --repo pyinfra-dev/pyinfra {number} --json updatedAt --jq .updatedAt` and includes it in the review header as `**PR Updated:** {timestamp}`.
--- a/.claude/skills/review-prs/SKILL.md
+++ b/.claude/skills/review-prs/SKILL.md
@@ -51,7 +51,13 @@ Review all open PRs on pyinfra-dev/pyinfra and maintain review files in `.prs/`:
 
 ## Parallelism
 
-- Process multiple PRs in parallel using Agent tool where possible (batch into groups of 3-5).
+Use **rolling concurrency**, not fixed batches. The goal is to keep ~5 PR reviews in flight at all times so a slow review never blocks a fast one.
+
+- Spawn each PR-review agent with `run_in_background: true` so completions arrive as individual notifications instead of blocking on a whole batch.
+- **Initial fill**: in a single message, launch up to 5 background agents (one per PR) to saturate the in-flight pool.
+- **On each completion notification**: if any PRs remain in the queue, immediately launch one new background agent to replace the finished one — keep the in-flight count at 5 until the queue is empty.
+- Track three sets in your head (or via TaskCreate if it helps): `queued` (not yet started), `in_flight` (background agents running), `done` (review file written). Move PRs between sets as agents complete.
+- Do not wait for the full pool to drain before launching replacements — the whole point is to avoid that.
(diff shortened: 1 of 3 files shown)


DonDebonair:
I tried the skill and it works pretty well! Should give us a good head-start with reviews.

2 things I noticed though:

**1. Subagents cannot use bash, so the main agent does all the reviews in serial**

This could probably be fixed by adding something like this to the frontmatter:

```
tools: Read, Grep, Glob, Bash
```

**2. Each PR review gets a verdict, as the skill suggests, but in the summary at the end, some PRs get multiple verdicts**

This is the output when I ran it:

```
  GO (merge-ready) — 20 PRs:
  - #1431 shellcheck CI, #1558 systemd state fix, #1601 brew version parse, #1625 docker custom command, #1642 command injection fix, #1647 requires_command sentinel, #1655 dnf5 compat, #1660 DNF modules,
  #1662 --json flag, #1666 _temp_dir askpass, #1669 EtcHosts/Last/LoadAverage, #1671 UFW module, #1672 AI agent docs, #1675 docker compose/build/login, #1676 group data import leak, #1682 llms.txt, #1683 PR
  review skill, #1684 repo filename field, #1685 BSD rc.d dispatch, #1686 facts docs labels, #1687 download reconcile, #1690 git no-pull when up-to-date, #1691 CI improvements

  NO-GO — 2 PRs:
  - #1301 — draft/incomplete docs, not linked in toctree
  - #1530 — locale.gen entry truncated by sed (strips required charset suffix)

  NEEDS-DISCUSSION — 11 PRs:
  - #1482 — major v4.0 asyncio/AsyncSSH rewrite
  - #1487/#1597 — apt key/sources series (API changes, series dependency)
  - #1556/#1689 — conflicting sudo single-try fixes (pick one)
  - #1557 — Ansible bridge (GPLv3 file, optional dep, belongs in core?)
  - #1594 — removes choco (breaking), adds upgrade facts
  - #1658 — EXECUTION_ONLY flag design questions
  - #1690 — git pull skip (extra network calls)
  ```
  
  #1690 for example is listed as both `GO` and `NEEDS-DISCUSSION`.
  
  The skill does not explicitly say that it should give this summary at the end, so Claude made that up. We could make the summary explicit and give guidance that the classification should match what's in the reviews.

**Comment by Fizzadar** [comment]
@DonDebonair updated the prompt + frontmatter per above. One thing I noticed it this requires having accept edits enabled or the agents will get stuck writing (eventually the main chat will read their output and write them itself), not sure that's easy to pre-approve permissions for though 🤔

File: .claude/skills/review-prs/SKILL.md [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+tools: Read, Grep, Glob, Bash
+---
+
+Review all open PRs on pyinfra-dev/pyinfra and maintain review files in `.prs/`:
+
+## Steps
+
+1. **Run sync script**: Execute `bash .claude/skills/review-prs/sync.sh` to clean up stale reviews and get the list of PRs to update/create.
+
+2. **For each PR** in the sync output (Update + New only, NOT Unchanged), launch agents in parallel (batches of 3-5) to review:
```
DonDebonair:
```suggestion
2. **For each PR** in the sync output (Update + New only, NOT Unchanged), launch agents using rolling concurrency (see Parallelism below) to review:
```

.claude/skills/review-prs/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+- **Code snippets**: When referencing problematic code, include the relevant snippet (keep short).
+- **Deep inspection**: Always read the actual source code being modified. Never assume behavior - trace code paths, check callers, verify types. Use Grep/Read extensively.
+- **Verdict options**:
+   * **GO** = safe to merge as-is or with trivial nits only
+   * **NO-GO** = has bugs, security issues, breaking changes, or significant design problems
+   * **NEEDS-DISCUSSION** = requires maintainer input on approach/design
+
+## Parallelism
+
+- Process multiple PRs in parallel using Agent tool where possible (batch into groups of 3-5).
```
DonDebonair:
```suggestion
Use **rolling concurrency**, not fixed batches. The goal is to keep ~5 PR reviews in flight at all times so a slow review never blocks a fast one.

- Spawn each PR-review agent with `run_in_background: true` so completions arrive as individual notifications instead of blocking on a whole batch.
- **Initial fill**: in a single message, launch up to 5 background agents (one per PR) to saturate the in-flight pool.
- **On each completion notification**: if any PRs remain in the queue, immediately launch one new background agent to replace the finished one — keep the in-flight count at 5 until the queue is empty.
- Track three sets in your head (or via TaskCreate if it helps): `queued` (not yet started), `in_flight` (background agents running), `done` (review file written). Move PRs between sets as agents complete.
- Do not wait for the full pool to drain before launching replacements — the whole point is to avoid that.
```

DonDebonair (changes requested):
Parallelism works now, so that's great. The only issue I found, is that it only processes full batches of 5, waiting until the batch is fully processed before taking on another review.

Making the changes that I suggest, fixes this. I tested it out and it works great that way!

DonDebonair (approved):

PR #1684, opened 2026-04-21
facts.{yum,dnf,zypper}: add filename field to each repository entry

### LINKED ISSUE #1042: fact YumRepositories should include the repo filename
The pattern of inspecting repositories and removing some is not possible because the `.repo` filename is not provided by the `YumRepositories` fact, but `yum.repo(..., present=False)` requires the filename of the repo.

YumRepositories should return output including the filename:
```
[
    {
        "name": "CentOS-$releasever - Apps",
        "baseurl": "http://mirror.centos.org/$contentdir/$releasever/Apps/$basearch/os/",
        "gpgcheck": "1",
        "enabled": "1",
        "gpgkey": "file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial",
        "filename": "/etc/yum.repos.d/centos.repo"
    },
]
```

I'm not sure if the full path should be included, since `yum.repo` doesn't expect it.

EDIT: oh... I see now how this would be impossible given the current implementation of YumRepositories

wowi42:
Closes #1042.

## Problem

`YumRepositories`, `DnfRepositories`, and `ZypperRepositories` concatenate every `.repo` file via `cat`, so each parsed entry loses its source filename. That makes the inspect-then-clean-up pattern unworkable: `yum.repo(..., present=False)` needs the filename, but the fact never exposes it. The issue author ran into exactly this.

## Change

- New helper `make_cat_files_command_with_markers(marker, *filenames)` in `src/pyinfra/facts/util/__init__.py`. It uses a portable `for f in ...; do [ -f "$f" ] && { printf "MARKER %s\n" "$f"; cat "$f"; }; done || true` loop so missing files and unexpanded globs are skipped silently, and each file's contents are preceded by a single marker line naming the path.
- `_parse_yum_or_zypper_repositories` in `src/pyinfra/facts/util/packaging.py` now recognises the `#pyinfra-filename:` marker and attaches a `filename` key to every repo entry it emits. Non-marker comments keep their existing skip behaviour, so any user-written comments in `.repo` files remain ignored.
- `YumRepositories`, `DnfRepositories`, and `ZypperRepositories` switched to the new helper; docstrings updated to show the new `filename` field in the example output.
- Fact fixtures updated (including a second dnf case) to cover the new command string and assert the `filename` field.

The change is additive: existing repo dict keys keep their shape.

## Checklist

- [x] Pull request includes tests for new behaviour (fact fixtures cover the filename marker for yum, dnf, and zypper, including the spaces-in-name case)
- [x] Documentation updated (fact docstrings show the new `filename` field)
- [x] Tests pass (`uv run pytest` - 1608 passed)
- [x] Type checking & code style passes (`uv run ruff check`, `uv run ruff format`, `uv run mypy`, `uv run python scripts/lint_arguments_sync.py`)

### DIFF
--- a/src/pyinfra/facts/dnf.py
+++ b/src/pyinfra/facts/dnf.py
@@ -4,8 +4,8 @@ from typing_extensions import override
 
 from pyinfra.api import FactBase
 
-from .util import make_cat_files_command
-from .util.packaging import parse_yum_repositories
+from .util import make_cat_files_command_with_markers
+from .util.packaging import REPO_FILENAME_MARKER, parse_yum_repositories
 
 
 class DnfRepositories(FactBase):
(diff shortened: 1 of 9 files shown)


Fizzadar (approved):
Nice clean fix for this, LGTM!

PR #1686, opened 2026-04-21
scripts/generate_facts_docs: add canonical-name label for re-exported facts

wowi42:
Fixes:

```
docs/operations/zfs.rst:8: WARNING: undefined label: 'facts:zfs.zfsdatasets' [ref.ref]
```

`facts/zfs.py` re-exports each class under a short alias (`Datasets = ZfsDatasets`, etc.). `remove_dups()` keeps the alphabetically-first name (the alias), so the fact doc only labels `facts:zfs.Datasets`, but the operation doc references `facts:zfs.ZfsDatasets` because `operations/zfs.py` imports the canonical name.

This commit emits the class's canonical `__name__` as an extra label when it differs from the alias picked by `remove_dups`. Both forms resolve; modules without aliases are unaffected; existing URLs stay valid.

## Checklist

- [x] Based on `3.x`
- [x] `sphinx-build` is now warning-free
- [x] `uv run pytest`, `ruff check`, `ruff format` pass

### DIFF
--- a/scripts/generate_facts_docs.py
+++ b/scripts/generate_facts_docs.py
@@ -104,6 +104,13 @@ def build_facts_docs():
                     )
 
             lines.append(".. _facts:{0}.{1}:".format(module_name, name))
+            # Modules that re-export classes under an alias (e.g. facts/zfs.py
+            # exposes both ZfsDatasets and Datasets pointing at the same class)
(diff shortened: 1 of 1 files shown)


DonDebonair:
LGTM

PR #1687, opened 2026-04-21
operations.files.download: reconcile mode/user/group on existing files

### LINKED ISSUE #1200: files.download does not correct the file mode
## Describe the bug
files.download does not correct the mode of already existing files, even if they differ from the specified ones. The outcome of the operation is therefore dependent on the state of the machine before the operation, and may differ from the state requested in the operation. I would consider this a bug.

## To Reproduce
```console
$ pyinfra -y @local files.download https://get.k3s.io ./here mode=u=rw
[...]
$ ll here
-rw-r--r-- 1 rickra rickra 36K Sep  7 08:00 here
$ pyinfra -y @local files.download https://get.k3s.io ./here mode=777
[...]
$ ll here  # here, the permissions should be adjusted, but arent
-rw-r--r-- 1 rickra rickra 36K Sep  7 08:00 here
$ rm here
$ pyinfra -y @local files.download https://get.k3s.io ./here mode=777
[...]
$ ll here
-rwxrwxrwx 1 rickra rickra 36K Sep  7 08:04 here
```

This is because [files.download only checks if any file exists or the hashes differ](https://github.com/pyinfra-dev/pyinfra/blob/3.x/pyinfra/operations/files.py#L122) from optionally supplied ones. There should be an additional mode check.

## Expected behavior
If the file mode differs, the file mode should at least be adjusted to match the requested mode.
I would personally argue that the file should be redownloaded when the mode argument is changed, since I would translate the operation name "download" as "download if the state does not match the expected" (as in file does not exist, file too old, ...).

wowi42:
download() short-circuits to a noop whenever the destination exists (and cache_time / checksums don't say otherwise), so changing the mode, user or group argument had no effect unless you manually deleted the target first. Compare the requested mode/owner against the File fact and emit chmod/chown when they differ; only noop when the file and its metadata both match.

Widen the mode parameter's type to int | str | None to match files.file() and let ensure_mode_int() reassignment type-check cleanly. Closes #1200.

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -80,7 +80,7 @@ def download(
     dest: str,
     user: str | None = None,
     group: str | None = None,
-    mode: str | None = None,
+    mode: int | str | None = None,
     cache_time: int | None = None,
     force=False,
     sha384sum: str | None = None,
@@ -129,6 +129,7 @@ def download(
         )
     """
 
+    mode = ensure_mode_int(mode)
     info = host.get_fact(File, path=dest)
 
     # Destination is a directory?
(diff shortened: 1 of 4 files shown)


Fizzadar (approved):
Oof good spot, LGTM!

PR #1690, opened 2026-04-22
operations.git: don't pull when already up to date

### LINKED ISSUE #994: git pull returns Success when already has all changes
## Describe the bug

When pulling from a git repository it outputs the Success state in situations where it should output No Changes

debug output: 
```
[lima] >>> sh -c 'cd /home/fimdomeio.linux/src/neovim && git pull'
[lima] Already up to date.
```

## To Reproduce
pull from a repo.

I'm running  this operation:
```
git.repo(
    name="pull from neovim repo",
    src="https://github.com/neovim/neovim", 
    dest="/home/fimdomeio.linux/src/neovim", 
)

```
operating system is a Ubuntu Virtual Machine (Ubuntu 23.04)

## Expected behavior
I expect "No changes" as a Status

## Meta
pyinfra --support 
```
  System: Darwin
      Platform: macOS-13.4.1-arm64-arm-64bit
      Release: 22.5.0
      Machine: arm64
    pyinfra: v2.7
    Executable: /Users/fimdomeio/.local/bin/pyinfra
    Python: 3.11.4 (CPython, Clang 14.0.3 (clang-1403.0.22.14.1))
```
installed with pipx

wowi42:
Fixes #994.

`git.repo` and `git.worktree` always yielded `git pull`, so pyinfra reported the operation as changed even when git replied "Already up to date". Compare local `HEAD` to the remote branch tip via two new facts (`GitLocalCommit`, `GitRemoteBranchCommit`) and skip the pull when they match.

### DIFF
--- a/src/pyinfra/facts/git.py
+++ b/src/pyinfra/facts/git.py
@@ -69,3 +69,48 @@ class GitTrackingBranch(GitFactBase):
         if m:
             return m.group(1)
         return None
+
+
+_SHA_RE = re.compile(r"[0-9a-f]{7,64}")
+
+
+class GitLocalCommit(GitFactBase):
+    """
+    Returns the SHA of the current ``HEAD`` of a local git repository, or ``None``
+    when the repository does not exist or the command fails.
+    """
+
+    @override
+    def command(self, repo) -> str:
+        return "! test -d {0} || (cd {0} && git rev-parse HEAD 2>/dev/null)".format(repo)
+
+    @override
+    def process(self, output):
+        if not output:
+            return None
+        line = list(output)[0].strip()
+        return line if _SHA_RE.fullmatch(line) else None
+
+
+class GitRemoteBranchCommit(GitFactBase):
+    """
+    Returns the SHA of the tip of ``branch`` on ``remote`` as reported by
+    ``git ls-remote``. Returns ``None`` when the remote is unreachable, the
+    branch does not exist on the remote, or the repository is missing.
+    """
+
+    @override
+    def command(self, repo, remote: str = "origin", branch: str | None = None) -> str:
+        ref = branch if branch else "HEAD"
+        return ("! test -d {0} || (cd {0} && git ls-remote {1} {2} 2>/dev/null | head -n1)").format(
+            repo, remote, ref
+        )
+
+    @override
+    def process(self, output):
+        if not output:
+            return None
+        parts = list(output)[0].strip().split("\t", 1)
+        if parts and _SHA_RE.fullmatch(parts[0]):
+            return parts[0]
+        return None
--- a/src/pyinfra/facts/git.py
+++ b/src/pyinfra/facts/git.py
@@ -71,24 +71,25 @@ class GitTrackingBranch(GitFactBase):
         return None
 
 
-_SHA_RE = re.compile(r"[0-9a-f]{7,64}")
+_SHA_RE = re.compile(r"[0-9a-f]{40}|[0-9a-f]{64}")
 
 
 class GitLocalCommit(GitFactBase):
     """
-    Returns the SHA of the current ``HEAD`` of a local git repository, or ``None``
-    when the repository does not exist or the command fails.
+    Returns the SHA of ``ref`` (defaults to ``HEAD``) in a local git repository,
+    or ``None`` when the repository does not exist, the ref is unknown, or the
+    command fails.
     """
 
     @override
-    def command(self, repo) -> str:
-        return "! test -d {0} || (cd {0} && git rev-parse HEAD 2>/dev/null)".format(repo)
+    def command(self, repo: str, ref: str = "HEAD") -> str:
+        return "! test -d {0} || (cd {0} && git rev-parse {1} 2>/dev/null)".format(repo, ref)
 
     @override
-    def process(self, output):
+    def process(self, output: list[str]):
         if not output:
             return None
-        line = list(output)[0].strip()
+        line = output[0].strip()
         return line if _SHA_RE.fullmatch(line) else None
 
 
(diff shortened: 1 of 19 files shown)


src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
         if m:
             return m.group(1)
         return None
+
+
+_SHA_RE = re.compile(r"[0-9a-f]{7,64}")
```
DonDebonair: `git ls-remote` always returns 40 or 64-character hashes. So why have 7 as minimum in the regex? It could create ambiguity.

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+
+
+class GitLocalCommit(GitFactBase):
+    """
+    Returns the SHA of the current ``HEAD`` of a local git repository, or ``None``
+    when the repository does not exist or the command fails.
+    """
+
+    @override
+    def command(self, repo) -> str:
```
DonDebonair:
```suggestion
    def command(self, repo: str) -> str:
```

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+
+    @override
+    def command(self, repo, remote: str = "origin", branch: str | None = None) -> str:
+        ref = branch if branch else "HEAD"
+        return ("! test -d {0} || (cd {0} && git ls-remote {1} {2} 2>/dev/null | head -n1)").format(
+            repo, remote, ref
+        )
+
+    @override
+    def process(self, output):
```
DonDebonair:
```suggestion
    def process(self, output: list[str]):
```

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+        if not output:
+            return None
+        parts = list(output)[0].strip().split("\t", 1)
```
DonDebonair: According to `FactBase`, `output` is already a list. So why force it into a list again?

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+
+    @override
+    def command(self, repo) -> str:
+        return "! test -d {0} || (cd {0} && git rev-parse HEAD 2>/dev/null)".format(repo)
+
+    @override
+    def process(self, output):
+        if not output:
+            return None
+        line = list(output)[0].strip()
```
DonDebonair: According to FactBase, output is already a list. So why force it into a list again?

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+    Returns the SHA of the current ``HEAD`` of a local git repository, or ``None``
+    when the repository does not exist or the command fails.
+    """
+
+    @override
+    def command(self, repo) -> str:
+        return "! test -d {0} || (cd {0} && git rev-parse HEAD 2>/dev/null)".format(repo)
+
+    @override
+    def process(self, output):
```
DonDebonair:
```suggestion
    def process(self, output: list[str]):
```

src/pyinfra/facts/git.py
```diff
@@ -69,3 +69,48 @@ def process(self, output):
...
+
+class GitRemoteBranchCommit(GitFactBase):
+    """
+    Returns the SHA of the tip of ``branch`` on ``remote`` as reported by
+    ``git ls-remote``. Returns ``None`` when the remote is unreachable, the
+    branch does not exist on the remote, or the repository is missing.
+    """
+
+    @override
+    def command(self, repo, remote: str = "origin", branch: str | None = None) -> str:
```
DonDebonair:
```suggestion
    def command(self, repo: str, remote: str = "origin", branch: str | None = None) -> str:
```

src/pyinfra/operations/git.py
```diff
@@ -154,14 +161,36 @@ def repo(
...
         if branch and branch in (host.get_fact(GitTag, repo=dest) or []):
             git_commands.append(StringCommand("checkout", QuoteString(branch)))
             is_tag = True
         if pull and not is_tag:
-            if rebase:
+            skip_pull = False
+            # If the branch hasn't changed and the local HEAD already matches the
+            # remote tip, a `git pull` would be a no-op. Skip it so pyinfra reports
+            # the operation unchanged rather than always "Success".
+            if not branch_switching:
```
DonDebonair: Why do we only compare local vs remote commits if we didn't switch branch? It could be that the local branch we switched to, has the same tip as remote, right? And then we'd want to skip pull I think.

DonDebonair: Also, after pulling, should we check out the branch that was originally checked out before switching branches?

DonDebonair (changes requested):
Good idea, left some comments to improve the PR.

wowi42:
Thanks for the review @DonDebonair! Addressed everything in `0b1ed264`:

- **`_SHA_RE` regex** : Tightened to `r"[0-9a-f]{40}|[0-9a-f]{64}"` to only accept full SHA-1/SHA-256 hashes, as `git ls-remote` always returns full-length hashes.
- **Type annotations** : Added `repo: str` on both `command()` methods and `output: list[str]` on both `process()` methods.
- **Redundant `list(output)[0]`** : Dropped the extra `list()` wrapping; `FactBase.process` already receives a `list[str]`.
- **Skip-pull on branch switch** : Good catch. Removed the `branch_switching` gate so the check always runs. To make it correct during a switch, `GitLocalCommit` now accepts an optional `ref` argument (defaults to `HEAD`), and the operation passes `ref=effective_branch` so we evaluate the target branch's local tip (not current `HEAD`) against the remote tip. If the local target branch already matches, we still run `fetch` + `checkout` but skip the `pull`. Added a new test fixture `branch_switch_up_to_date.json` covering this path, plus `branch_ref.json` for the new fact argument.
- **Restoring the original branch after pulling** : Intentionally left out of this PR. The operation has always permanently switched to the requested `branch`, so restoring the original branch post-pull would be a behaviour change beyond the scope of this fix. Happy to open a separate issue if you'd like that tracked.

DonDebonair (approved):

PR #1691, opened 2026-04-24
ci: automate release process and harden test workflow

wowi42:
Automates the release pipeline and tightens the existing test workflow.

## Release automation

- New `release.yml`: pushing a `v*` tag builds with `uv build`, checks the artifact versions match the tag, publishes to PyPI via trusted publishing (OIDC, `release` environment), then creates a GitHub release.
- The release body comes from `scripts/generate_changelog.py`, run against the previous tag, with `gh` attribution for external contributors. It falls back to `Release vX.Y.Z` if the script produces nothing.
- New Release Drafter setup keeps a draft of the next release updated from merged PRs. Categories and version resolution map to the labels that already exist in this repo (`API`, `CLI`, `operations`, `facts`, `connectors`, `documentation`, `dependency issue`, `new feature`, `bug`). A `breaking` label needs creating before it can drive major bumps.

## Test workflow

- Concurrency groups cancel superseded runs, for pull requests only, so branch and tag pushes always produce a full result.
- Tests also trigger on `v*` tags.
- `fail-fast: false` on both test matrices, so one failing job does not hide the results of the others.

## Security

- Every action introduced by this PR is pinned to a commit SHA at its latest release: checkout v6.0.3, setup-python v6.2.0, setup-uv v8.2.0, pypi-publish v1.14.0, action-gh-release v3.0.0, release-drafter v7.3.1.
- The Dependabot config was dropped after review (supply chain trade-off, see discussion above).

## Setup needed before this works

- Create the `release` GitHub environment and configure the PyPI trusted publisher for this repo and workflow.
- Optionally create a `breaking` label for major version resolution.

---


.github/dependabot.yml
```diff
@@ -0,0 +1,41 @@
+version: 2
```
Fizzadar: With all the supply chain attacks right now (and beyond, this feels like the beginning) I feel automating dependency update is high risk, low gain. 

Historically legacy setup py meant users got the latest matching major dependency versions but uv changed that to specific versions.

I wonder if we do the delay thing, but that also sucks because if everyone does no one gains anything. 

I think this needs discussion, I recommend moving it out of scope for this PR do now and opening a new issue just for that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,54 @@
...
+permissions:
+  contents: write
+  id-token: write
+
+jobs:
+  release:
+    runs-on: ubuntu-24.04
+    environment: release
+    steps:
+      - uses: actions/checkout@v5
```
Fizzadar: Similar to other comment - let’s pin these to exact commits, supply chain attacks again.

Fizzadar (changes requested):
Quick drive by review, two security related concerns. Release stuff looks neat will play around with it over the weekend!

wowi42:
Thanks for the quick review @Fizzadar! Addressed both security concerns in `26cdc4c3`:

**Dependabot** : Dropped `.github/dependabot.yml` from this PR. Agreed on the supply-chain risk trade-off; happy to open a separate issue to discuss a policy (delay windows, manual triage, or scoping to specific ecosystems) without blocking the release-automation work.

**Pinning new actions to commit SHAs** : Pinned every new action reference introduced by this PR to an exact commit, with a trailing `# v<tag>` comment for readability:

- `.github/workflows/release.yml`
  - `actions/checkout@93cb6efe…  # v5`
  - `actions/setup-python@a26af69b…  # v5`
  - `astral-sh/setup-uv@d0cc045d…  # v6`
  - `pypa/gh-action-pypi-publish@cef22109…  # release/v1`
  - `softprops/action-gh-release@3bb12739…  # v2`
- `.github/workflows/release-drafter.yml`
  - `release-drafter/release-drafter@6a93d829…  # v6`

I left the pre-existing actions in `test.yml` / `docs.yml` alone since they weren't introduced by this PR, happy to do a follow-up that sweeps the whole repo if you'd like that tracked as a separate effort.

DonDebonair:
I wonder if we should first decide on if/how we want to adopt Conventional Commits and see how this changes the use of Release Drafter.

wowi42:
Pushed a few follow-up commits to harden the release flow and align it with the repo. Summary of what changed:

**Release changelog extraction and artifact check** (`0c1033de`)
- The changelog step ran `generate_changelog.py` from the tagged commit. The script's own tag detection found the current tag and produced an empty range, so release notes came out blank. It now resolves the previous tag from the tag's parent (`git describe --tags --abbrev=0 "${VERSION}^"`) and passes that to the script, so the range is correct.
- Renamed the heredoc marker to `CHANGELOG_EOF` and switched the empty check to `[ -s ... ]` so an empty file falls back to a plain `Release <version>` body instead of emitting nothing.
- Added a step that lists `dist/pyinfra-<version>-py3-none-any.whl` and the sdist before publishing, so a tag/artifact version mismatch fails the job before it reaches PyPI.
- Quoted `$GITHUB_OUTPUT` and used `GITHUB_REF_NAME` instead of stripping `refs/tags/` by hand.

**Release Drafter labels** (`63d53eca`)
- The categories and version-resolver referenced labels that do not exist in this repo (`core`, `cli`, `docs`, `enhancement`, `bugfix`, ...). They now map to the actual labels: `API`, `CLI`, `documentation`, `dependency issue`, `new feature` (minor), `bug` (patch). Added a comment noting that `breaking` still needs creating before it can drive a major bump.
- Dropped the `pull_request` trigger and the `pull-requests: write` permission from the release-drafter workflow. The draft only needs to update on push to the version branches, so this removes write access it did not need.

**Test workflow concurrency** (`a3926518`)
- `cancel-in-progress` was always true, which could cancel a branch or release-tag run mid-way. It is now `${{ github.event_name == 'pull_request' }}`, so only superseded PR runs get cancelled and branch/tag pushes always run to completion.

**Action pins** (`df983a54`)
- Bumped the pinned SHAs to current releases (checkout v6.0.3, setup-python v6.2.0, setup-uv v8.2.0, release-drafter v7.3.1, action-gh-release v3.0.0) and corrected the version comments on the pins.

PR #1675, opened 2026-04-20
operations.docker: add compose, login/logout, and build operations

wowi42:
## Summary

- Adds three new docker operations in one place, since they share design (flag assembly, `DockerImage`/`DockerAuths`-based idempotency) and it's easier to review together.
- `docker.compose`: run `docker compose up -d` / `down` against an on-host compose file, with support for `--project-name`, `--pull`, `--build`, `--force-recreate`, `--remove-orphans`, and `-v`. Not idempotent from pyinfra's side; Docker's own layer/state handling keeps re-runs cheap.
- `docker.login` / `docker.logout`: idempotent registry auth. Checks `${DOCKER_CONFIG:-~/.docker}/config.json` via a new `DockerAuths` fact; `login` pipes the password to `--password-stdin` and masks it in logs.
- `docker.build`: wraps `docker build` (or `docker buildx build` via `builder=`). Supports `tags`, `dockerfile`, `build_args`, `labels`, `target`, `platform`, `network`, `cache_from`, `secrets`, `pull`, `no_cache`, and tag-based idempotency (no-op when every declared tag already exists locally, unless `force=True`).

Supersedes #1674 and #1673.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [x] Pull request includes tests for any new/updated operations/facts
- [x] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`)

PR #1674, opened 2026-04-19
operations.docker: add compose operation

wowi42:
Closes #1276.

Adds `docker.compose` so a Docker Compose stack can be deployed as a single pyinfra operation, instead of falling back to `server.shell('docker-compose up -d', _chdir=..., _env=...)`.

## Summary

`docker.compose(src, project="", present=True, pull="", build=False, force_recreate=False, remove_orphans=True, remove_volumes=False, compose_command="docker compose")`

- `src` accepts a single path or a list of paths on the target (repeated `-f`).
- `project` maps to `--project-name`; omitted → compose default (directory name).
- `present=True` runs `up -d`, `present=False` runs `down`.
- `pull` / `build` / `force_recreate` / `remove_orphans` map 1:1 to the documented `docker compose up` flags.
- `remove_volumes` is only honored on `down` (adds `-v`).
- `compose_command` defaults to v2 (`docker compose`); set to `\"docker-compose\"` for the legacy v1 binary.
- `_env` and `_chdir` are the standard pyinfra global operation kwargs and work out of the box, which is what the issue author actually needed for compose variable interpolation.
- Validation: empty `src` → `OperationError`; `pull` must be `\"\"`, `\"always\"`, `\"missing\"`, or `\"never\"`.

### Why not `docker.stack`?

The original issue proposed `docker.stack(...)`. `docker stack` is the Docker Swarm command (deploys a compose file to a Swarm cluster), which is a different workflow. The single-host compose flow the issue actually describes is `docker compose` / `docker-compose`, so the operation is named `docker.compose`.

### Idempotency

Marked `is_idempotent=False`, following the `prune` precedent. `docker compose up -d` is internally idempotent on the Docker side (unchanged services are skipped, changed ones are recreated), so re-runs are safe. A follow-up PR can add a `DockerComposeProject` fact to do a config-hash comparison and make the operation pyinfra-idempotent, but that is intentionally out of scope here.

### Not in this PR

- `DockerComposeProject` fact for config-hash-based idempotency.
- Uploading the compose file from the controller (users combine `files.put` + `docker.compose`).
- Swarm (`docker stack deploy`).

## Tests

7 fixtures under `tests/operations/docker.compose/`:

- `compose_up_basic.json` — single src, default flags.
- `compose_up_multi_file.json` — list of srcs.
- `compose_up_with_project_and_pull.json` — `project`, `pull=\"always\"`.
- `compose_up_full_flags.json` — `build`, `force_recreate`, `remove_orphans`.
[...]

PR #1673, opened 2026-04-19
operations.docker: add login/logout operations

wowi42:
Closes #1548.

Adds `docker.login` / `docker.logout` operations so private registry authentication can be expressed as a deploy step instead of a raw `server.shell('docker login ...')` call.

### `docker.login(username, password, server=\"\", force=False)`

- Pipes the password to `docker login --password-stdin`, so it does not appear in the target's process list.
- Password is wrapped in `MaskString`, so pyinfra's command log shows `***`.
- Idempotent via the new `DockerAuths` fact (see below): no-ops when the server is already present in `auths`. `force=True` re-runs login (for credential rotation).
- `server` defaults to Docker Hub (`https://index.docker.io/v1/`).
- Raises `OperationError` if `username` or `password` is empty.

### `docker.logout(server=\"\")`

- Runs `docker logout [SERVER]`.
- No-ops when the server is not present in `auths`.
- `server` defaults to Docker Hub.

### `DockerAuths` fact

- Reads `${DOCKER_CONFIG:-$HOME/.docker}/config.json`.
- Returns the list of registry server keys under `auths` (empty list if the file is missing or malformed).
- Works with `credsStore` / `credHelpers` setups: the `auths` key is still present in that case, even when the credential itself lives in an external helper.

## Notes

- No new dependencies.
- No changes to existing operations.

## Tests

- 6 operation fixtures for `docker.login` covering: basic login, Docker Hub default server, special-character password (masking + shlex quoting), noop-when-already-authenticated for both private and Docker Hub servers, `force=True` bypass.
- 3 operation fixtures for `docker.logout` covering: private registry, Docker Hub, noop when not authenticated.
- 3 fact fixtures for `DockerAuths` covering: populated config, missing config file, malformed JSON.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [x] Pull request includes tests for any new/updated operations/facts
- [x] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`)

PR #1693, opened 2026-04-24
operations.docker: add compose operation

wowi42:
## Summary

- Adds `docker.compose` to run `docker compose up -d` / `down` against an on-host compose file.
- Supports `--project-name`, `--pull`, `--build`, `--force-recreate`, `--remove-orphans`, and `-v`.
- Not idempotent from pyinfra's side; Docker's own layer/state handling keeps re-runs cheap.

## Test plan

### DIFF
--- a/src/pyinfra/operations/docker.py
+++ b/src/pyinfra/operations/docker.py
@@ -672,7 +672,7 @@ def logout(server: str | None = None):
 
 @operation(is_idempotent=False)
 def compose(
-    src: str,
+    project_directory: str,
     files: str | list[str] | None = None,
     project_name: str | None = None,
     present: bool = True,
@@ -686,15 +686,15 @@ def compose(
     """
     Deploy a Docker Compose stack on the target.
 
-    + src: project directory on the target (maps to ``--project-directory``).
-      Compose discovers ``compose.yaml`` / ``compose.yml`` /
-      ``docker-compose.yaml`` / ``docker-compose.yml`` inside this directory by
-      default.
+    + project_directory: project directory on the target (maps to
+      ``--project-directory``). Compose discovers ``compose.yaml`` /
+      ``compose.yml`` / ``docker-compose.yaml`` / ``docker-compose.yml`` inside
+      this directory by default.
     + files: optional path or list of paths to specific compose file(s) on the
       target (maps to ``-f``); use this to override the default discovery or to
       layer overrides.
     + project_name: compose project name (maps to ``--project-name``; defaults
-      to compose's own default — typically the basename of ``src``)
+      to compose's own default — typically the basename of ``project_directory``)
     + present: ``True`` runs ``up -d``, ``False`` runs ``down``
     + pull: policy for ``up -d --pull`` (``None``, ``"always"``, ``"missing"``, ``"never"``)
     + build: pass ``--build`` on ``up``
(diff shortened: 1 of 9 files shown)


DonDebonair (changes requested):
I'm very happy with this PR because it's needed for one of my own pyinfra use cases. There are some Docker Compose options that this implementation does not cover, but those can be added in a later PR.

One thing I want to put up for discussion: the meaning of the `src` parameter and how it resolves to `-f` in the command. When you normally use Docker Compose, it will happily use the compose file in the current working directory, as long as that file is named `compose.yaml`, `compose.yml`, `docker-compose.yaml` or `docker-compose.yml`. It will use the cwd as project name, unless overridden in the compose file or using `--project-name`.

I like that, because you don't have to explicitly specify the compose file and I think that is how most people use Docker Compose.  When you're configuring servers remotely, you have to make the cwd explicit. Ansible let's you do this with Docker Compose, by specifiying `project_src`. You still don't have to explicitly specify compose file(s).

My proposal would be to make this operation work similarly by making the following changes:

1. `src` takes a `str` and maps to `--project-directory`
2. we add an optional `files` parameter which maps to `-f`

What do you think?

maisim:
> 1. `src` takes a `str` and maps to `--project-directory`

why not use the name `project_directory  ? To get it transparent

src/pyinfra/operations/docker.py
```diff
@@ -554,3 +554,100 @@ def plugin(
...
+    force_recreate: bool = False,
+    remove_orphans: bool = True,
+    remove_volumes: bool = False,
+    compose_command: str = "docker compose",
+):
+    """
+    Deploy a Docker Compose stack on the target.
+
+    + src: path (or list of paths) to compose file(s) already present on the target
+    + project: compose project name (maps to ``--project-name``; defaults to compose's own default)
```
maisim: Same here, why not project_name, to match docker compose cli ?

wowi42:
Thanks @DonDebonair! Agree, the project-directory shape is much closer to how people actually use Compose. Applied your proposal in `88d2a392`:

- `src: str` is now the project directory and maps to ``--project-directory``. Compose's own default discovery picks up ``compose.yaml`` / ``compose.yml`` / ``docker-compose.yaml`` / ``docker-compose.yml`` inside it without anything else from the user.
- New `files: str | list[str] | None = None` parameter maps to ``-f`` for explicit selection or layered overrides (e.g. ``files=["docker-compose.yml", "docker-compose.prod.yml"]``).
- ``project: str | None`` still maps to ``--project-name`` and is unchanged; when omitted Compose now defaults to the basename of the project directory, which I think matches the local-cwd ergonomics you described.
- Docs/examples updated to show both the discovery case and the override case; tests refreshed and a new `compose_up_single_file` test added.

Agree the missing flags (env-files, profiles, scale, wait, abort-on-container-exit, etc.) are best left for follow-up PRs once this shape is settled. ``scripts/dev-test.sh`` and ``scripts/dev-lint.sh`` both green.

DonDebonair:
@wowi42 I think @maisim has a good point. Shall we rename `src` to `project_directory` and `project` to `project_name`?

wowi42:
Agreed @DonDebonair / @maisim — naming after the CLI flags is the most discoverable option. Applied both in `0e469ece` (and the prior `f22d4281` for `project`):

- `src` → `project_directory` (maps to `--project-directory`)
- `project` → `project_name` (maps to `--project-name`)

Also resolved the conflict with `3.x` (now carries the merged `docker.login` / `docker.logout` from #1694). `scripts/dev-test.sh` and `scripts/dev-lint.sh` both green.

PR #1694, opened 2026-04-24
operations.docker: add login/logout operations

wowi42:
## Summary

- Adds idempotent `docker.login` / `docker.logout` operations for Docker registry auth.
- Adds a new `DockerAuths` fact that reads `${DOCKER_CONFIG:-~/.docker}/config.json`.
- `login` pipes the password to `--password-stdin` and masks it in logs.
- No-ops when the server already appears in `DockerAuths` (unless `force=True`).

## Test plan

### DIFF
--- a/src/pyinfra/facts/docker.py
+++ b/src/pyinfra/facts/docker.py
@@ -394,7 +394,7 @@ class DockerImageHistory(_DockerJsonLinesFactBase):
         ).format(image_id)
 
 
-class DockerAuths(FactBase):
+class DockerAuths(FactBase[list[str]]):
     """
     Returns the list of registry servers the current user is authenticated
     against, read from ``${DOCKER_CONFIG:-$HOME/.docker}/config.json``.
@@ -410,7 +410,7 @@ class DockerAuths(FactBase):
         )
 
     @override
-    def process(self, output):
+    def process(self, output: list[str]) -> list[str]:
         try:
             data = json.loads("".join(output))
         except json.JSONDecodeError:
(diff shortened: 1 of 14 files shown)


src/pyinfra/facts/docker.py
```diff
@@ -392,3 +392,27 @@ def command(self, image_id) -> str:
...
+
+    @override
+    def command(self) -> str:
+        return (
+            'config="${DOCKER_CONFIG:-$HOME/.docker}/config.json"; '
+            '[ -r "$config" ] && cat "$config" || echo "{}"'
+        )
+
+    @override
+    def process(self, output):
```
DonDebonair:
```suggestion
    def process(self, output: list[str]):
```

Also, what is the return type of `process` here? Also `list[str]`?

DonDebonair (changes requested):
Love it! Almost feel guilty about the few dumb nitpicks.

src/pyinfra/facts/docker.py
DonDebonair: If you know the return type of `process` btw, you can make `DockerAuths` more typesafe by supplying a type parameter to `FactBase` as well.

wowi42:
Thanks @DonDebonair, actually a nice quality bump. Addressed both in `97cb1b02`:

- Annotated `def process(self, output: list[str]) -> list[str]:` on `DockerAuths`.
- Parameterised the fact as `class DockerAuths(FactBase[list[str]])` so the return type flows through to call sites (`host.get_fact(DockerAuths)` is now typed as `list[str]`).

`scripts/dev-test.sh` and `scripts/dev-lint.sh` both green.

DonDebonair (approved):
LGTM!

PR #1695, opened 2026-04-24
operations.docker: add build operation

wowi42:
## Summary

- Adds `docker.build` wrapping `docker build` (or `docker buildx build` via `builder=`).
- Supports `tags`, `dockerfile`, `build_args`, `labels`, `target`, `platform`, `network`, `cache_from`, `secrets`, `pull`, and `no_cache`.
- Tag-based idempotency: no-ops when every declared tag already exists locally, unless `force=True`.

## Test plan

### DIFF
--- a/src/pyinfra/operations/docker.py
+++ b/src/pyinfra/operations/docker.py
@@ -279,11 +279,13 @@ def build(
     platform: str | None = None,
     network: str | None = None,
     cache_from: str | list[str] | None = None,
+    cache_to: str | list[str] | None = None,
     secrets: list[str] | None = None,
     pull: bool = False,
     no_cache: bool = False,
     force: bool = False,
-    builder: str = "docker build",
+    builder: str | None = None,
+    build_cmd: str = "docker build",
 ):
     """
     Build a Docker image from a context directory or URL.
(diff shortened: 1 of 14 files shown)


src/pyinfra/operations/docker.py
```diff
@@ -268,6 +268,118 @@ def image(image: str, present: bool = True, force: bool = False):
...
+    labels: dict[str, str] | None = None,
+    target: str | None = None,
+    platform: str | None = None,
+    network: str | None = None,
+    cache_from: str | list[str] | None = None,
+    secrets: list[str] | None = None,
+    pull: bool = False,
+    no_cache: bool = False,
+    force: bool = False,
+    builder: str = "docker build",
```
DonDebonair: This could be named `build_cmd`. Both `docker build` (which is an alias for `docker buildx build` on modern versions of Docker btw) and `docker buildx build` have a `--builder` option, which I would expect the `builder` parameter to refer to.

We could do both: have a `builder` parameter which adds the `--builder` option and have a `build_cmd` parameter, which specifies the command used.

src/pyinfra/operations/docker.py
```diff
@@ -268,6 +268,118 @@ def image(image: str, present: bool = True, force: bool = False):
...
+
+    + path: build context (local directory or git/http URL)
+    + tags: tag or list of tags to apply to the built image (maps to ``-t``)
+    + dockerfile: path to the Dockerfile (maps to ``-f``; relative to the context)
+    + build_args: ``--build-arg`` values as a mapping
+    + labels: ``--label`` values as a mapping
+    + target: ``--target`` stage for multi-stage Dockerfiles
+    + platform: ``--platform`` (e.g. ``linux/amd64``)
+    + network: ``--network`` (e.g. ``host``)
+    + cache_from: image or list of images for ``--cache-from``
```
DonDebonair: Technically, these cannot only be image refs, but also specifications for different [cache backends](https://docs.docker.com/build/cache/backends/). The way you process this parameter, supports different backends, but that's not clear from the parameter description.

Also, should we support `--cache-to` as well?

src/pyinfra/operations/docker.py
```diff
@@ -268,6 +268,118 @@ def image(image: str, present: bool = True, force: bool = False):
...
+        else []
+    )
+
+    if tag_list and not force:
+        missing = [tag for tag in tag_list if not host.get_fact(DockerImage, object_id=tag)]
+        if not missing:
+            host.noop("Image(s) {0} already exist!".format(", ".join(tag_list)))
+            return
+
+    command_bits: list = [builder]
```
DonDebonair:
```suggestion
    command_bits: list[str] = [builder]
```

DonDebonair (changes requested):
Looks great! Left a couple of suggestions

PR #1703, opened 2026-04-28
fix(facts): add semicolons to TmpDir shell script for sh -c compatibility

### LINKED ISSUE #1699: TempDir fails when executed with sudo
## Describe the bug
I have recently upgraded from 2.x to 3.x and started getting failed builds. It seems that files.sync operation fails as follows:
```
[myhost] Loaded fact files.File (path=/path/to/file)
[myhost] >>> sudo -H -n -i sh -c 'test -e /path/to/file && ( sha1sum /path/to/file 2> /dev/null || shasum path/to/file 2> /dev/null || sha1 /path/to/file 2> /dev/null ) || true'
[myhost] Loaded fact files.Sha1File (path=/path/to/file)
[myhost] >>> sudo -H -n -i sh -c 'if [ -n "$TMPDIR" ] && [ -d "$TMPDIR" ] && [ -w "$TMPDIR" ]; then
    echo "$TMPDIR"
elif [ -n "$TMP" ] && [ -d "$TMP" ] && [ -w "$TMP" ]; then
    echo "$TMP"
elif [ -n "$TEMP" ] && [ -d "$TEMP" ] && [ -w "$TEMP" ]; then
    echo "$TEMP"
else
    echo ""
fi'
    [myhost] sh: 1: Syntax error: "then" unexpected (expecting "fi")
    [myhost] Error: could not load fact: server.TmpDir 
```

## To Reproduce
I am using SSH connector with `'_sudo': True` and `'_use_sudo_login': True`.

## Expected behavior
I expect the files to sync ;).
I believe the problem is in missing semicolons in conditional bodies, i.e.

```sh
sudo -H -n -i sh -c 'if [ -n "$TMPDIR" ] && [ -d "$TMPDIR" ] && [ -w "$TMPDIR" ]; then
    echo "$TMPDIR";  # semicolon here
elif [ -n "$TMP" ] && [ -d "$TMP" ] && [ -w "$TMP" ]; then
    echo "$TMP";  # here
elif [ -n "$TEMP" ] && [ -d "$TEMP" ] && [ -w "$TEMP" ]; then
    echo "$TEMP";  # here
else
    echo "";  # and here
fi'
```

Should fix the problem. I tested it by running the whole snippet including `sudo` part without and with semicolons, the result was exactly the same - error in the first case, no errors in the second case.

## Meta
Pyinfra v 3.7, installed via pip.

wucm667:
Fixes #1699

## Problem

The `TmpDir` fact's shell script fails when executed via `sh -c` under sudo with login shell (`sudo -H -n -i sh -c '...'`), producing:

```
sh: 1: Syntax error: "then" unexpected (expecting "fi")
```

This breaks `files.sync` operations when using SSH connector with `_sudo: True` and `_use_sudo_login: True`.

## Root Cause

The shell script in `TmpDir.command()` uses multi-line `if/elif/else` blocks with `echo` statements. When passed to `sh -c`, some shells (particularly under sudo login shell) don't properly treat newlines as statement separators, causing the syntax error.

## Changes

- **File**: `src/pyinfra/facts/server.py`
  - Added semicolons after each `echo` statement in `TmpDir.command()` to ensure proper statement termination:
    - `echo "$TMPDIR"` → `echo "$TMPDIR";`
    - `echo "$TMP"` → `echo "$TMP";`
    - `echo "$TEMP"` → `echo "$TEMP";`
    - `echo ""` → `echo "";`

- **Files**: `tests/facts/server.TmpDir/*.json` (4 files)
  - Updated command fixtures to match the new script output

## Testing

All 4 TmpDir fact tests pass:
```
tests/test_facts.py::server.TmpDir::test_server.TmpDir_default_fallback PASSED
tests/test_facts.py::server.TmpDir::test_server.TmpDir_temp_set PASSED
tests/test_facts.py::server.TmpDir::test_server.TmpDir_tmp_set PASSED
tests/test_facts.py::server.TmpDir::test_server.TmpDir_tmpdir_set PASSED
```

### DIFF
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -65,13 +65,13 @@ class TmpDir(FactBase):
     def command(self):
         return """
 if [ -n "$TMPDIR" ] && [ -d "$TMPDIR" ] && [ -w "$TMPDIR" ]; then
-    echo "$TMPDIR"
+    echo "$TMPDIR";
 elif [ -n "$TMP" ] && [ -d "$TMP" ] && [ -w "$TMP" ]; then
-    echo "$TMP"
+    echo "$TMP";
 elif [ -n "$TEMP" ] && [ -d "$TEMP" ] && [ -w "$TEMP" ]; then
-    echo "$TEMP"
+    echo "$TEMP";
 else
-    echo ""
+    echo "";
 fi
         """.strip()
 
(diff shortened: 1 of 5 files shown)


Fizzadar (approved):
Nice catch, thank you @wucm667!

PR #1706, opened 2026-04-29
fix(operations.server): normalize sysctl value for comparison

wowi42:
A space-separated string value (e.g. "1024 65535") was compared against the fact's parsed list (e.g. [1024, 65535]), so the operation always re-ran. Normalize string values containing whitespace into a list of ints, and collapse single-element lists to the bare value, so the comparison matches the fact shape.


Fizzadar (approved):
Neat, LGTM!

PR #1707, opened 2026-04-29
feat(operations.server): pre-check shadow tools in user op

wowi42:
Detect missing useradd/userdel/usermod via the Which fact and raise OperationError with a clear message before the command would fail at runtime. On Alpine, the message points to the shadow package; on other distros it just reports the missing binary.


Fizzadar (commented):
Hrm, I'm not sure on this - without this change the user still gets "command not found" errors for these binaries - is there really any benefit to wrapping those in these cases? I don't want every operation to have such guardrails or we'll be adding thousands of lines of similar checks.

wowi42:
For some context: `server.user` and `server.group` already branch on FreeBSD to call `pw useradd` / `pw usermod` / `pw userdel` instead of the shadow tools, so the operation isn't strictly tied to `useradd` & co being present.

That said, your broader point about not wrapping every operation in guardrails is fair, and I agree with it as a default. Where I'd carve out an exception is that user creation is foundational enough, and the failure mode on Alpine without the `shadow` package confusing enough, that a single `Which` precheck pointing at the right package feels worth it. Not a pattern I'd want spread across hundreds of operations either, so we're aligned on that.

Fizzadar:
> That said, your broader point about not wrapping every operation in guardrails is fair, and I agree with it as a default. Where I'd carve out an exception is that user creation is foundational enough..

Agree! LGTM

Fizzadar (approved):

PR #1711, opened 2026-04-30
fix(operations.server): detach reboot from SSH session

### LINKED ISSUE #1708: Server.reboot operation  incompatible with ProxyCommand
## Describe the bug

When connected to a host with `ProxyCommand` in its SSH config, the `server.reboot` operation hangs indefinitely.

The issue is likely due to [L89](https://github.com/pyinfra-dev/pyinfra/blob/b63d2c1eab653f84d9f05ec347c14074db70ac47/src/pyinfra/operations/server.py#L89). 

- The server drops the connection as expected
- The proxy process  remains alive
- Because the proxy is still running, the connection is not seen as closed
- pyinfra waits indefinitely for a return code that will never arrive

Setting `paramiko.transport.set_keepalive(4)` resolves the hang, but isn't a viable fix — the transport object isn't exposed by pyinfra, and the stale proxy connection causes threading/atexit errors during teardown.

## To Reproduce

1. Configure a host with `ProxyCommand ssh -W %h:%p <proxy_host>` in  its`~/.ssh/config`
2. Perform a`server.reboot` operation

## Expected behavior

The operation should detect that the server has gone away (or timeout) and proceed with reconnection.

## Meta
```
❯ uv run pyinfra --support

    System: Darwin
      Platform: macOS-26.4.1-arm64-arm-64bit-Mach-O
      Release: 25.4.0
      Machine: arm64
    pyinfra: v3.6.1
      click: v8.3.1
      distro: v1.9.0
      gevent: v25.9.1
      jinja2: v3.1.6
      packaging: v26.0
      paramiko: v3.5.1
      pydantic: v2.12.5
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.1
      typing-extensions: v4.15.0
    Python: 3.13.12 (CPython, Clang 21.1.4 )
```

wowi42:
The reboot command ran inline on the SSH channel, so paramiko blocked on `recv_exit_status` for the remote process. With a direct connection the channel returns -1 when the server drops, but with a ProxyCommand the proxy keeps the channel open and pyinfra hangs indefinitely.

Run the reboot in a backgrounded subshell with stdin/stdout/stderr redirected to /dev/null so the parent shell exits cleanly with code 0 and SSH closes the channel right away. The actual reboot fires one second later, by which time the channel is already gone.

Closes #1708


Fizzadar (changes requested):
As noted in the issue I think we need some way to guarantee we know the server has actually rebooted. The current code sort of solves this because it'll exit 1 when the SSH server is shutdown, but of course this breaks proxying.

Fizzadar:
@wowi42 #1737 merged now so we can rebase + land this!

Fizzadar (approved):
LGTM!

PR #1712, opened 2026-04-30
fix(operations.iptables,facts.iptables): extras comparison losing duplicate flags and quoted values

JustScreaMy:

---

Previously, storing extras as a flat set caused duplicate flags (like -m) to be lost and quoted strings (like --comment) to be split incorrectly. This broke idempotency, causing pyinfra to re-add existing rules on every run.

Changes in this PR:
- parses `extras` with `shlex.split()` to properly handle quoted values
- store `extras` as sorted `[flag, value]` pairs instead of flat set
- introduce a shared `parse_extras()` helper for fact and operations
- add a new comment helper argument for `iptables.rule` operation


JustScreaMy:
Hi! Don't want to rush anything here, I saw the PR got labeled, so just wanted to check if there's anyone who could give it a review when they get a chance? Happy to address any feedback. Thanks!

src/pyinfra/operations/iptables.py
```diff
@@ -147,15 +157,18 @@ def rule(
     if source_port:
         extras = "{0} --sport {1}".format(extras, source_port)
 
-    # Convert the extras string into a set to enable comparison with the fact
-    extras_set = set(extras.split())
+    if comment:
+        extras = "{0} -m comment --comment '{1}'".format(extras, comment)
```
Fizzadar: Comment here needs to be escaped using `HiddenValue` (just merged today, `from pyinfra.api import HiddenValue`).

Fizzadar (changes requested):
Sorry for the delay, @JustScreaMy - this looks great, one small fix needed!

src/pyinfra/operations/iptables.py
JustScreaMy: Hello, thanks for the review, no problem changing the operation logic for this. Just wanted to ask, is the `HiddenValue` here necessary? `HiddenValue` is currently used only for secrets, which a comment shouldn't really be in my opinion.

Fizzadar: Ooops I'm completely wrong, we just want `QuoteString` here I think sorry about that!

JustScreaMy: I have been fiddling with `QuoteString` in my updated iptables module for a few hours now, and using the `QuoteString` kinda breaks the whole operation, because of the extras support.

In my opinion, given the current implementation, the whole operation would need to be updated, because of how the individual parts are put together and parsed, for which I currently do not have the capacity. 

Is there some middle ground that we can agree on? Because for me, this fix for duplicate iptables rules is pretty important.

Fizzadar: > Is there some middle ground that we can agree on?

Sorry for the delay @JustScreaMy! Yeah - `shlex.quote(comment)` instead of the hardcoded quotes should do it w/o any rework.

JustScreaMy:
Hi!
I have implemented the changes you have requested, fixed the pull request title and everything to pass all the checks.
Is there anything else i should do for this to be merged?

Fizzadar (approved):
Thank you @JustScreaMy, this looks good to me!

PR #1713, opened 2026-04-30
docs: convert docs to use zensical

Fizzadar:
live here: https://docs.pyinfra.com/tmp-zensical/

IMO this is a significant improvement (and drops custom themeing and rst which has a high learning curve)


DonDebonair (approved):
LG(reat)TM!

PR #1718, opened 2026-05-03
fix(gpg.key): use _matches_keyid for full-fingerprint idempotency check

maisim:
When keyid is a full 40-char fingerprint, the previous inline comparison

    clean_keyid in existing_key_id.upper()

failed because `GpgKeyrings` returns short key IDs (8/16 chars) as dict keys — the fingerprint is never a substring of the short ID.

`_matches_keyid()` already handles all cases (short ID, long ID, full fingerprint) via endswith/startswith logic. Use it consistently with `_remove_key_from_keyrings` / `_remove_key_from_keyring`, which were already calling the helper.

### Behavior delta

The old `in` check additionally allowed a "middle substring" match (e.g. `clean_keyid` sitting inside an existing ID). In practice GPG IDs are always suffixes of the fingerprint, so no real keyring is affected, but the match is now strictly anchored (equality / `endswith` / `startswith` on either side).

### Test matrix added

`tests/operations/gpg.key/`:
- `keyserver_single_idempotent_full_fingerprint.json` — full 40-char fingerprint vs existing 10-char short ID (the bug being fixed)
- `keyserver_idempotent_full_fingerprint_vs_long_id.json` — full fingerprint vs existing 16-char long ID (different-length `endswith` case)
- `keyserver_idempotent_short_id_vs_long_id.json` — 8-char short keyid vs existing 16-char long ID (existing path, previously uncovered)
- `local_file_idempotent_full_fingerprint_vs_long_id.json` — same scenario via `src=` instead of `keyserver=` (shared idempotency path)

### DIFF
--- a/src/pyinfra/operations/gpg.py
+++ b/src/pyinfra/operations/gpg.py
@@ -341,11 +341,9 @@ def key(
                 # Check if all requested keys already exist
                 all_keys_exist = True
                 for kid in keyids_to_check:
-                    # Remove 0x prefix if present for comparison
                     clean_keyid = kid.replace("0x", "").replace("0X", "").upper()
                     key_exists = any(
-                        clean_keyid in existing_key_id.upper()
-                        or existing_key_id.upper().endswith(clean_keyid)
+                        _matches_keyid(existing_key_id, clean_keyid)
                         for existing_key_id in existing_keys.keys()
                     )
                     if not key_exists:
--- a/tests/operations/gpg.key/keyserver_idempotent_full_fingerprint_vs_long_id.json
+++ b/tests/operations/gpg.key/keyserver_idempotent_full_fingerprint_vs_long_id.json
@@ -1,0 +1,42 @@
+{
+    "args": [],
+    "kwargs": {
+        "keyserver": "hkps://keyserver.ubuntu.com",
+        "keyid": ["9DC858229FC7DD38854AE2D88D81803C0EBFCD88"],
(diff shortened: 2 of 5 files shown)


wowi42:
Fix is correct. `_matches_keyid` already had the `clean_key.endswith(existing_upper)` direction needed when the user supplies a full fingerprint and `GpgKeyrings` returns the short or long ID as the dict key, so this just closes the consistency gap with `_remove_key_from_keyrings` and `_remove_key_from_keyring` which were already calling the helper.

A few items before merge:

1. The old `clean_keyid in existing_key_id.upper()` allowed a "middle substring" match, which `_matches_keyid` does not. In practice GPG IDs are always suffixes of the fingerprint so this should not affect anyone, but worth a line in the PR description so the behavior delta is documented.

2. Please add fixtures to lock in the broader matrix:
   - full fingerprint vs an existing 16-char long ID (different-length `endswith` case)
   - 8-char short keyid vs a 16-char long ID (existing path, not regressed but currently uncovered)
   - same scenario via `src=` instead of `keyserver=` if the install paths differ

3. Please tick the PR template checkboxes (branch, tests, docs, lint, type-check) so reviewers can see the checklist state at a glance.

maisim:
Thanks @wowi42  for your relevant feedback, as usual.

Fizzadar (approved):
Nice, thanks for the quick turnaround @maisim!

PR #1721, opened 2026-05-04
style: extend linting with modern Python syntax checks

DonDebonair:
This PR enables [pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up) for `ruff`. Pyupgrade enforces modern (Python 3.10+) syntax. Examples: f-strings, builtin generic containers (e.g. `dict` vs `Dict`) and more.

I extended the linting to also test for "unsafe" rules. The "unsafe" part here does not refer to the rules themselves, but to `ruff`'s ability to safely and automatically make changes for these rules. I manually checked every "unsafe" change.

This PR obviously touches many files, so my advice would be not to try and review every change. Focus on the config changes.


DonDebonair:
The docs fail to build because of a change in type hints. In many place - including places that `scripts/generate_arguments_doc.py` touches - union types are changed from `Union[X, Y]` to `X | Y`. In Python 3.13 and below, these lead to different types (`typing.Union` vs `tying.UnionType`). The former has a `__name__` attribute, the latter doesn't.

In Python 3.14 [this is changed](https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-typing-union). Upgrading the Python version for docs generation to 3.14 fixes the issue, but somehow that change isn't picked up by GH Actions.

I think we can safely merge this, because I did make the required change to the docs workflow.

Fizzadar (approved):
Nice! Love this. GH actions are so weird from PRs - no idea why it didn't pick up the change, will work once merged - LGTM!

PR #1722, opened 2026-05-04
fix(operations.crontab): detect change of crontab entry with only spe…

sebastiaanhoogeveen:
Fix an issue where crontab.crontab with special_time specified would always result in a change being detected, unless explicitliy specifying the other time arguments with a value of None. This was caused by the default values of the other time fields ("*") being compared to the Crontab fact that was returned (None) for crontab entries with a special time set. This PR fixes the issue by changing the default values to None and only assigning the value of "*" if special_time is not specified.

Also includes a small documentation fix in the existing example and added some examples with more complex cron schedules.


wowi42:
Thanks for the fix, the root cause analysis is right and the patch is sound.

A few small items before merge:

1. Syntax error in the new `@reboot` docstring example, missing comma after `name`:

   ```python
   crontab.crontab(
       name="Create a directory on reboot"
       command="/usr/bin/mkdir /var/run/my_directory",
       special_time="@reboot"
   )
   ```

   Worth fixing since users copy-paste from docstrings.

2. The new "Cron schedule" paragraph says *"The default value of all these arguments is '*'."*, but the signature default is now `None`. Could be reworded to something like *"the effective default is `*` when `special_time` is not set"*.

3. Optional: a fixture covering the reverse transition (existing entry has `special_time=@reboot`, call with `minute=5, hour=1` and no `special_time`) would lock in that the `None` to `*` substitution doesn't regress change detection in the other direction. The current fixture covers the exact bug, the existing `edit_special_time.json` covers special-to-special, but special-to-regular is uncovered.

Nothing blocking, behavior change for callers passing `None` explicitly (the documented workaround) is preserved.

sebastiaanhoogeveen:
Fixed the documentation syntax error, changed wording of the defaults and added two tests for changing to and from special_time. I squashed the commits and added the format()-to-f-string change to my commit to hopefully make a clean merge simpler.

Fizzadar (approved):
LGTM, thanks @sebastiaanhoogeveen!

Issue #1725, opened 2026-05-05
Track: split #1594 into phased PRs (unified package model rollout)


wowi42:
Tracking issue for the staged rollout of the unified package model originally proposed in #1594. That PR is too large to review and has rotted against `3.x`, so it will be closed and replaced by a sequence of small PRs.

## Goal

Land the work from #1594 in independently reviewable chunks. Each phase is shippable on its own and does not break unrelated managers. The shared helper `ensure_packages()` accepts both `dict[str, set[str]]` and `dict[str, PackageInfo]` so managers can migrate one at a time.

## Phases (sequential, one PR per phase)

- [x] Phase 1: core model. Add `PackageInfo` and `build_package_map()` in `facts/util/packages.py`; make `operations/util/packaging.ensure_packages()` accept both shapes. No manager changes.
- [ ] Phase 2: `pkg` (FreeBSD). Migrate `PkgPackages` to `PackageInfo`, add `PkgUpgradeablePackages` and `PkgLockedPackages`, add `pkg.update()` and `pkg.upgrade()` operations.
- [ ] Phase 3: `apk` (Alpine).
- [ ] Phase 4: `pacman` (Arch).
- [ ] Phase 5: `apt` and `deb` facts (Debian/Ubuntu).
- [ ] Phase 6: `pkgin` (NetBSD).
- [ ] Phase 7: `opkg` (OpenWrt).
- [ ] Phase 8: `xbps` (Void).
- [ ] Phase 9: `dnf` and `yum` (RHEL family).
- [ ] Phase 10: `zypper` (SUSE).
- [ ] Phase 11: `brew`.
- [ ] Phase 12: `cargo`.
- [ ] Phase 13: `gem`.
- [ ] Phase 14: `npm`. Includes the `NpmOutdatedPackages` directory quoting test fix.
- [ ] Phase 15: `pip`.
- [ ] Phase 16: new manager `portage` (Gentoo).
- [ ] Phase 17: new manager `paludis` (Exherbo).
- [ ] Phase 18: remove `choco` (Chocolatey). Breaking change, needs maintainer sign-off and a changelog note.

PRs will land sequentially so each phase builds on a known-good main.

Closes #1594.

PR #1594, opened 2026-03-10
feat: unified package model with idempotent upgrades across all package managers

wowi42:
## Summary

Introduce a common `PackageInfo` data model (name, version, status) and `build_package_map()` helper so all package managers share a consistent representation of installed, upgradeable, and held packages. Operations like `upgrade()` are now idempotent — they check for available upgrades before running. The `ensure_packages()` function accepts both the old `dict[str, set[str]]` and new `dict[str, PackageInfo]` formats for backward compatibility.

## Changes

- **Unified package model**: Added `PackageInfo` dataclass and `build_package_map()` in `facts.util.packages` to normalize installed/outdated/held package data across all managers.
- **Migrated managers**: apk, apt, brew, cargo, dnf, gem, npm, opkg, pacman, pip, pkg, pkgin, xbps, yum, zypper. Added upgradeable/held sub-facts for each manager that supports them.
- **New `pkg` operations**: Added `pkg.update()` and `pkg.upgrade()`.
- **Added portage and paludis**: New package manager support for Gentoo (`portage`) and Exherbo (`paludis`), including facts, operations, and tests.
- **Removed choco (Chocolatey)**: Removed all choco facts, operations, tests, and metadata.
- **Test fix**: Added missing `NpmOutdatedPackages` fact to npm directory quoting test.

## Checklist

- [X] Pull request is based on the default branch (`3.x` at this time)
- [X] Pull request includes tests for any new/updated operations/facts
- [X] Pull request includes documentation for any new/updated operations/facts
- [X] Tests pass (see `scripts/dev-test.sh`)
- [X] Type checking & code style passes (see `scripts/dev-lint.sh`)

PR #1726, opened 2026-05-05
feat(facts.util.packages): add PackageInfo model (phase 1 of #1725)

wowi42:
## Summary

Phase 1 of the staged rollout tracked in #1725 (replacing the rotted #1594).

- Add `PackageInfo` dataclass, `PackageStatus` enum, and `build_package_map()` in `src/pyinfra/facts/util/packages.py`.
- Update `operations/util/packaging.ensure_packages()` to accept either the legacy `dict[str, set[str]]` or the new `dict[str, PackageInfo]`. When `PackageInfo` is provided, `HELD` packages always noop, `UPGRADEABLE` packages are upgraded when `latest=True`, and `INSTALLED` packages with no available upgrade noop with an "up to date" message.
- Legacy callers are unchanged: with the old shape `latest=True` still blindly adds every versionless package to the upgrade list, and noop messages keep the existing format.
- Make multi-version output deterministic: `build_package_map()` and `_format_version()` sort their input set before consuming it, so noop strings and the chosen `installed_version` are stable across runs. Set iteration order is hash-randomized for strings, so the previous behaviour was non-deterministic.
- No package-manager facts or operations are migrated yet, so the existing dict shape is what every caller still returns.

## Test plan

- [x] `tests/test_facts_packages.py`: unit tests for `PackageInfo` and `build_package_map`, including status precedence, frozen instances, empty version sets, and deterministic selection from a multi-version set.
- [x] `tests/test_operations_utils.py::TestEnsurePackagesDualFormat`: 14 cases covering both shapes; pinned-version match/mismatch against `PackageInfo`; uninstall paths (installed, held, missing); `latest` interaction with `HELD`, `UPGRADEABLE`, and pinned versions; sorted multi-version noop string.

## Requirements


src/pyinfra/facts/util/packages.py
```diff
@@ -0,0 +1,66 @@
...
+    installed_version: str
+    available_version: str | None = None
+    status: PackageStatus = PackageStatus.INSTALLED
+
+
+def build_package_map(
+    installed: dict[str, set[str]],
+    upgradeable: dict[str, str] | None = None,
+    held: set[str] | None = None,
+) -> dict[str, PackageInfo]:
```
DonDebonair, line 65: Given that in the input, `installed` accepts multiple versions of the same package, should the output not also support multiple package versions?

As a sidenote: `apt` does not support multiple package versions to be installed simultaneously, so I'm not sure why we support it in the first place. Maybe other package managers do support this?

src/pyinfra/operations/util/packaging.py
```diff
@@ -203,15 +249,33 @@ def ensure_packages(
...
+                status = _get_package_status(current_packages, pkg_name)
+
+                if status == PackageStatus.HELD:
+                    host.noop(f"package {pkg_name} is held")
+                    continue
+
+                # Present packages w/o version specified: candidate for upgrade
+                if not package.has_version:
+                    if status == PackageStatus.UPGRADEABLE:
+                        upgrade_packages.append(package.inst_vers)
```
DonDebonair, line 265: Should this also happen when `latest == False`?

DonDebonair: Ah, I see now on L295 that `upgrade_packages` is only used when `latest == True`.

src/pyinfra/facts/util/packages.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+
+    result: dict[str, PackageInfo] = {}
+    _upgradeable = upgradeable or {}
+    _held = held or set()
+
+    for name, versions in installed.items():
+        # Sort so a multi-version set produces the same PackageInfo.installed_version
+        # across runs (sets have hash-randomized iteration order for strings).
+        version = next(iter(sorted(versions)), "")
```
DonDebonair: If we pick only 1 version when multiple versions are installed, doesn't it make sense to pick the highest version?

DonDebonair (commented):
I find the logic of `build_package_map` a bit confusing, because in case of multiple installed versions, we only acknowledge 1. I left some comments to discuss this.

src/pyinfra/facts/util/packages.py
wowi42: Good catch, fixed in a20ed4c2. `PackageInfo.installed_version: str` is now `installed_versions: tuple[str, ...]` (sorted ascending), with `installed_version` kept as a property returning the highest entry for the common single-version case. `_has_package` matches a pinned request against any installed version, and `_format_version` joins them all in the noop output. Single-version managers just store a 1-tuple.

On the sidenote: apt itself sticks to one version per package per architecture, but multi-arch (i386 + amd64 of the same name) and rpm-family installonly packages (kernels) genuinely produce multiple entries. portage SLOTs are another case. So the set on the input side is reasonable even though most managers collapse it to one element.

src/pyinfra/facts/util/packages.py
wowi42: Done in a20ed4c2. `installed_version` is now `installed_versions[-1]` (or `""` when none reported). The sort uses a natural-order key (digit runs as integers) so `5.10` sorts after `5.2` and `9.0` before `9.0-1`. Good enough as a cross-distro default; not a substitute for rpm/dpkg vercmp, but those would need per-manager logic that does not belong in this util.

src/pyinfra/operations/util/packaging.py
wowi42: Right, `upgrade_packages` is only consumed when `latest=True`, so the early `continue` on HELD only matters there. Closing.

wowi42:
Pushed a20ed4c2 addressing the multi-version question. `PackageInfo` now carries every installed version in a sorted tuple, with `installed_version` as a "highest" convenience accessor. `build_package_map` sorts naturally so `5.10 < 6.1`. New tests in `test_facts_packages.py` and `test_operations_utils.py` cover the rpm-installonly multi-version case end to end.

src/pyinfra/facts/util/packages.py
```diff
@@ -0,0 +1,98 @@
...
+    ``installed_versions`` holds every installed version, sorted ascending so
+    the highest is last. Most package managers only ever install one version
+    of a package, but rpm-family installonly packages (kernels), portage
+    SLOTs and dpkg multi-arch can produce more than one. The
+    :pyattr:`installed_version` property returns the highest installed
+    version (or ``""`` when none are reported).
+    """
+
+    name: str
+    installed_versions: tuple[str, ...] = ()
```
DonDebonair, line 52: Why `tuple` and not `set`?

src/pyinfra/facts/util/packages.py
```diff
@@ -0,0 +1,98 @@
...
+    available_version: str | None = None
+    status: PackageStatus = PackageStatus.INSTALLED
+
+    @property
+    def installed_version(self) -> str:
```
DonDebonair: Maybe this should return `str | None`?

src/pyinfra/facts/util/packages.py
```diff
@@ -0,0 +1,98 @@
...
+
+    INSTALLED = "installed"
+    UPGRADEABLE = "upgradeable"
+    HELD = "held"
+
+
+_VERSION_PART_RE = re.compile(r"\d+|\D+")
+
+
+def _version_sort_key(version: str) -> tuple[tuple[int, int | str], ...]:
```
DonDebonair, line 25: I would like to see a test for this. I know it's implicitly tested through other tests, but a dedicated test helps understand the logic.

DonDebonair (changes requested):
Left 2 more comments/requests.

Also, ideally, I assume that as part of the grand plan, we migrate all package manager facts and ops to use `PackageInfo` right? And will we then get rid of `dict[str, set[str]]` everywhere?

src/pyinfra/facts/util/packages.py
wowi42: Tuple on purpose, for three reasons:

1. Order matters. `build_package_map` sorts versions with `_version_sort_key` so `installed_version` can return the highest as `[-1]`. A `set` has no order (string iteration order is randomized per run), so that breaks.
2. `PackageInfo` is `@dataclass(frozen=True)`. A frozen dataclass should be immutable and hashable. `tuple` is both; `set` is mutable and unhashable, so `hash(PackageInfo(...))` would raise `TypeError`. `frozenset` is hashable but still unordered, so it loses reason 1.
3. Tuple equality stays stable and order-sensitive, which keeps the dataclass `==` checks in the tests predictable.

Dedup is not a reason to use a set here: the input is already `set[str]` and is deduped there before being sorted into the tuple. No code change needed; I can add a short comment on the field if that helps.

src/pyinfra/facts/util/packages.py
wowi42: Agreed. Fixed in 43ddefdf.

`installed_version` now returns `str | None` and yields `None` (not `""`) when no versions are reported. An empty string is never a valid package version, and this matches the project convention of distinguishing unset from empty. Docstring and tests updated. No production code reads the property yet, so the contract change is safe.

src/pyinfra/facts/util/packages.py
wowi42: Added in bca1fbe4. There is now a dedicated `TestVersionSortKey` suite that exercises `_version_sort_key` directly, including the `5.2` vs `5.10` case from the docstring and the `9.0` vs `9.0-1` revision case.

While adding it I found the existing `test_multiple_versions_sorted_natural_order` used data (`{6.1.0-13, 6.1.0-12, 5.10.0-26}`) that sorts identically under plain lexicographic sort, so it did not actually prove natural ordering. I switched it to discriminating data and added a second `build_package_map` case that fails under lexicographic order.

wowi42:
Pushed two commits addressing the open threads:

- `43ddefdf`: `installed_version` returns `str | None` instead of `""` (r3247665140).
- `bca1fbe4`: dedicated `_version_sort_key` tests plus a stronger `build_package_map` natural-order test (r3247693849).

Replied on r3247627819 explaining why `installed_versions` is a tuple, not a set; no code change there.

Tests and lint pass locally: `pytest tests/test_facts_packages.py tests/test_operations_utils.py`, `ruff check`, `ruff format --check`, `mypy`.

On the roadmap question: yes. This PR is phase 1 of #1725. It only adds the model and the dual-format bridge in `ensure_packages`; no package-manager fact or operation is migrated yet, which is why every caller still returns `dict[str, set[str]]`. Later phases migrate facts and ops to return `PackageInfo`, and once nothing returns the legacy shape we drop `dict[str, set[str]]` support from `ensure_packages`. Keeping both shapes working now is deliberate so the migration can land incrementally without breaking external connectors and operations.

src/pyinfra/facts/util/packages.py
DonDebonair: Fair, what about `list`?

Fizzadar (approved):
This is awesome, @wowi42! LGTM

DonDebonair (approved):

pirate:
woohoo! awesome work guys! thanks for building this!

PR #1730, opened 2026-05-07
fix(facts.iptables): requires_command() error in when table specified

doug-fitzmaurice-rowden:
`command()` supports the `table` argument, but the same args are passed to `requires_command()`, which was throwing an error because it did not expect `table`.

This deploy:
```
iptables.rule(
    name="Masquerade traffic",
    table="nat",
    chain="POSTROUTING",
    source="10.0.0.0/24",
    out_interface="eth0",
    jump="MASQUERADE",
    _sudo=True,
)
```
Would result in this error:
```
TypeError: IptablesRules.requires_command() got an unexpected keyword argument 'table'
```

This change brings the fact in line with others by accepting args and kwargs.

### DIFF
--- a/src/pyinfra/facts/iptables.py
+++ b/src/pyinfra/facts/iptables.py
@@ -92,7 +92,7 @@ class IptablesRules(FactBase):
     default = list
 
     @override
-    def requires_command(self) -> str:
+    def requires_command(self, *args, **kwargs) -> str:
         return "iptables-save"
 
     @override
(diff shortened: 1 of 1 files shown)


src/pyinfra/facts/iptables.py
Fizzadar:
```suggestion
    def requires_command(self, *args, **kwargs) -> str:
```

Fizzadar:
```suggestion
    def requires_command(self, *args, **kwargs) -> str:
```

Fizzadar (approved):
Good spot, ty for fixing @doug-fitzmaurice-rowden. Added two more places w/same issue.

Fizzadar:
No idea what GitHub has done with those suggestions, added directly

PR #1733, opened 2026-05-07
fix(operations): handle output from the 'uutils' version of stat (eg. recent Ubuntu)

egnor:
On systems with the "uutils" version of the "stat" utility (eg. recent Ubuntu), the output is quoted differently, which breaks file-stat checks and leads to unnecessary changes to files. This fixes that by recognizing different styles of quoting.

- [x] Tests pass (see `scripts/dev-test.sh`) - modulo what seems to be a pre-existing failure in apt.packages.test_apt_packages_update_cached

### DIFF
--- a/src/pyinfra/facts/files.py
+++ b/src/pyinfra/facts/files.py
@@ -67,6 +67,21 @@ CHAR_TO_PERMISSION = (
 )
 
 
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Neat fix, LGTM! Thank you @egnor

PR #1689, opened 2026-04-21
fix(connectors.local): fragile sudo handling resulting in faillocks and premature errors.

Dexmachi:
Closes #1043
Closes #1688 

# How:
Added a {0}.{PPID}.called marker + env LC_ALL=C to sudo calls, allowing Pyinfra to keep track of sudo call attempts and to have a standardized retry catch.

the marker gets removed alongside the sudo password file, in the same call.

# Why:
@local sudo handling was fragile, requiring a specific locale and not dealing with auto sudo retries, resulting in faillocks and premature errors.

# Test Changes:
Had to change test_ssh and test_util to assert "env LC_ALL=C " at the start of sudo and su calls.

# Before/After:

## Before(normal usage):
```bash
/opt/***-app/venv/bin/pyinfra @local playbooks/install1pkg/deploy.py
    Attempting to work in a virtualenv.
    If you encounter problems, please install pyinfra inside the virtualenv.


--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation files...
    Loading: playbooks/install1pkg/deploy.py
    [@local] Ready: playbooks/install1pkg/deploy.py

--> Detected changes:
    Operation                             Change       Conditional Change   
    server.shell (commands=['echo Hy'])   1 (@local)   -                    

    Detected changes may not include every change pyinfra will execute.
    Hidden side effects of operations may alter behaviour of future operations,
    this will be shown in the results. The remote state will always be updated
    to reflect the state defined by the input operations.

    Detected changes displayed above, skip this step with -y
                             
--> Beginning operation run...
--> Starting operation: server.shell (commands=['echo Hy'])
Sinto muito, tente novamente.
Sinto muito, tente novamente.
    [@local] sudo: 3 tentativas de senha incorreta
    [@local] Error: executed 0 commands
```

## After(normal usage):
```bash
pyinfra @local playbooks/install1pkg/deploy.py -y
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@local] Connected

--> Preparing operation files...
    Loading: playbooks/install1pkg/deploy.py
    [@local] Ready: playbooks/install1pkg/deploy.py

--> Skipping change detection
--> Beginning operation run...
--> Starting operation: server.shell (commands=['echo Hy'])
Sorry, try again.
    [@local] sudo: 1 incorrect password attempt
    [@local] sudo: no password was provided
    [@local] sudo: 1 incorrect password attempt
[...]

Issue #1499, opened 2025-11-09
`sudo-rs` not supported


alessio-perugini:
## Describe the bug

I don't think it's a bug of pyinfra, but I wanted to report here in case other users have the same issue.
Ubuntu 25.10 changed default implementation of sudo to `sudo-rs`. However it seems that there are some issues when pyinfra require sudo to prompt user password interactevely failing with:

```
--> Preparing operation files...
    Loading: deploy.py
    [@local] sudo-rs: interactive authentication is required
    [@local] Error: could not load fact: server.TmpDir
```

I've tested then to switch to old sudo implementation and everything work smooth.

## To Reproduce
Use sudo-rs, and run a pyinfra command with __Sudo 

## Expected behavior
Same as `sudo`

Fizzadar:
Unfortunately `sudo-rs` is not currently supported because they didn't implementing a bunch of commonly used flags/etc for reasons I don't understand: 

- https://github.com/trifectatechfoundation/sudo-rs/issues/1299
- https://github.com/trifectatechfoundation/sudo-rs/issues/567

And the biggest one of all is askpass support, which is 100% required for pyinfra - looks like they realized their mistake due to the number of reports! So support should be landing soon ([PR](https://github.com/trifectatechfoundation/sudo-rs/pull/1354)):

- https://github.com/trifectatechfoundation/sudo-rs/issues/1249

---

Until then - workaround is to switch back to classic sudo, workaround here: https://github.com/trifectatechfoundation/sudo-rs/issues/1299#issuecomment-3567268773

ginabythebay:
I just got bitten by this one and ended up switching the affected system to use the old sudo for now.

I wonder if there could be a command line option added to pyinfra that would allow users to choose `sudo.ws` rather than `sudo`?

I am not convinced at this point that the sudo-ws folks are treating this as urgently as I would like.

mixmastamyk:
Hi, 

Apparently v0.2.11 of sudo-rs was shipped last December:  https://github.com/trifectatechfoundation/sudo-rs/releases/tag/v0.2.11

Am trying to configure Ubuntu 26.04, that ships 0.2.13.  
Yet pyinfra is still saying: 

    [...] sudo: interactive authentication is required

Would appreciate this getting fixed, we're going to be doing a lot of this soon.  Cheers,

0cjs:
I'm not sure that we can count on `-E` working anyway; the security policy may not allow it. Similar for $HOME.

[...]

PR #1734, opened 2026-05-07
fix(operations): handle password prompt from sudo-rs (eg. recent Ubuntu)

egnor:
On systems using `sudo-rs` (eg. recent Ubuntu), sudo password prompts take a slightly different format, which breaks the password prompt spotting logic. (this bit Ansible here: https://github.com/ansible/ansible/issues/85837) This PR fixes that by recognizing both password prompt strings.

This should PARTIALLY address #1499, at least enough to limp forward even if that version of `sudo` doesn't take all the same flags.

This is adjacent to #1689 - it's orthogonal, but touches the same code, and whichever lands second will need a rebase.

- [x] Tests pass (see `scripts/dev-test.sh`) (except for pre-existing `apt.packages.test_apt_packages_update_cached`)


Fizzadar (approved):
Simple quick fix, LGTM. #1689 bundles other changes so will land this now and rebase there. Thanks @egnor!

PR #1735, opened 2026-05-08
feat: Add server.Uptime fact

Tbruno25:
This PR adds the `server.Uptime` fact.

#1708 for reference.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [x] Pull request includes tests for any new/updated operations/facts
- [ ] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`)
- [x] Pull request title follows the 
      [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format

PR #1737, opened 2026-05-08
fix: Use Uptime fact in server.reboot

Tbruno25:
This PR utilizes server.Uptime to verify the server has successfully rebooted.

#1708 for reference
Depends on #1735

### DIFF
--- a/src/pyinfra/operations/server.py
+++ b/src/pyinfra/operations/server.py
@@ -114,20 +114,21 @@ def reboot(delay=10, interval=1, reboot_timeout=300):
         while True:
             host.connect(show_errors=False)
 
-            if not host.connected:
+            if host.connected:
+                post_uptime = host.get_fact(Uptime)
+                logger.debug(
+                    "Connected (current_uptime=%ss, pre_reboot_uptime=%ss)",
+                    post_uptime,
+                    pre_uptime,
+                )
+
+                if post_uptime < pre_uptime + delay:
+                    logger.debug("Reboot confirmed.")
+                    break
+
+                logger.debug("Host reachable but uptime unchanged; reboot still in progress")
+            else:
                 logger.debug("Waiting for host to become reachable...")
-                continue
-
-            post_uptime = host.get_fact(Uptime)
-            logger.debug(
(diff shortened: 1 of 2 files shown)


src/pyinfra/operations/server.py
```diff
@@ -103,11 +110,26 @@ def wait_and_reconnect(state, host):  # pragma: no cover
...
 
+        pre_uptime = pre_reboot_uptime[0]
+
         while True:
             host.connect(show_errors=False)
-            if host.connected:
+
+            if not host.connected:
+                logger.debug("Waiting for host to become reachable...")
+                continue
```
Fizzadar: This means if `host.connected` never becomes true we'll be stuck indefinitely since we explicitly skip the retries check below.

src/pyinfra/operations/server.py
```diff
@@ -103,11 +110,26 @@ def wait_and_reconnect(state, host):  # pragma: no cover
...
+
+            post_uptime = host.get_fact(Uptime)
+            logger.debug(
+                f"Connected (current_uptime={post_uptime}s, pre_reboot_uptime={pre_uptime}s)"
```
Fizzadar: Should use `%s` formatting here and extra args (so when debug not enabled we do no formatting work).

src/pyinfra/operations/server.py
```diff
@@ -103,11 +110,26 @@ def wait_and_reconnect(state, host):  # pragma: no cover
...
+            )
+
+            if post_uptime < pre_uptime + delay:
```
Fizzadar: I don't think the +delay here is necessary? (but harmless)

Fizzadar (changes requested):
Couple of tweaks needed!

src/pyinfra/operations/server.py
Tbruno25: great catch 🤦

src/pyinfra/operations/server.py
Tbruno25: It's to protect an edge case in which the server is immediately rebooted back to back.

Fizzadar (approved):
LGTM! Thanks for the quick turnaround @Tbruno25 🙏

PR #1739, opened 2026-05-10
fix(operations.server): handle OpenBSD useradd properly

epicrazzmatazz:
calling the `server.user` operation with `create_home=False` causes `useradd` to exit on OpenBSD as the binary does not support the `-M` flag. According to the manual page, the home directory is not created unless `-m` flag is present; no funky stuff like `/etc/login.defs` as far as I can tell. See https://man.openbsd.org/useradd and #1004

### DIFF
--- a/src/pyinfra/operations/server.py
+++ b/src/pyinfra/operations/server.py
@@ -996,9 +996,8 @@ def user(
 
         if create_home:
             args.append("-m")
-        elif os_type != "FreeBSD":
-            if os_type != "OpenBSD":
-                args.append("-M")
+        elif os_type not in ("FreeBSD", "OpenBSD"):
+            args.append("-M")
 
         if password and os_type != "FreeBSD":
             args.extend(["-p", QuoteString(password)])
--- a/src/pyinfra/operations/server.py
+++ b/src/pyinfra/operations/server.py
@@ -996,8 +996,9 @@ def user(
 
         if create_home:
             args.append("-m")
-        elif os_type not in {"FreeBSD", "OpenBSD"}:
-            args.append("-M")
+        elif os_type != "FreeBSD":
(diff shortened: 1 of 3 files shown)


wowi42:
Thanks for the fix, the OpenBSD useradd behaviour matches the man page. Two small suggestions before merge:

1. Flatten the nested condition to keep the surrounding style:

   ```python
   elif os_type not in ("FreeBSD", "OpenBSD"):
       args.append("-M")
   ```

2. Add a fixture under `tests/operations/server.user/` covering OpenBSD (the path is currently untested). Per AGENTS.md the convention is YAML/JSON fixtures, not Python tests. Something like:

   ```json
   {
       "args": ["someuser"],
       "kwargs": {"create_home": false},
       "facts": {
           "server.Os": "OpenBSD",
           "server.Users": {},
           "server.Groups": {},
           "server.Which": {"command=useradd": "/usr/sbin/useradd"}
       },
       "commands": ["useradd someuser"]
   }
   ```

   A companion `create_home: true` fixture would also be cheap and lock in that `-m` is still emitted on OpenBSD.

Note: the explicit `OpenBSD` check (rather than a broad `BSD`) is correct - NetBSD's useradd does support `-M`.

Fizzadar (approved):
LGTM, thanks @epicrazzmatazz!

epicrazzmatazz:
For future contributors: AFAICT the `server.user` operation may not work on NetBSD. NetBSD `useradd -M` has a different meaning than on Linux. On NetBSD, `-M` specifies the permission bits for the created home directory, while on Linux `-M` is a flag to override `/etc/login.defs`.

wowi42:
@epicrazzmatazz Ah shit, I will have a look on it and setup a NetBSD in the pyinfra testing farm

PR #1745, opened 2026-05-12
feat: Add `fetch_tags` flag to `git.repo`

inducer:
In my use case, I have an existing git repository that I would like to update to a tag that does not yet exist in the local checkout. Right now this fails because `git fetch` does not pick up new tags, and the `GitTag` fact check fails to realize the new revision is a tag (since it doesn't exist). Using `git fetch --tags` makes sure that the tag is there before its existence is checked.

- [x] Tests pass (see `scripts/dev-test.sh`) (

### DIFF
--- a/src/pyinfra/operations/git.py
+++ b/src/pyinfra/operations/git.py
@@ -93,14 +93,16 @@ def repo(
     src: str,
     dest: str,
     branch: str | None = None,
-    pull=True,
-    rebase=False,
+    pull: bool = True,
+    rebase: bool = False,
     user: str | None = None,
     group: str | None = None,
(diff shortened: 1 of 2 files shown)


DonDebonair:
Thanks for this PR! This changes the behavior of `git.repo` though. Alternative: add a boolean parameter `tags` that defaults to `False`, so previous behavior remains unchanged.

DonDebonair:
@inducer if you make the suggested change, we can merge this

inducer:
How's this?

Fizzadar (approved):
LGTM, thanks @inducer!

PR #1750, opened 2026-05-14
feat(connectors.ssh): auto-load OpenSSH user certificates from ssh_config

### LINKED ISSUE #1569: pyinfra seems to have trouble using openssh CA signed user certificates
I've setup a user pyinfra for all hosts using certificate based access with on openssh CA.
Although I've informed pyinfra inventory of all the required field still I need to create entries for all hosts in ssh/config with the specific key and signed certificate files for the pyinfra user. 

I;ll solve this by moving pyinfra from venv to container. However, when using venv in a sudoer account on the control client it creates a lot of administrative overhead in the ssh/config (double entries for normal user account and pyinfra user like myhost and myhost-pyinfra entries.

Whatever works for ssh from CLI should work when running pyinfra.
So if this works 'ssh pyinfra@host' with openssh CA signed host/user certificates then pyinfra running in venv should just be able to work without any overhead required (or just adding the info in the inventory.py)

wowi42:
## Summary

Make OpenSSH CA-signed user certificates work without the inventory-level workaround documented in #1569. When `~/.ssh/config` supplies `IdentityFile` (and optionally `CertificateFile`), pyinfra now pre-resolves the private key, attaches the certificate via `paramiko.PKey.load_certificate`, and passes the key as `pkey` instead of relying on paramiko's `key_filename` flow (which cannot attach an OpenSSH cert on its own).

Other fixes folded into the same change because they sit on the cert-loading path:

- `get_private_key` failed to find the adjacent `-cert.pub` for `~`-prefixed `ssh_key` values; the implicit lookup is now anchored on the resolved key path.
- The implicit lookup also tried `<key>.pub` after `<key>-cert.pub`. `<key>.pub` is the bare public key, not a certificate, and `load_certificate` happily overwrites any previously-attached cert with a non-cert public blob. Dropped.

The cert-attaching logic now lives in a state-free helper, `load_key_with_certificate(key_filename, key_password=None, certificate_filename=None, cwd=None)`. `get_private_key` keeps the existing per-state caching by wrapping it; sshuserclient calls it directly so no pyinfra state is leaked into the paramiko-extension layer.

Closes #1569

## Test plan

- [x] New session-scoped pytest fixture in `tests/test_connectors/conftest.py` that uses `ssh-keygen` to generate a CA, a user ed25519 key, and a CA-signed `-cert.pub`. Skips automatically if `ssh-keygen` is missing.
- [x] New `tests/test_connectors/test_ssh_util.py` covers: adjacent cert attaches, explicit `certificate_filename` honoured, key with no cert returns the bare key, `~`-prefixed `ssh_key` regression, missing key raises.
- [x] New tests in `tests/test_connectors/test_sshuserclient.py` covering ssh_config `IdentityFile` + adjacent cert, ssh_config `CertificateFile` overriding the implicit path, and the legacy fall-through when `IdentityFile` does not exist on disk.
- [x] Three existing mocked assertions in `tests/test_connectors/test_ssh.py` updated to expect `testkey-cert.pub` rather than the dropped `testkey.pub` fallback.
- [x] `uv run pytest tests/test_connectors/` — 187 passed.
- [x] `scripts/dev-lint.sh` — clean (ruff, ruff format, mypy, arguments sync).

## Requirements

- [x] Pull request includes documentation for any new/updated operations/facts (no public-API surface added; behaviour matches OpenSSH semantics for `IdentityFile` and `CertificateFile`)

### DIFF
--- a/src/pyinfra/connectors/ssh_util.py
+++ b/src/pyinfra/connectors/ssh_util.py
@@ -64,48 +64,81 @@ def _load_private_key_file(filename: str, key_filename: str, key_password: str):
     raise exception
 
 
-def get_private_key(state: "State", key_filename: str, key_password: str) -> PKey:
-    if key_filename in state.private_keys:
-        return state.private_keys[key_filename]
+def _resolve_key_paths(key_filename: str, cwd: str | None = None) -> list[str]:
+    candidates = [path.expanduser(key_filename)]
+    if cwd:
+        candidates.append(path.join(cwd, path.expanduser(key_filename)))
+    return candidates
 
-    ssh_key_filenames = [
-        # Global from executed directory
-        path.expanduser(key_filename),
-    ]
 
-    if state.cwd:
-        # Relative to the CWD
-        path.join(state.cwd, key_filename)
+def load_key_with_certificate(
+    key_filename: str,
+    key_password: str | None = None,
+    certificate_filename: str | None = None,
+    cwd: str | None = None,
+) -> PKey:
+    """
+    Load a paramiko ``PKey`` from disk and attach an OpenSSH certificate when one
+    is available. Has no dependency on pyinfra state so it can be reused both
+    from the CLI key path (via :func:`get_private_key`) and from the ssh_config
+    path in :mod:`pyinfra.connectors.sshuserclient`.
 
-    key = None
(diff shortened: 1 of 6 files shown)


Fizzadar (approved):
LGTM, much improved!

PR #1753, opened 2026-05-15
refactor: replace os.path usage with pathlib

wowi42:
## Summary

Replace `os.path` and `from os import path` call sites with `pathlib` across the library (`api`, `connectors`, `facts`, `operations`) and the CLI.

The migration is not a blanket `os.path` to `pathlib.Path` swap. Path strings in pyinfra serve three different roles, and each needs a different replacement:

- **Genuinely remote / POSIX paths** use `PurePosixPath` / `posixpath`: `files.move`/`copy`, `_raise_or_remove_invalid_path`, the `files.sync` walk loop, the scp client, `LinuxDistribution` release-file parsing, and the deploy-dir relative joins in `files.put`/`get`/`template` and `server` authorized-keys handling. The remote target is always unix, so these must not depend on the controller platform. This also fixes a latent bug where `os.path.join` on a Windows controller would emit backslashes into remote unix paths.
- **Real local controller paths** use the platform-aware `pathlib.Path`: `get_file_path` resolves `state.cwd` (from `getcwd()`) and `current_exec_filename` (an actual local file). On Windows these are native backslash paths. `pathlib.Path` is itself the platform branch (`WindowsPath` on Windows, `PosixPath` elsewhere), so it parses them correctly, whereas forcing posix semantics produced empty dirnames and broke nested deploy/task includes.
- **Local filesystem predicates** use `Path` methods: `exists`, `is_file`, `is_dir`, `is_symlink`, `mkdir`, `samefile`, `readlink`.

### Intentional non-changes

- `os.path.relpath` is retained in `api/util.py` and `pyinfra_cli/cli.py`: `pathlib` has no non-strict relpath on Python 3.10 (`Path.relative_to(walk_up=True)` is 3.12+), and swapping it would silently change behaviour for paths without a common prefix.
- `os.path.normcase` in `virtualenv.py` was dropped (no `pathlib` equivalent); venv detection is now case-sensitive, a minor behaviour change on Windows only.

### Tests

- `tests/util.py` adds `Path` method patches alongside the existing `os.path.*` patches; the latter are kept because some modules still reach paths via legacy callers during a run.
- Connector tests patch the shared `posixpath` module (or `Path` methods) instead of the old `path.X` aliases. `from os import path` previously made `client.path` and `config.path` the same module object, so one patch covered both; the equivalent is restored by patching `posixpath`.

### Key lesson

The original `from os import path` was doing double duty: on Windows `os.path` is `ntpath`, which transparently accepts both `/` and `\`. A single `os.path` call site could therefore be handling a real local Windows path and a pyinfra POSIX-style path interchangeably. Neither `pathlib.Path` (normalises separators to `\` on Windows) nor `posixpath` (cannot parse native Windows paths) reproduces that on its own. The migration is only correct once each site is classified as remote-POSIX, real-local, or a filesystem predicate, and given the matching tool.

## Requirements

- [x] Pull request includes documentation for any new/updated operations/facts (behavioural change only, no surface change)


src/pyinfra/operations/server.py
```diff
@@ -930,9 +931,9 @@ def user_authorized_keys(
     def read_any_pub_key_file(key):
         try_path = key
         if state.cwd:
-            try_path = path.join(state.cwd, key)
+            try_path = posixpath.join(state.cwd, key)
```
Fizzadar: This is a local path so forcing posix here will break windows.

src/pyinfra/operations/files.py
```diff
@@ -690,54 +690,48 @@ def sync(
...
-    for dirpath, dirnames, filenames in os.walk(src, topdown=True, followlinks=False):
-        remote_dirpath = Path(os.path.normpath(os.path.relpath(dirpath, src))).as_posix()
+    for dirpath, dirnames, filenames in os.walk(src_path, topdown=True, followlinks=False):
+        dirpath_path = Path(dirpath)
+        remote_dirpath = dirpath_path.relative_to(src_path).as_posix()
 
         # Filter excluded dirs and detect directory symlinks
         for child_dir in dirnames[:]:
-            child_path = os.path.normpath(os.path.join(remote_dirpath, child_dir))
+            child_path = (PurePosixPath(remote_dirpath) / child_dir).as_posix()
```
Fizzadar: Local path, should not be assume posix.

src/pyinfra/operations/files.py
```diff
@@ -919,12 +913,12 @@ def get(
     """
 
     if add_deploy_dir and state.cwd:
-        dest = os.path.join(state.cwd, dest)
+        dest = posixpath.join(state.cwd, dest)
```
Fizzadar: Another posix-on-windows issue

src/pyinfra/operations/files.py
```diff
@@ -1109,19 +1103,19 @@ def put(
         assert isinstance(src, (str, Path))
         # Add deploy directory?
         if add_deploy_dir and state.cwd:
-            src = os.path.join(state.cwd, src)
+            src = posixpath.join(state.cwd, src)
```
Fizzadar: Another posix-on-windows

src/pyinfra/operations/files.py
```diff
@@ -1419,7 +1413,7 @@ def template(
     '''
 
     if not hasattr(src, "read") and state.cwd:
-        src = os.path.join(state.cwd, src)
+        src = posixpath.join(state.cwd, src)
```
Fizzadar: More posix-on-windows, two more below

src/pyinfra/connectors/sshuserclient/client.py
```diff
@@ -146,9 +147,9 @@ def get_ssh_config(user_config_file=None):
     logger.debug("Loading SSH config: %s", user_config_file)
 
     if user_config_file is None:
-        user_config_file = path.expanduser("~/.ssh/config")
+        user_config_file = posixpath.expanduser("~/.ssh/config")
```
Fizzadar: Another here

Fizzadar (changes requested):
There's quite a lot of inconsistency in here with `posixpath` being used in many local path locations which will break Windows absolute paths and similar.

It might be cleaner to break this into two PRs: one to handle local paths and one for remote paths (selfishly this would make reviewing much easier too :)).

wowi42:
Conflicts with `3.x` resolved (the `tests/util.py` harness moved to `pyinfra-testing`, so the `Path` method patches now live in `tests/conftest.py` wrapping `pyinfra_testing`'s `patch_files`).

All the posix-on-local-path comments are addressed in c8b99a2a:
- `server.py`, `files.get`/`put`/`template`: deploy-cwd joins now use `pathlib.Path`
- `files.sync`: the `exclude_dir` match path is built from local `Path` objects and only converted to posix for remote-style matching
- `sshuserclient` client + config: back to `os.path` for local ssh config/known_hosts paths

Two notes on the flagged lines I did *not* change:
- `files.move`/`files.copy` (the "two more below"): both `src` and `dest` are **remote** paths there, so `posixpath`/`PurePosixPath` is intentional — same for `posixpath.dirname` in `download` and the scp client's remote path handling.
- `os.path.relpath` stays in `api/util.py`/`pyinfra_cli/cli.py` since `pathlib` has no non-strict relpath on Python 3.10.

Happy to split into local-path vs remote-path PRs if you'd still prefer that — the commits are separated so it should be easy to carve up.

Fizzadar (approved):
Wonderful - LGTM! Thank you @wowi42!

Issue #1756, opened 2026-05-16
--limit <group> still merges non-selected groups' data, causing host data overwrites


etiennecollin:
## Describe the bug

When using `--limit <group_name>`, host data from other groups is still merged into the inventory before filtering occurs. As a result, if the same host exists in multiple groups, the last declared group's data overwrites the others for conflicting keys, even when that group is not part of the limit selection.

This makes it impossible to safely define environment or role-specific host data for the same host across multiple groups when relying on `--limit`.

For example, given two groups containing the same host with different values for `important_data`, running with `--limit group_a` still results in the value from `group_b` being applied if `group_b` is declared later in the inventory.

## To Reproduce

Sometimes, a user might want to define different groups which assign different data to the same host based on the group using the `--limit <group_name>` flag. For example, see this sample inventory file:

```py
group_a = [
    (
        "@local",
        {
            "important_data": "foo"
        }, # Some data, let's call it data_a
    )
]

group_b = [
    (
        "@local",
        {
            "important_data": "bar"
        }, # Some other data, let's call it data_b
    )
]
```

In this case deploying using `--limit group_a` will use `data_b`, because `data_b` overwrites `data_a`. In other words, if we get the `important_data`, we get `bar` instead of the expected `foo`.

Here is a simple partial stack trace of the overwriting behaviour when running `pyinfra inventory.py deploy.py --limit group_a`:

- `pyinfra_cli/cli.py:383` where `inventory = make_inventory(` is called
- `pyinfra_cli/inventory.py:206` where `return make_inventory_from_files(inventory, override_data, cwd, group_data_directories)` is called
- `pyinfra_cli/inventory.py:362` where `fake_inventory = Inventory((all_hosts, all_data), **fake_groups)` is called
- `pyinfra/api/inventory.py:58` where `self.make_hosts_and_groups(names, groups)` is called
- `pyinfra/api/inventory.py:75` where `name_to_data[name].update(data)` is called

In this case, if the same key is used across multiple groups for the same host, the value for the key is overwritten to the last declared value. Later in the `make_hosts_and_groups` function, `host_data` is created with `host_data = name_to_data[name]`and  is used to fill `self.host_data[sub_name] = sub_data` which is used to initialize the `Host` objects.

## Expected behavior

When using `--limit`, only the selected groups should contribute host data during inventory construction. In the example above, running:

```sh
pyinfra inventory.py deploy.py --limit group_a
```

should result in:

```py
[...]

PR #1757, opened 2026-05-16
fix: limit-aware inventory creation

etiennecollin:
## Description

This draft PR is a proof of concept for #1756.

The current issue is that `--limit <group>` still allows host data from non-selected groups to be merged during inventory construction. If the same host exists in multiple groups with overlapping keys, values from later groups overwrite those from the selected group.

This implementation explores passing the `limit` iterable into the `Inventory` object so that host data from non-selected groups can be ignored during inventory construction.

The goal here is mainly to validate the approach; I do not expect this implementation to be merged as-is, since it was written fairly quickly and may not align with the intended architecture of the project. It may serve as inspiration for a proper implementation.

## Testing

The included tests pass with `scripts/dev-test.sh`, ~but I encountered the following error in one of my real deployments, which does not occur on pyinfra `v3.8.0`. So there is likely at least one edge case or unintended side effect in the current implementation.~ The error also happens on 3.8.0, so it might be my deployment that is flawed.

```
--> pyinfra error: Cycle detected in operation ordering DAG.
    Error: ('nodes are in a cycle', ['6122ecd5c8bf5453cf3cefd3b6786f892d5d01af', 'af7a299a1fa7001b19dc57f7f9a1249385411c40-0', 'b2c4a8bab430d883f638396faaf8251b721a058d-0', '0ae435d43f8108d80868fdf5c2a700ed581f139f-0', 'eccb013008ef20c412778a9b62f4dfbf3d7013f2-0', 'accace0f63f34b5b36b28d67ad44597a71d21803', '6122ecd5c8bf5453cf3cefd3b6786f892d5d01af'])

    This can happen when using loops in operation code, please see: https://docs.pyinfra.com/en/latest/deploy-process.html#loops-cycle-errors
```

## Checklist

### DIFF
--- a/src/pyinfra/api/inventory.py
+++ b/src/pyinfra/api/inventory.py
@@ -31,6 +31,7 @@ class Inventory:
     Args:
         names_data: tuple of ``(names, data)``
         override_data: dictionary of data overrides
+        limit: Iterable, restrict host data loading by name and group name
         ssh_*: deprecated, use ``override_data.ssh_*``
         winrm_*: deprecated, use ``override_data.winrm_*``
         **groups: map of group name -> ``(names, data)``
@@ -43,7 +44,7 @@ class Inventory:
     def empty():
         return Inventory(([], {}))
 
-    def __init__(self, names_data, override_data=None, **groups):
+    def __init__(self, names_data, override_data=None, limit=None, **groups):
         # Setup basics
         self.groups = defaultdict(list)  # lists of Host objects
         self.host_data: dict[str, dict] = defaultdict(dict)  # dict of name -> data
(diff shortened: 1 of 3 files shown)


MikeSchincariol:
I have (had) this issue as well with my deployment inventory as well. In my use case, I need to be able to have servers by the same name in more than one group and the extra data passed to each server in each group differs. Even within the same group I will have the same server listed more than once.

For my situation, each server instance needs to be a distinct instance with its own data. I got around the current limitation by using a method to create each server instance, and the method injects a UUID'like value alongside the server name. Something like this:

```
class Server():

    server_uuid = count(0)

    def init(self, host:str):
        self.host = host

    def __call__(self, **kwargs):

        retval = (
                self.host + ":" +  f"{next(Server.server_uuid):04d}",
                {
                     ....properties...
                }
        )

        retval[1].update(**kwargs)
        return retval
```

But this is also somewhat inelegant because the UUID appears in the log and can be confusing (although for my use it hasn't been an issue)

One solution could be that Pyinfra implements and respects a uniqueness ID alongside the other host data such that the host data is kept in sync with the specific host it was applied to. But I don't have pseudo code to show this.

etiennecollin:
Comment moved to https://github.com/pyinfra-dev/pyinfra/issues/1756#issuecomment-4529637424 for better traceability.

Fizzadar (changes requested):
Replied in #1756 just adding the review for tracking PRs :)

PR #1709, opened 2026-04-29
feat(api): New MaskString

gwelch-contegix:
This implements a new MaskString type which preserves the masking of the string in most contexts.
This does not attempt to fix an un-masked string showing up in 'diffed' output.

This makes it so that when MaskString is used in an operation argument it will show up masked instead of with the real value.
This also updates the masked value to be `*MASKED*` and updates all current usage to output valid commands.
Previously the unquoted `***` value would show up in commands during verbose output with no context as to why it was there.

This MaskString implementation attemps to preserve normal string operations as much as possible and when it can't it only operates on the masked value (by default `*MASKED*`) see the tests for how it interacts with the `str` class on standard operations.

As this would be a breaking change for custom usage of MaskString it would may make sense to add this as another name and deprecate `MaskString` instead of replacing it directly.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [x] Pull request includes tests for any new/updated operations/facts
- [ ] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`)

Note that AI was used to generate part of the tests that's why some of the values chosen change between tests.

Fixes #1148

PR #1759, opened 2026-05-18
fix(operations.files): shell-quote files.block content

wowi42:
## Summary

`files.block` passed the block content to `awk` as a positional argument wrapped in raw quote characters via string concatenation instead of proper shell quoting. Any content containing double quotes, single quotes, `$(...)` or backticks broke the generated command.

Reproduction (from upstream issue):

```python
files.block(path="/root/.bashrc", content='eval "$(zoxide init --cmd cd bash)"')
```

The create path used a heredoc, so the first run succeeded; the second run (update existing block) failed with `sh: Syntax error: "(" unexpected`. The line-match path was affected the same way. Unquoted content was also a shell-injection vector.

## Fix

- Update and line-match paths now pass content through `QuoteString` (`shlex.quote`).
- The create-path heredoc delimiter is quoted (`<<'PYINFRAHERE'`) so the body is literal too.
- Content is always written verbatim and is never expanded on the remote host. This also removes the previous limitation where content containing raw single quotes could break.
- `try_prevent_shell_expansion` no longer affects quoting (content is always literal); it is kept and accepted for backwards compatibility and documented as deprecated.

## Tests

- 3 new YAML fixtures reproducing the issue across the update, line-match and no-file code paths with `"`, `'` and `$(...)` in the content (written test-first, confirmed failing before the fix).
- 7 pre-existing fixtures that encoded the old unsafe quoting updated to the corrected, shell-safe output.
- Full suite passes (`1737 passed`); `scripts/dev-lint.sh` clean (ruff, format, mypy, arguments-sync).

Fixes the bug reported : https://github.com/pyinfra-dev/pyinfra/issues/1758

## Requirements

- [x] Pull request includes documentation for any new/updated operations/facts (behavioural change only, no surface change)

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -1930,7 +1930,8 @@ def block(
     + line: regex before or after which the content should be added if it doesn't exist.
     + backup: whether to backup the file (see ``files.line``). Default False.
     + escape_regex_characters: whether to escape regex characters from the matching line
-    + try_prevent_shell_expansion: tries to prevent shell expanding by values like `$`
+    + try_prevent_shell_expansion: deprecated and ignored; ``content`` is always written
+      literally (no shell expansion) and is safely shell-quoted
     + marker: the base string used to mark the text.  Default is ``# {mark} PYINFRA BLOCK``
     + begin: the value for ``{mark}`` in the marker before the content. Default is ``BEGIN``
     + end: the value for ``{mark}`` in the marker after the content. Default is ``END``
@@ -1950,8 +1951,9 @@ def block(
 
     Removal ignores ``content`` and ``line``
 
-    Preventing shell expansion works by wrapping the content in '`' before passing to `awk`.
-    WARNING: This will break if the content contains raw single quotes.
+    ``content`` is written to the file verbatim. It is shell-quoted before being passed to
+    ``awk``, so shell metacharacters (``$(...)``, quotes, backticks) are kept literal and never
+    expanded on the remote host.
 
     **Examples:**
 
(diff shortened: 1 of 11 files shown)


src/pyinfra/operations/files.py
```diff
@@ -2109,7 +2103,7 @@ def block(
                 out_prep,
                 prog,
                 q_path,
-                the_block,
+                QuoteString(block_with_markers),
```
Fizzadar, line 2111: This needs to switch to `HiddenValue` from https://github.com/pyinfra-dev/pyinfra/pull/1709.

Fizzadar (changes requested):
LGTM just one change because it'll break since the other PR (I'd do it but I don't have permission :))

src/pyinfra/operations/files.py
wowi42: Could you clarify what should switch to `HiddenValue` here? A few things I am unsure about:

- The line you flagged is `QuoteString(block_with_markers)`. The PR does not use `MaskString` anywhere (neither did the pre-PR code at this spot, `the_block` was a plain `str`).
- After rebasing on `3.x` with #1709 merged, `QuoteString` still accepts `str | StringCommand | HiddenValue`, so the call as written keeps working. All `files.block` fixtures still pass and `dev-lint.sh` is clean.
- `HiddenValue` masks the value in command output. The block content here is user supplied config text (aliases, config snippets), not a secret, so masking it in verbose output seems like a behaviour change rather than a fix.

Do you want the block content wrapped as `QuoteString(HiddenValue(block_with_markers))` so it shows up masked in output, or did you mean a different change? Happy to apply whatever you intended once I am sure.

Fizzadar: Ignore me, I've no idea what I was smoking that day because this is the _second_ PR I have made this mistake on! `QuoteString` is correct; 🤦  for me

Fizzadar (approved):
Thanks @wowi42!

PR #1764, opened 2026-05-19
fix(operations.git): keep git.repo src in sync with config origin (#1763)

### LINKED ISSUE #1763: git.repo does not honor change of src
## Describe the bug
When changing `src` of [git.repo](https://docs.pyinfra.com/en/3.x/operations/git.html#git-repo) for an existing working copy, the new value of src is not assigned to the working copy's `origin`.

Consequently, after switching to a remote with newer commits, those are not present in the working copy.

## To Reproduce
```python
# File: git_repo_does_not_honor_change_of_src.py

from pyinfra.operations import git
from os import path

tmp_dir = "/tmp/git_repo_does_not_honor_change_of_src"
repo_original = "https://github.com/ansible/ansible"
repo_changed = "https://github.com/pyinfra-dev/pyinfra.git"

git.repo(dest=tmp_dir, src=repo_original)

git.repo(dest=tmp_dir, src=repo_changed)
```

```shell
# Run commands

uv run pyinfra @local git_repo_does_not_honor_change_of_src.py
(cd /tmp/git_repo_does_not_honor_change_of_src && git remote -v) # Still Ansible :(
```
## Expected behavior
The repository's `origin` is updated to `src`, with new commits visible. (Above example doesn't make sense as there is no common ancestry or set of files.)

## Meta
Pyinfra 3.8.

pascal-cm:
Fixes #1763

### DIFF
--- a/src/pyinfra/operations/git.py
+++ b/src/pyinfra/operations/git.py
@@ -160,6 +160,9 @@ def repo(
 
     # Ensuring existing repo
     else:
+        # Keep repository origin in sync with src, which may differ from last run
+        git_commands.append(StringCommand("remote", "set-url", "origin", QuoteString(src)))
+
         is_tag = False
         current_branch = host.get_fact(GitBranch, repo=dest)
         if branch is not None and current_branch != branch:
--- a/tests/operations/git.repo/branch_pull.json
+++ b/tests/operations/git.repo/branch_pull.json
@@ -25,6 +25,7 @@
         }
     },
     "commands": [
+        "cd /home/myrepo && git remote set-url origin myrepo",
         "cd /home/myrepo && git fetch",
         "cd /home/myrepo && git checkout mybranch",
         "cd /home/myrepo && git pull"
--- a/tests/operations/git.repo/branch_switch_up_to_date.json
+++ b/tests/operations/git.repo/branch_switch_up_to_date.json
@@ -25,6 +25,7 @@
         }
     },
     "commands": [
+        "cd /home/myrepo && git remote set-url origin myrepo",
         "cd /home/myrepo && git fetch",
         "cd /home/myrepo && git checkout mybranch"
     ]
--- a/tests/operations/git.repo/checkout_branch_needs_quotes.json
+++ b/tests/operations/git.repo/checkout_branch_needs_quotes.json
@@ -17,6 +17,7 @@
         }
     },
     "commands": [
+        "cd /home/myrepo && git remote set-url origin myrepo",
         "cd /home/myrepo && git fetch",
         "cd /home/myrepo && git checkout 'feature/evil; rm -rf /'"
     ]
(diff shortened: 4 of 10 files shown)


src/pyinfra/operations/git.py
```diff
@@ -160,6 +160,9 @@ def repo(
 
     # Ensuring existing repo
     else:
+        # Keep repository origin in sync with src, which may differ from last run
+        git_commands.append(StringCommand("set-url", "origin", QuoteString(src)))
```
DonDebonair: I think the proper command is `git remote set-url origin <src>`

pascal-cm: > I think the proper command is `git remote set-url origin <src>`

Your're absolutely correct. Fix coming. Thank you!

wowi42 (commented):
**1. Idempotency regression (main concern).** The `set-url` is emitted unconditionally for every existing repo on every run. pyinfra cannot know that re-setting the same URL is a no-op, so `git.repo` now always reports as changed/Success. This defeats the recently added skip-pull / "already up to date" no-op behavior. The smell is visible in the diff: `up_to_date.json` now has a non-empty `commands` list while still asserting `noop_description` (the operation both calls `host.noop()` and yields a command). The test harness checks the two independently, so it passes while the runtime behavior is contradictory.

The fix should be conditional. The existing `GitConfig` fact already exposes `remote.origin.url` (`git config --local -l`), which fits the fact-driven, idempotent style of the rest of the operation:

```python
git_config = host.get_fact(GitConfig, repo=dest)
current_origin = (git_config.get("remote.origin.url") or [None])[0]
if current_origin != src:
    git_commands.append(StringCommand("remote", "set-url", "origin", QuoteString(src)))
```

**2. Fails on repos without an `origin` remote.** `git remote set-url origin <url>` errors with "No such remote 'origin'" when origin does not exist (a checkout pyinfra did not clone, or one using a different remote name). `git.repo` previously worked on such adopted repos for branch/pull, so this is a robustness regression. Gating on the `GitConfig` fact (only emit when `remote.origin.url` exists and differs) also resolves this.

**3. Fact ordering vs skip-pull.** Facts are gathered before commands execute, so `GitRemoteBranchCommit` queries the old origin URL. When `src` genuinely changes, the skip-pull comparison can match the old remote tip and wrongly skip the pull, so #1763 is only partially fixed on that path. A conditional `set-url` does not fully solve this; ideally, when origin changes, skip-pull should be forced off.

**4. New fixture names look swapped and do not test the claimed cases.**
- `clone_src_origin_changed_url.json` uses `src="myrepo"` (the default the other fixtures use).
- `clone_src_origin_same_url.json` uses `src="changedrepo"`.

The names are inverted relative to content, and neither exercises the "origin already equals src" path (the implementation never skips). With a conditional fix, add a fixture where `GitConfig` reports `remote.origin.url=<src>` asserting no `set-url` is produced, and one where it differs asserting it is.



The fix is correct in intent, but the unconditional implementation regresses idempotency and breaks on repos lacking an `origin` remote. A `GitConfig`-gated conditional is a small change that resolves both.

Fizzadar:
#1843 has now fixed this!

PR #1766, opened 2026-05-20
feat: Allow `fetch_tags="force"` in `git.repo`

inducer:

This is a follow-on #1745. Some repositories (https://github.com/discourse/discourse notably) rely on moving tags to indicate the current "released" version. #1745 allows issuing `git fetch --tags`, but that refuses to update tags if they've been moved upstream. This adds the possibility of specifying `fetch_tags="force"` to allow updating to succeed anyway.


DonDebonair:
I think this is misleading, as the `--force` option of `git fetch` does not only refer to tags. Better to add a separate boolean `force` parameter.

inducer:
> Better to add a separate boolean `force` parameter.

I disagree, actually. :slightly_smiling_face:  For two reasons:

- `git.repo` can issue many separate `git` commands. A function-global `force` parameter could/should likely be interpreted as passing `--force` for every one of those, at least where applicable. At least for my use case, that's not the desired intent.
- Considering the [documentation](https://git-scm.com/docs/git-fetch), there are two implications described for `--force` on `git fetch`, one is the one intended here, and the other is applicable only if a refspec is explicitly given (which isn't the case here).

DonDebonair:
I suggest we stay true to the interface of the `git` command. Both `--tags` and `--force` are switches, so they should be represented as booleans (i.e. `force=False` and `tags=False`). That causes the least surprises.

Fizzadar:
Agreed with @DonDebonair. @inducer is there a situation where that would not work? I’m reading the git docs and the additional force behavior would not apply here, I don’t see a use case for a specific “force tags but nothing else” argument…

inducer:
So just to be clear: you'd like a `force: bool` argument that adds `--force` to every git command that's being run by that operation if given as true?

Fizzadar:
> So just to be clear: you'd like a `force: bool` argument that adds `--force` to every git command that's being run by that operation if given as true?

Yes, for the relevant git commands that take it (so just `pull` + `fetch`).

inducer:
How's this?

src/pyinfra/operations/git.py
```diff
@@ -118,7 +119,9 @@ def repo(
     + update_submodules: update any git submodules
     + recursive_submodules: update git submodules recursively
     + depth: create a shallow clone with a history truncated to the specified number of commits
-    + fetch_tags: Whether all tags should be fetched prior to attempting to check out the specified revision
+    + fetch_tags: Whether all tags should be fetched prior to attempting to
+      check out the specified revision.
+    + force: Supply a '--force' flag to every git subcommand that supports it.
```
Fizzadar, line 124:
```suggestion
    + force: Execute `fetch`, `pull` and `checkout` commands with `--force`
```

src/pyinfra/operations/git.py
```diff
@@ -210,15 +203,17 @@ def repo(
...
-                git_commands.append("pull --rebase")
+                git_commands.append(StringCommand("pull", "--rebase", *force_flag_list))
             else:
-                git_commands.append("pull")
+                git_commands.append(StringCommand("pull", *force_flag_list))
 
     if update_submodules:
         if recursive_submodules:
-            git_commands.append("submodule update --init --recursive")
+            git_commands.append(
```
Fizzadar, line 212: I don't think submodules should be included by default here, I made a proposed arg doc string change to match. Submodules would probably need a `force_submodules` flag equivalent.

Fizzadar: (also, this is my bad for not accounting for those when I mentioned pull+fetch earlier).

Fizzadar (commented):
Hi @inducer sorry for the delay on this one, nearly there 🙏

inducer:
How's this?

Fizzadar (approved):
Thank you @inducer, this looks good now!

PR #1767, opened 2026-05-20
docs: restore loop cycle error guidance

### LINKED ISSUE #1494: Cycle error in 3.5.1 and missing docs
## Describe the bug

I do get a loops-cycle-error in my deploy code. I think I understand the cause of it because I use some loops. The problem is, that all documentation related to that error which should be here as per the pyinfra error message:

https://docs.pyinfra.com/en/latest/deploy-process.html#loops-cycle-errors

was removed in the 3.x branch with 236d7fe83c73a47ba11aacdc811c37af94889b55. And I am wondering if that was a mistake, because it seems the ordering in the DAG is still relevant. If that is the case, it would be prudent to restore the original documentation for that because any nontrivial setup will use loops sooner or later.

On a sidenote: The search functionality for the docs is woefully inadequate. For example searching for "loop" does not yield any relevant results, despite numerous appearances of the word "loop" in the api reference for the host module.

puneetdixit200:
## Summary
- restore the missing `Loops & Cycle Errors` section in `docs/deploy-process.rst`
- add a stable `#loops-cycle-errors` anchor for the link referenced by the cycle error message
- document using `host.loop` to disambiguate operation ordering inside loops

Fixes #1494.

## Validation
- `git diff --check -- docs/deploy-process.rst`
- `DOCS_VERSION=latest uv run --python 3.14 --group docs sphinx-build -b html docs build/docs`
- `rg -n "loops-cycle-errors|Loops & Cycle Errors" build/docs/deploy-process.html docs/deploy-process.rst`
- `uv run --python 3.14 pytest tests/test_cli/test_cli_deploy.py::TestCliDeployState::test_deploy -q`

## AI assistance disclosure
This contribution was prepared with assistance from OpenAI Codex. I reviewed the AI policy, issue context, existing `host.loop` implementation, generated documentation output, and validation results before submitting.

### DIFF
--- a/docs/deploy-process.rst
+++ b/docs/deploy-process.rst
@@ -183,3 +183,84 @@ Since this gets executed before nginx is installed by ``apt.packages`` operation
         reloaded=True,
         _if=remove_default_site.did_change,
     )
+
+.. _loops-cycle-errors:
+
+Loops & Cycle Errors
+~~~~~~~~~~~~~~~~~~~~
+
+In CLI mode ``pyinfra`` builds a single DAG to determine the order in which
+operations are executed. This usually produces the order a deploy author
+expects, but loops can make the same operation line appear multiple times with
+different per-host paths through the deploy. When those paths disagree about
+which operation must run first, the ordering graph contains a cycle and
+``pyinfra`` raises a cycle error.
+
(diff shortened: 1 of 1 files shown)


Fizzadar (approved):
LGTM, thanks for putting this back @puneetdixit200!

PR #1770, opened 2026-05-20
docs: clarify fact-failure semantics for requires_command and check_preconditions

### LINKED ISSUE #1768: docs: using-operations.rst is stale on fact-failure semantics
## Problem

`docs/using-operations.rst` line 170 states:

> When facts fail due to an error the host will be marked as failed just as it would when an operation fails.

This was written before the `requires_command` rework in #1647 (commit a2252c39, `feat(facts): add requires_command guard sentinel and check_preconditions() hook`).

Current behavior in 3.x:

- Fact command exits non-zero: host is failed (old behavior, still correct).
- `requires_command` binary missing on host: `MissingCommandError` is caught, fact returns `default()`, host is NOT failed.
  - Prepare phase: silent.
  - Execute phase: warning logged, default returned (v4 will raise).
- `check_preconditions()` not satisfied: same phase-aware behavior.

This is documented in `docs/api/facts.md` but the `using-operations.rst` "Fact Errors" section was not updated.

## Reproducer

```python
from pyinfra import host

class TestFact(FactBase):
    def requires_command(self) -> str:
        return "nonexistingcmd"
    command = "thiswillnotwork"
    def process(self, output) -> bool:
        return False

myfact = host.get_fact(TestFact)  # returns None, host not failed
server.shell(name="Say hello", commands="echo 'hello'", _if=myfact)
```

The operation runs and succeeds, which contradicts the doc.

## Suggested fix

Update the "Fact Errors" section to distinguish:

- fact command non-zero exit (host failed unless `_ignore_errors`)
- `requires_command` miss (silent default, phase-aware)
- `check_preconditions()` miss (same)

Link out to `docs/api/facts.md` for the full design.

wowi42:
## Summary

The "Fact Errors" section in `docs/using-operations.rst` stated that any fact failure marks the host as failed. That is only true when the fact command exits with a non-zero status.

The `requires_command` guard (added in #1647) and the `check_preconditions()` hook take a phase-aware silent path instead:

- prepare phase: skip is silent, fact returns `default()`
- execute phase (v3): warning logged, fact returns `default()`
- execute phase (v4): will raise

Update the section to cover all three paths and link out to the full design in `docs/api/facts.md`.

Closes #1768

### DIFF
--- a/docs/using-operations.rst
+++ b/docs/using-operations.rst
@@ -167,12 +167,19 @@ See :doc:`facts` for a full list of available facts and arguments.
 Fact Errors
 ^^^^^^^^^^^
 
-When facts fail due to an error the host will be marked as failed just as it would when an operation fails. This can be avoided by passing the ``_ignore_errors`` argument:
+When a fact command exits with a non-zero status the host is marked as failed, just as when an operation fails. This can be avoided by passing the ``_ignore_errors`` argument:
 
 .. code:: python
 
     if host.get_fact(LinuxName, _ignore_errors=True):
         ...
+
+Two other "fact skipped" paths do **not** fail the host. They are phase-aware:
+
+- ``requires_command`` declares a binary that must be present on the host. If the binary is absent, the fact returns its ``default()`` value. In the prepare phase the skip is silent; in the execute phase a warning is logged. In v4 the execute phase will raise instead.
+- ``check_preconditions()`` lets a fact declare a runtime prerequisite (for example a kernel module loaded). When the precondition is not met the same phase-aware behavior applies.
+
+See :doc:`api/facts` for the full design and exception hierarchy.
 
 The ``inventory`` Object
 ------------------------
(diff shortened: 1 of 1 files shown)


docs/using-operations.rst
```diff
@@ -167,13 +167,20 @@ See :doc:`facts` for a full list of available facts and arguments.
...
     if host.get_fact(LinuxName, _ignore_errors=True):
         ...
 
+Two other "fact skipped" paths do **not** fail the host. They are phase-aware:
+
+- ``requires_command`` declares a binary that must be present on the host. If the binary is absent, the fact returns its ``default()`` value. In the prepare phase the skip is silent; in the execute phase a warning is logged. In v4 the execute phase will raise instead.
+- ``check_preconditions()`` lets a fact declare a runtime prerequisite (for example a kernel module loaded). When the precondition is not met the same phase-aware behavior applies.
+
+See :doc:`api/facts` for the full design and exception hierarchy.
+
```
Fizzadar:
```suggestion
.. Important::
    Facts may choose to silently ignore errors for missing commands (eg mysql not installed) and instead return a default value. In v4 this will raise an error during the execution phase.
```

I think this is more user-friendly? I don't think we need mention precondition as it's a property of the fact itself, not relevant to the user.

Fizzadar (changes requested):
One comment/suggestion

Fizzadar (approved):

PR #1774, opened 2026-05-21
fix(operations.files): handle large download cache time

### LINKED ISSUE #1394: Large `cache_time` value to `files.download` causes OverflowError exception
We used `cache_time=9999999999999` to effectively disable redownloading of the file which is supposed to never change. The file downloaded on a first time, but on a second time pyinfra crashed with the exception:
```
-​-> An internal exception occurred:

  File "/home/*********/relay/venv/lib/python3.11/site-packages/pyinfra/operations/files.py", line 129, in download
    ctime = host.get_fact(Date).replace(tzinfo=None) - timedelta(seconds=cache_time)
    ^^^^^^^^^^^^^^^^^
OverflowError: date value out of range

-​-> The full traceback has been written to pyinfra-debug.log
```

One way to fix it would be to handle the underflow and assume that the cache has not expired if there is an underflow.

Another option that would have helped us is to not redownload the file if checksum is specified and the file with the correct checksum is already there. This would be preferable for our usecase, then we would not need to specify `cache_time` at all.

See https://github.com/chatmail/relay/issues/602 for downstream issue.

puneetdixit200:
## Summary
- prevent `files.download` from raising `OverflowError` when `cache_time` is too large to subtract from the current date
- treat overflowed positive cache windows as not expired
- add an operation fixture for the reported underflow case

Fixes #1394

## AI assistance
OpenAI GPT-5 helped draft this patch and run verification. This disclosure is included per `AI_POLICY.md`.

## Verification
Red test before the fix:
- `uv run pytest tests/test_operations.py -k 'files_download_download_cache_time_underflow' -q` failed with `OverflowError: date value out of range`

After the fix:
- `uv run pytest tests/test_operations.py -k 'files_download_download_cache_time_underflow' -q`
- `uv run pytest tests/test_operations.py -k 'files_download' -q`
- `uv run pytest tests/test_operations.py -q`
- `uv run ruff format --check src/pyinfra/operations/files.py`
- `uv run ruff check src/pyinfra/operations/files.py`
- `uv run mypy src/pyinfra/operations/files.py`
- `git diff --check`

Note: full `uv run ruff check` currently reports existing repository-wide findings outside this change, including `src/pyinfra/connectors/scp/client.py`, `src/pyinfra/facts/*.py`, `src/pyinfra/operations/docker.py`, `src/pyinfra/progress.py`, and `src/pyinfra_cli/*.py`.

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -151,7 +151,12 @@ def download(
         if cache_time:
             # Time on files is not tz-aware, and will be the same tz as the server's time,
             # so we can safely remove the tzinfo from the Date fact before comparison.
-            ctime = host.get_fact(Date).replace(tzinfo=None) - timedelta(seconds=cache_time)
+            try:
+                ctime = host.get_fact(Date).replace(tzinfo=None) - timedelta(
+                    seconds=cache_time,
+                )
+            except OverflowError:
+                ctime = datetime.min if cache_time > 0 else datetime.max
             if info["mtime"] and info["mtime"] < ctime:
                 download = True
 
--- a/tests/operations/files.download/download_cache_time_underflow.yaml
+++ b/tests/operations/files.download/download_cache_time_underflow.yaml
@@ -1,0 +1,14 @@
+args:
+  - http://myfile
+kwargs:
+  dest: /home/myfile
+  cache_time: 9999999999999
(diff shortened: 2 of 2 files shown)


Fizzadar (approved):
LGTM, thanks @puneetdixit200!

PR #1776, opened 2026-05-25
feat(operations.go,facts.go): implement go operations and facts

yagarea:

---

Due to the project AI policy I disclose that tests were created using LLM assistance and I will have to do more manual checks on them. Therefore I am creating PR with *WIP* tag so code itself can reviewed in a meantime. I have manually checked the code and already used it in my projects but it still needs to get checked if it uses best practices and conventions by pyinfra maintainers.


dvdkon:
Looks good to me and works on our server.

wowi42:
Reviewed the code. Clean implementation overall. The Semantic Import Versioning handling, `(devel)` filtering, and consistent `QuoteString` usage show real care.

**Solid:**
- Shell safety is correct throughout: all user values into `go install` are `QuoteString`-wrapped, and the removal path quotes the derived binary name.
- Fact parsing is robust. It resets `current_path` on non-tab lines, skips `dep`/`build`, and drops `(devel)` / `command-line-arguments` non-installs.
- The set-vs-list in the fact fixture is fine: `json_encode` serializes sets via `sorted(list(...))`, so `{"v0.9.0"}` matches `["v0.9.0"]`.
- Batching is safe. `to_install` only holds `@version`/`@latest` entries, so it never mixes bare and versioned args (the actual Go module-mode restriction).

**Worth addressing:**
- `latest=True` is never idempotent: an installed unversioned package always re-emits `go install ...@latest` and reports "changed" every run. Unavoidable offline (matches pip), but please add a one-line docstring note.
- Binary-name collision on removal: two packages whose paths share a last element plus major version resolve to the same binary name, so `rm -f` could remove the wrong one. Edge case, a short comment would help.
- `go env` re-runs once per removed package since it's inside each `yield`. Minor.

**Test gaps** (you already flagged WIP):
- No fixture for `latest=True` when the package is not yet installed (should hit `to_install`, not `to_upgrade`).
- None for the explicit `@latest` suffix dedup, nor for passing a bare `str` instead of a list.

No correctness blockers in the code. Main asks before merge: fill those test gaps and note the `latest=True` re-install behavior.

yagarea:
@wowi42 Thank you for your feedback. I hope I have addressed all your comments.

Fizzadar (approved):
LGTM! Thank you @yagarea

yagarea:
Thank you for review and merging!

PR #1777, opened 2026-05-25
fix(facts.openrc): make OpenrcStatus match service names containing a dot

chimekkoo:
Makes `pyinfra.facts.openrc.OpenrcStatus` match services containing a dot, e.g. to make `pyinfra.operations.openrc.service(...)` report a correct service state in such case.
Also, tests checking OpenrcStatus and OpenrcEnabled behaviour for services containing a dot were added.


Fizzadar (approved):
Nice fix, thanks @chimekkoo!

PR #1778, opened 2026-05-26
feat(cli): allow http(s) URLs as deploy file arguments

wowi42:
## Summary

Adds support for `http(s)://` URLs as deploy file arguments:

```sh
pyinfra 10.0.0.1 https://raw.githubusercontent.com/example/repo/main/deploy.py
```

The URL is fetched via stdlib `urllib` to a temp `.py` and handed to the existing exec path. A one-line warning prints on every fetch.

## Why this is probably a bad idea

It runs arbitrary Python from a URL with no signature, no checksum, no pinning, no caching, no auth. The only safety net is "warn on fetch and trust the user." A typo'd URL, a compromised raw host, a redirected CDN, or a hijacked gist all execute as whatever pyinfra connects as, which is often root. For real deploys, keep using files in version control.

## Why I am opening it anyway

I wanted `uvx pyinfra 10.0.0.1 https://.../op.py` to work in one line, and it made me laugh that it actually does. Treat this as a discussion starter, not a serious feature pitch. If the shape is salvageable (opt-in flag, mandatory `@sha256:`, GitHub token auth, response caching), happy to iterate. If not, close at will, no hurt feelings.

## What is in v1

- `is_remote_url` / `fetch_remote_deploy_file` in `pyinfra_cli/util.py`, stdlib only.
- Dispatch in `_validate_operations` accepts mixed local + URL deploy file lists.
- Temp files cleaned up at process exit.
- 4 new unit tests; full suite still 1770 passing.
- Short docs note covering the v1 limits.

## Test plan

- [ ] `scripts/dev-lint.sh`
- [ ] `scripts/dev-test.sh`
- [ ] Manual smoke: `pyinfra @local https://gist.githubusercontent.com/.../raw/hello.py` against a host you control


Fizzadar (commented):
This is certainly an interesting one, see prior inventories suggestion here: https://github.com/pyinfra-dev/pyinfra/issues/447 (though functions solves that). I suppose one could write a function for ops that downloads + evals op code.

IMO this is cool but a security footgun, and something we should avoid having in pyinfra for that reason (nothing stops anyone using a shell script to do it).

What about accepting operation code on stdin like this for a compromise?

```sh
curl https://some/path/deploy.py | pyinfra INVENTORY -
```

Also evil 🙃

PR #1780, opened 2026-05-27
feat(connectors): add chain connector

maisim:
## What

A new "transparent" `@chain` connector that lets you nest pyinfra connectors with a natural `/@` delimiter, so you can target hosts that sit behind other hosts:

```shell
# Docker container on a remote host via SSH
pyinfra @ssh/mydoodba/@docker/myodoodev16-odoo-1 files.put src/ /opt/odoo/

# Chroot on a remote host via SSH
pyinfra @ssh/build-server/@chroot/rootfs server.shell "ls /"

# Three levels deep
pyinfra @ssh/jumpbox/@docker/mycontainer server.shell "whoami"
```

The `/@` delimiter reads like a path — it's obvious at a glance that you're descending through layers. It also avoids conflicting with existing connector syntax (`@connector/arg`).

## How it works

- `/@` in a hostname is detected at inventory time and routed to `ChainedConnector`
- Only the **outermost** connector establishes a real connection (SSH socket, local shell, ...)
- **Inner** connectors just *wrap* shell commands — `docker exec`, `chroot`, `cp` — which are composed into a single shell string and sent to the outer connector
- File uploads are pipelined through temp paths at each layer, cleaned up after

## What's chain-compatible today

| Connector | Outer | Inner | Why not inner? |
|-----------|-------|-------|-----------------|
| `ssh` | yes | no | needs paramiko sockets, not a plain shell string |
| `local` | yes | no | not wired for it |
| `docker` | yes | **yes** | `docker exec` / `docker cp` |
| `chroot` | yes | **yes** | `chroot` / `cp` |
| `dockerssh` | yes | no | SSH part is bypassed as inner, use `docker` instead |

Making a connector chain-compatible is a few lines of boilerplate — three `wrap_*` methods + a `runtime_id_field`. The PR includes docs on how to do it.

## Backward compatibility

Zero impact on existing code. No existing connector is modified in a breaking way. The chain syntax only activates when `/@` is present in a hostname. Everything else — `@ssh/host`, `@docker/container`, `@chroot/path` — works exactly as before.

## Files changed

- `src/pyinfra/connectors/chain.py` — the `ChainedConnector` itself (415 lines)
- `src/pyinfra/connectors/base.py` — `get_runtime_id()` + `wrap_*` interface
- `src/pyinfra/connectors/docker.py` — chain support for Docker
- `src/pyinfra/connectors/chroot.py` — chain support for Chroot
- `src/pyinfra/api/inventory.py` — `/@` detection
- `docs/connectors/chain.md` — user-facing docs
- `docs/api/connectors.md` — developer docs for making connectors chain-compatible
- `tests/test_connectors/test_chain.py` — 20 tests covering parsing, inventory detection, wrapping, runtime IDs, and error paths

Built using gemma4, Opus 4.6, DeepSeek v4

- [ ] Pull request title follows the 

### DIFF
--- a/docs/api/connectors.md
+++ b/docs/api/connectors.md
@@ -246,6 +246,53 @@ Without proper command wrapping, shell operators and complex commands will fail.
 
 For complete examples see pyinfra's built-in connectors in `pyinfra/connectors/docker.py`, `pyinfra/connectors/chroot.py`, `pyinfra/connectors/ssh.py` and `pyinfra/connectors/local.py`, as well as the command wrapping utilities in `pyinfra/connectors/util.py`.
 
+## Making a connector chain-compatible
+
+To allow a connector to be used as an **inner** layer in a :doc:`chain </connectors/chain>`,
+it must implement three wrapping methods and declare how to resolve its runtime identifier.
+
+### Runtime identifier
+
+Set the `runtime_id_field` class attribute to the data key that holds the connector's
+runtime identifier (container name, chroot directory, etc.). The default
+`get_runtime_id()` implementation reads `self.data[self.runtime_id_field]`:
+
+```py
+class MyConnector(BaseConnector):
+    runtime_id_field = "my_container_id"
+```
+
+For more complex resolution (e.g. combining multiple data keys), override
+`get_runtime_id()` directly:
+
+```py
+class MyConnector(BaseConnector):
+    def get_runtime_id(self) -> str:
+        return f"{self.data['remote']}:{self.data['name']}"
+```
+
+`get_runtime_id()` is called *before* `connect()`, so it must work solely from data
+available at construction time.
+
+### Command wrapping
+
+```py
+def wrap_exec_command(self, command: StringCommand, container_id: str) -> StringCommand:
+    """Return a command that runs ``command`` inside this connector's target."""
+    return StringCommand("my-tool", "exec", container_id, "--", "sh", "-c", QuoteString(command))
+
+def wrap_copy_into(self, src_on_parent: str, dest: str, container_id: str) -> StringCommand:
+    """Return a command that copies from parent into this connector's target."""
+    return StringCommand("my-tool", "cp", src_on_parent, f"{container_id}/{dest}")
+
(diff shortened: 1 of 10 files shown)


wowi42 (changes requested):
The design (outer connector owns the connection, inner layers become command wrappers) is sound, and the shell-quoting hygiene is good: all `wrap_*` implementations route user-controlled values through `QuoteString`. A few things need fixing before this can land:

Blocking:

1. `put_file` silently drops all connector control arguments: `outer.put_file(local_tmp, outer_tmp)` is called without `**kwargs`, and the per-layer copy/rm commands run via `outer.run_shell_command` with no sudo/env handling. `files.put(..., _sudo=True)` through a chain (e.g. into a root-owned chroot on an SSH host) silently ignores sudo and fails with permission denied.
2. `pr-lint` fails: the `[RFC]` prefix breaks the Conventional Commits title check. Drop the prefix or keep the PR in draft while it is an RFC.

Minor:

- `get_file` never cleans up the intermediate temp files inside inner layers (`chain-out-{i}-...`); only the outermost temp is removed. `put_file` cleans every layer, `get_file` should too.
- `put_file` buffers the whole file in memory and always round-trips through a local `mkstemp`, even when `filename_or_io` is already a local path that could go straight to `outer.put_file`. Large uploads will spike RAM.
- `run_shell_command` and `put_file` in `chain.py` lack type annotations (required on new code), and there is a mid-function `import os`.
- `_sudo_password` is broken by design for inner layers: the SUDO_ASKPASS script path is generated on the outer host but the wrapped command runs inside the container/chroot where that path does not exist. Worth documenting as a limitation.
- Chain detection in `inventory.py` only fires when the outer connector has an arg, so `@local/@docker/foo` silently misroutes to `LocalConnector` with a weird name instead of chaining or raising.
- `make_names_data` breaks after the first yield per connector, so inventory-style connectors that yield multiple hosts (terraform, vagrant) would be silently truncated to one host.

Fizzadar (commented):
Big fan of this @maisim! Also means we can drop dockerssh+podmanssh in v4.

PR #1785, opened 2026-06-02
feat(operations.git): support ssh://, raise explicit error on bad proto/keyscan combination

maisim:
git.repo's ssh_keyscan regex only matched the scp-like `user@host` form, so `ssh://`
sources - with or without a port - raised `OperationError` instead of keyscanning their
host.

- `ssh://[user@]host[:port]/path` and `user@host:path` are parsed for the host (and the
  port for `ssh://`), bracketed IPv6 included.
- Sources on other transports (`https://`, `git://`, `file://`) now raise an explicit
  `OperationError`: ssh-keyscan populates `known_hosts` so an SSH clone can run, and a
  clone over any other transport never reads it.
- The `lastindex` condition and the nominal IPv6 handling are gone with the rewrite.

**AI disclosure**: prepared with AI assistance — deepseek-4.1-flash 


wowi42:
Sorry for the late review.

Needs discussion before merge.

The PR extends `git.repo` URL parsing so `ssh_keyscan=True` extracts host and port from `ssh://` and `https://` URLs. The change is focused and CI passes, but the interaction between `ssh_keyscan` and HTTPS URLs is unclear.

Issue: when `src` is `https://host/path`, the PR runs `ssh-keyscan` against that host. ssh-keyscan is for SSH host keys, not HTTPS. For HTTPS clones the keyscan is irrelevant, and defaulting to port 22 instead of 443 is confusing. Either keyscan should be limited to SSH-family URLs (`git@host`, `ssh://`), or the PR body should explain why HTTPS triggers an SSH scan.

Also, the PR branch is based on `410d5f92`, which is well behind current `3.x`. It will need a rebase before merge.

Minor notes:
- `src/pyinfra/operations/git.py:149`: `m.lastindex and m.lastindex >= 2` can be simplified to `if m.group(2):`.
- The `\[? ... \]?` in the SSH regex does not actually parse IPv6 addresses.

maisim:
Hi @wowi42,

Thanks for the review. What failed was the combination https + ssh_keyscan=True, because the regex found no host there.

ssh_keyscan now only applies to user@host:path and ssh://host[:port]/path sources, which is what the flag is for — populate known_hosts so an SSH clone can run. Any other transport now raises an explicit OperationError.

Fizzadar (approved):
This looks great, thank you @maisim! Funny.. it actually uncovers a bug in `ssh.keyscan` where the port isn't properly handled, made https://github.com/pyinfra-dev/pyinfra/pull/1969 to address it

PR #1786, opened 2026-06-04
feat(operations.util.files): normalize octal mode spellings in ensure_mode_int

### LINKED ISSUE #1727: Add real octal value for mode to operatons.files and facts.files
Both the 'mode' argument to operations.files.* and the 'mode' output of facts.files.* contain the file mode as a decimal value (converted from a string containing the octal value using base 10). This makes it complex to use bit-wise comparisons and operations with common contants, such as thos contained in the stdlib 'stat' module.

For example, to make sure some specific binary has no setuid bit:

```
file_fact = host.get_fact(File, filename)
# convert using base 8
octal_mode = int(str(file_fact['mode']), 8)
if octal_mode & stat.S_ISUID:
    # remove setuid bit and use resultant value as octal string without '0o' prefix
    files.file(filename, mode=oct(octal_mode & ~stat.S_ISUID)[2:])
```

Modifying the current 'mode' argument and output probably breaks a lot of stuff, so instead I would fix this by adding a 'mode_octal' argument/fact key of type int that contains/accepts the mode as a real octal value, with operations.files operations adapted to disallow using 'mode' and 'mode_octal' at the same time.

An alternative less intrusive fix would be to add a function to operations.util.files that does the conversion when necessary, and adding documentation to the operations.files for that.

I would be happy to create a PR for either of those options.

mvanhorn:
## Summary
`files` operations now accept every common mode spelling: `644`, `"644"`, `"0644"`, and `"0o644"` all normalize to the canonical octal-digit integer that `chmod` and the file facts already use. Clearly invalid modes (containing digits 8 or 9) raise a `ValueError` instead of silently producing a wrong `chmod`.

## Why this matters
#1727 reports that working with `mode` values bitwise (e.g. masking `stat.S_ISUID`) is awkward because the value is the octal digit string as a plain int. Fizzadar suggested normalizing the accepted spellings under the existing `mode` argument so any input variant works. This implements that normalization in `ensure_mode_int` (`src/pyinfra/operations/util/files.py`), which all four `files` operation call sites already go through.

One caveat documented in the docstring: a true octal int literal (`0o644`) evaluates to `420` before pyinfra sees it, so it cannot be distinguished from a literal `420`. The `"0o644"` string form covers that intent explicitly.

## Testing
- `uv run pytest --disable-warnings -m 'not end_to_end'`: 1819 passed
- New `TestEnsureModeInt` cases in `tests/test_operations_utils.py` cover all accepted spellings, symbolic pass-through (`"u+x"`), `None`, setuid modes (`"4755"`), and the non-octal error paths
- `ruff check`, `ruff format --check`, and `mypy` clean on the touched files

Fixes #1727

### DIFF
--- a/src/pyinfra/operations/util/files.py
+++ b/src/pyinfra/operations/util/files.py
@@ -22,19 +22,39 @@ def unix_path_join(*parts) -> str:
 
 
 def ensure_mode_int(mode: str | int | None) -> int | str | None:
-    # Already an int (/None)?
-    if isinstance(mode, int) or mode is None:
+    """
+    Normalise the accepted mode spellings to the canonical octal-digit
+    integer pyinfra passes to ``chmod`` (and that facts emit), so ``644``,
+    ``"644"``, ``"0644"`` and ``"0o644"`` all mean ``rw-r--r--``.
+
+    Integers are read as their octal digit representation (``644`` means
+    ``rw-r--r--``), matching the existing operations/facts contract. A true
+    octal literal (``0o644``) is indistinguishable from its decimal value
+    (``420``) at runtime - use the ``"0o644"`` string form instead.
+
+    Symbolic modes (``"u+x"``) and ``None`` pass through unchanged.
+    """
+    if mode is None:
         return mode
 
-    try:
-        # Try making an int ('700' -> 700)
-        return int(mode)
+    if isinstance(mode, int):
+        digits = str(mode)
+    else:
+        digits = mode.strip().lower()
+        if digits.startswith("0o"):
+            digits = digits[2:]
+        if not digits.isdigit():
+            # Return as-is (ie +x which we don't need to normalise, it always gets run)
+            return mode
 
-    except (TypeError, ValueError):
-        pass
+    if any(c not in "01234567" for c in digits):
+        raise ValueError(
+            f"Invalid file mode: {mode!r} contains non-octal digits "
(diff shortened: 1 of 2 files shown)


src/pyinfra/operations/util/files.py
```diff
@@ -22,19 +22,39 @@ def unix_path_join(*parts) -> str:
...
+    Integers are read as their octal digit representation (``644`` means
+    ``rw-r--r--``), matching the existing operations/facts contract. A true
+    octal literal (``0o644``) is indistinguishable from its decimal value
+    (``420``) at runtime - use the ``"0o644"`` string form instead.
 
-    except (TypeError, ValueError):
-        pass
+    Symbolic modes (``"u+x"``) and ``None`` pass through unchanged.
+    """
+    if mode is None:
```
Fizzadar:
```suggestion
    if mode is None or isinstance(mode, bool):
```

Let's pass through bools here as-is because operations rely on that (ie `True` being copy from local system, `False` generally being don't apply perms, although that's not actually formalized anywhere). Otherwise we risk breaking peoples deploy code.

Fizzadar (approved):
This is great, thank you @mvanhorn! One small tweak I'll add now so we can land this.

mvanhorn:
Glad the octal mode normalization landed - thanks @Fizzadar!

mvanhorn:
Appreciate the merge, @Fizzadar. Normalizing the octal mode spellings should end the false-change diffs on ensure mode.

Issue #1746, opened 2026-05-13
assume_exists possibility for various files operations


szimszon:
## Is your feature request related to a problem? Please describe
I wanted to files.download a zip archive 
  and then files.unarchive it 
  and files.move one of the inner directories to another location and 
it failed because of a pre-check failed as the actual file is not yet downloaded...

## Describe the solution you'd like
I would like to skip check for the file presence with assume_exists set to True...

eliasp:
Same goes for `files.move`, which will fail in case `src` or `dest` are to be created by another operation:
- `src /var/tmp/pyinfra/my.cfg does not exist` (`src` is created by an earlier operation)
- `dest /var/tmp/pyinfra/backup is not an existing directory` (`dest` is created by an earlier operation)

See the deploy below to reproduce this:
```python
from pyinfra.operations import files

filename = "my.cfg"
tempdir_path = "/var/tmp/pyinfra"
backupdir_path = f"{tempdir_path}/backup"
origfile_path = f"{tempdir_path}/{filename}"
backupfile_path = f"{backupdir_path}/{filename}"

op_tempdir = files.directory(
    name="Create the tempdir",
    path=tempdir_path,
)

op_origfile = files.file(
    name="Some file to be backed up",
    path=origfile_path,
    present=True,
    _if=op_tempdir.did_succeed,
)

op_backupdir = files.directory(
    name="Create the backup dir",
    path=backupdir_path,
    _if=op_tempdir.did_succeed,
)

op_backup = files.move(
    name="Move file away",
    src=origfile_path,
    dest=backupdir_path,
    _if=lambda: (
        op_origfile.did_succeed
        and op_backupdir.did_succeed,
    )
)
```

eliasp:
There's also a related discussion: https://github.com/pyinfra-dev/pyinfra/discussions/1629

Fizzadar:
Alternative: should we just make these operations _not_ blow up (at prepare phase) if the file doesn't exist? We can still raise during execution. IMO would be the expected behavior in all cases?

vaz-ar:
> Alternative: should we just make these operations _not_ blow up (at prepare phase) if the file doesn't exist? We can still raise during execution. IMO would be the expected behavior in all cases?

Isn't this a breaking change ? it will alter the expected behaviour for the people who were expecting it to fail at the prepare phase.

pascal-cm:
[...]

PR #1787, opened 2026-06-04
feat(operations.files): Ensure the operations do not fail at the prepare stage if the source does not exists

vaz-ar:
Updated the move, copy and unarchive operations to only log the fact
    that the source file does not exists, to prevent the operations from
    failing at the prepare stage.

- [X] Pull request title follows the 

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -18,7 +18,6 @@ from typing import IO, Any
 from jinja2 import TemplateRuntimeError, TemplateSyntaxError, UndefinedError
 
 from pyinfra import host, logger, state
-from pyinfra.api.output import format_text
 from pyinfra.api import (
     FileDownloadCommand,
     FileUploadCommand,
@@ -31,6 +30,7 @@ from pyinfra.api import (
     operation,
 )
 from pyinfra.api.command import make_formatted_string_command
+from pyinfra.api.output import format_text
 from pyinfra.api.util import (
     get_call_location,
     get_file_io,
@@ -1488,7 +1488,7 @@ def move(src: str, dest: str, overwrite=False):
     """
 
     if host.get_fact(File, src) is None:
-        raise OperationError(f"src {src} does not exist")
+        logger.warning(f"src {src} does not exist")
 
     if not host.get_fact(Directory, dest):
         raise OperationError(f"dest {dest} is not an existing directory")
@@ -1514,7 +1514,7 @@ def copy(src: str, dest: str, overwrite=False):
     """
     src_is_dir = host.get_fact(Directory, src)
     if not host.get_fact(File, src) and not src_is_dir:
-        raise OperationError(f"src {src} does not exist")
+        logger.warning(f"src {src} does not exist")
 
     if not host.get_fact(Directory, dest):
         raise OperationError(f"dest {dest} is not an existing directory")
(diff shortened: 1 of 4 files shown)


vaz-ar:
This is related to #1746 and #1629

vaz-ar:
I have added tests but I am not 100% sure of them, let me know if things can be improved

Fizzadar (commented):
Hi @vaz-ar! Thank you for the PR - I think for solving this issue we'd be better implementing the "don't blow up during fact phase for missing files" change I referenced in the underlying issue: https://github.com/pyinfra-dev/pyinfra/issues/1746#issuecomment-4635433684.

vaz-ar:
@Fizzadar OK but why are you closing the PR ? I could have reworked the code.

Fizzadar:
@vaz-ar habit! re-opened :)

src/pyinfra/operations/files.py
```diff
@@ -1484,12 +1488,12 @@ def move(src: str, dest: str, overwrite=False):
     """
 
     if host.get_fact(File, src) is None:
-        raise OperationError(f"src {src} does not exist")
+        logger.warning(f"src {src} does not exist")
```
aktiur, line 1491: In this case and others:
- the operation should raise like before when `state.current_stage == StateStage.Execute`
- this should use `host.log` to identify on which host the log was emitted, and `info` seems to me to be more appropriate than `warning` for this usecase.

Something like the following?

```python
if host.get_fact(File, src) is None:
    if state.current_stage == StateStage.Execute:
        raise OperationError(f"src {src} does not exist")
    else:
        host.log(f"src {src} does not exist in prepare phase", logger.info)
```

src/pyinfra/operations/files.py
```diff
@@ -1484,12 +1488,12 @@ def move(src: str, dest: str, overwrite=False):
...
 
     if not host.get_fact(Directory, dest):
         raise OperationError(f"dest {dest} is not an existing directory")
```
aktiur, line 1494: I think this case should be handled exactly like the previous one: the `dest` directory could also be created in a previous operation.

src/pyinfra/operations/files.py
```diff
@@ -1510,12 +1514,12 @@ def copy(src: str, dest: str, overwrite=False):
     """
     src_is_dir = host.get_fact(Directory, src)
     if not host.get_fact(File, src) and not src_is_dir:
-        raise OperationError(f"src {src} does not exist")
+        logger.warning(f"src {src} does not exist")
 
     if not host.get_fact(Directory, dest):
         raise OperationError(f"dest {dest} is not an existing directory")
```
aktiur, line 1520: Same case as R1494, the directory could be created by a previous operation.

src/pyinfra/operations/files.py
aktiur, line 2246: Another directory case.

aktiur (changes requested):
Thanks for the PR! I came across the same "download then unarchive" issue as you.

Instead of a systematic warning, I think it'd be better if:
- the relevant operations were emitting an info log in prepare phase
- they were raising just like before in the execute phase.

Moreover, there are similar checks in `copy`, `move` and `unarchive` that verify that the destination is an existing directory. I think the same logic should apply there.

vaz-ar:
@aktiur feel free to open a PR on this, I won't have time to work on this anymore.

PR #1788, opened 2026-06-05
ci: add supply chain security scanning and pin action SHAs

wowi42:
Add supply chain attack protection to CI pipeline.

## Changes

- **Pin all GitHub Actions to commit SHAs** in test.yml and docs.yml
  - Mitigates tag-hijack supply chain attacks
  - Previously only pr-lint.yml had pinned actions
- **New supply-chain.yml workflow**
  - osv-scanner scans uv.lock for known vulnerabilities, uploads SARIF to Security tab
  - Bandit static analysis scans src/ for security anti-patterns (secrets, exec, SQLi)
  - CycloneDX SBOM generation via pip-audit, stored as build artifact
  - All jobs non-blocking (continue-on-error) for initial rollout
  - Runs on deps/code changes, weekly cron, and manual dispatch
- **New dependabot.yml config**
  - Weekly pip and github-actions updates
  - Grouped by production, test, and docs dependency types

## Motivation

Supply chain attacks on open source infrastructure tools are a growing threat.
pyinfra runs with elevated privileges on target systems: a compromised dependency
could be catastrophic. This PR adds defense-in-depth measures that:

- Block action tag hijacking (pinning SHAs)
- Detect known vulnerabilities in locked dependencies (osv-scanner)
- Catch security anti-patterns in source code (bandit)
- Provide transparent SBOM for auditability
- Keep dependencies current with automated, grouped updates

All tools used are open source and self-hostable.


Fizzadar (changes requested):
Love the intent here @wowi42! Few things though, and I think this should be split - it's really three PRs (SHA pinning, the scanning workflow, Dependabot); the pinning alone is a clean quick win we could land first.

On the pins: four are annotated-tag *object* SHAs rather than commit SHAs - `setup-uv`, `ruff-action`, `typos` and `codeql-action/upload-sarif`. They resolve today because the tags still point at them, but if upstream retags they drop out of the ref graph (already happened to `codeql` v4 - `411bbbe5` isn't referenced by any ref now, just not GC'd yet). Dependabot's `github-actions` updater tracks commit SHAs too, so it won't manage tag-object pins. Re-pin to the dereferenced commits.

Bit ironic but the scanner itself is the weak link: osv-scanner is `curl`'d unpinned from `releases/latest` with no checksum, in a job holding `security-events: write` - exactly the class this PR sets out to fix. Pin a release tag + verify a published checksum, or use `google/osv-scanner-action` pinned to a SHA.

Two more:
- the SARIF upload will red-X every fork PR touching `src/**` - fork PRs get a read-only token so `security-events: write` isn't granted and `upload-sarif` fails. Gate it on non-fork events.
- Dependabot uses `package-ecosystem: "pip"` but we're uv-managed (it won't touch `uv.lock`), and the `docs`/`test` group patterns are stale (`sphinx*` -> `zensical`, `pyinfra-testgen` -> `pyinfra-testing`), so those updates fall through ungrouped.

Whether we want Dependabot on at all is probably a separate call, historically I have found it more annoying/demanding than useful.

PR #1789, opened 2026-06-06
feat: add --exclude CLI argument to restrict target hosts

### LINKED ISSUE #1381: Inverse `limit`

I would like to be able to use an `--exclude` filter, same as the `--limit`, but with the reverse functionality.

e.g. `pyinfra inventory/myinventory.py main.py --exclude db1` would _exclude_ the db1 server from execution.

An implementation just like --limit, but to exclude servers.

mvanhorn:
## Summary

Adds an `--exclude` CLI argument that removes hosts from the execution set, the inverse of `--limit`. It accepts the same matching forms as `--limit` (exact host name, glob pattern, group name), can be passed multiple times, warns on patterns that match nothing, and applies after limits so `--limit app_servers --exclude app-2.net` works the way you'd expect.

## Why this matters

#1381 asks for exactly this: "I would like to be able to use an `--exclude` filter, same as the `--limit`, but with the reverse functionality." Today the only way to skip one host is to enumerate every other host via `--limit`, which doesn't scale on larger inventories.

The implementation extracts the existing `--limit` pattern-matching into a shared helper so both flags resolve names, globs, and groups identically, then `_apply_inventory_exclude` subtracts the matches from the limited set, or from the full inventory when no limit is given.

## Testing

New unit tests cover exclude by host name, glob, and group, exclude combined with `--limit`, the exclude-everything edge, and the no-match warning path. A new `--json` end-to-end test asserts an excluded host is absent from operation results. Full suite: 1816 passed. `ruff check`, `ruff format`, and `mypy` are clean on the changed files, and `scripts/lint_arguments_sync.py` reports in sync. CLI docs gained an Exclude section mirroring the Limit examples.

Fixes #1381

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -98,6 +98,24 @@ pyinfra inventory.py deploy.py --limit "db*"
 pyinfra inventory.py deploy.py --limit app_servers --limit db-1.net
 ```
 
+### Exclude
+
+It is possible to exclude hosts from the inventory at execution time using the `--exclude` argument. Multiple `--exclude`s can be provided. The value must either match a specific host by name or via glob style pattern, or match a group name:
+
+```sh
+# Execute against all hosts except db-1.net
+pyinfra inventory.py deploy.py --exclude db-1.net
+
+# Execute against all hosts except hosts in the `db_servers` group
+pyinfra inventory.py deploy.py --exclude db_servers
+
+# Execute against all hosts except names matching db*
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
Really nice, LGTM! Thank you @mvanhorn

mvanhorn:
Glad to get the exclude flag in as well, thanks @Fizzadar. As the inverse of limit, it makes targeting host subsets much easier.

PR #1791, opened 2026-06-06
fix(facts.zfs,facts.server.KernelModules): zfs operations broken on freebsd

jgelens:
Commit 5698bf1 introduced a precondition check which breaks the ZFS operations on FreeBSD. I updated the fact used by this check to include FreeBSD support which fixes this problem.

### DIFF
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -567,28 +567,45 @@ class KernelModules(FactBase):
 
     @override
     def command(self):
-        return "! test -f /proc/modules || cat /proc/modules"
+        self._kernel = host.get_fact(Kernel)
+
+        if self._kernel.strip() == "FreeBSD":
+            return "kldstat | tail +2"
+        else:
+            return "! test -f /proc/modules || cat /proc/modules"
 
     default = dict
 
     @override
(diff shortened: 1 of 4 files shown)


suzunn:
I noticed `KernelModules.process()` now depends on `self._kernel` being set as a side effect of `command()`. That is probably fine in the normal fact execution path, but it makes the parser a little easier to break from fixture-style tests or any future reuse that calls `process()` directly.

I would consider either deriving a normalized `is_freebsd` value through one small helper used by both methods, or guarding `process()` with `getattr(self, "_kernel", "")` and keeping the Linux parser as the default. A regression case for FreeBSD with ZFS absent would also be useful here, because the important behavior is not just parsing `zfs.ko`, but making sure the precondition still fails cleanly when neither `zfs` nor `zfs.ko` is present.

jgelens:
Sounds good, but I would probably do that in a different PR, could you create a ticket for that?
The check I used is done the same way in other Facts as well. So better to change these all at once. The commit I spoke off in the description breaks stuff I use daily, so better to merge this first.

Fizzadar (approved):
LGTM, thanks @jgelens!

PR #1792, opened 2026-06-08
fix(operations.zfs): avoid mutable default properties leaking across operation calls

### LINKED ISSUE #1781: zfs.dataset properties "leak" across multiple calls - mutable default
## Describe the bug
[`zfs.dataset`](https://docs.pyinfra.com/en/3.x/operations/zfs.html#zfs-dataset) properties provided via `**extra_props` "leak" across multiple calls of `zfs.dataset` via `properties` parameter, which defaults as an empty dict (`properties={}`)

## To Reproduce

here is a minimal example to reproduce (bare metal server with some zfs storage):

```python
from pyinfra.operations import zfs
# parent dataset
zfs.dataset(dataset_name="tank/home")
# children dataset 
zfs.dataset(dataset_name="tank/home/user1", quota="1T")
zfs.dataset(dataset_name="tank/home/user2", quota="1T")
```
briefly - we are creating a zfs dataset hierarchy for user's home folders on a shared storage server. We provide `quota` for each user using `**extra_props`, like so `quota="1T"` (only for "children" datasets inside `tank/home` - that's the intention). However, `extra_props` is added/updated to the `properties` parameter https://github.com/pyinfra-dev/pyinfra/blob/298f356130c6266ee53bdd29cf9144ee622fbbef/src/pyinfra/operations/zfs.py#L50 , which is defaulted as an empty dict `properties={}` - which is [mutable](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) ...

since we're calling `zfs.dataset` multiple times and moreover in 2 passes (i'm not 100% sure on how pyinfra works here) - that `extra_props` supplied `quota` parameter "leaks" everywhere, including the parent dataset `"tank/home"`, which is not the intention ...

## Expected behavior

provided above code should set `quota` for "children" datasets only , and leave the parent's `quota` default ...

below code works as expected but is more annoying to type :
```python
from pyinfra.operations import zfs
# parent dataset
zfs.dataset(dataset_name="tank/home")
# children dataset 
zfs.dataset(dataset_name="tank/home/user1", properties={"quota":"1T"})
zfs.dataset(dataset_name="tank/home/user2", properties={"quota":"1T"})
```

also if I understand correctly python functions should not default to a mutable arguments (unless it is intentional ?!) and a [pythonic way to go around that is:](https://stackoverflow.com/questions/13796199/what-is-the-most-pythonic-way-to-use-an-empty-dictionary-as-a-default-argument)

```python
def func(
   # ...
   param=None,
):
    # ...
    if param is None:
        param = {}  # this will be empty every time we call func
```
I'd be happy to PR something along these lines if maintainers agree/interested ...
or maybe llms could fix something like that in 5 mins ... because this pattern is present in *all* operations in the "zfs" module ...

## Meta

```sh
(infra) ➜  infra_new git:(main) ✗ pyinfra --support

    System: Darwin
      Platform: macOS-15.7.7-arm64-arm-64bit-Mach-O
      Release: 24.6.0
      Machine: arm64
    pyinfra: v3.8.0
      click: v8.3.1
      distro: v1.9.0
      gevent: v25.9.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.4.4
      types-paramiko: v4.0.0.20260508
      typing-extensions: v4.15.0
    Executable: /Users/sergpolly/Desktop/hardware_config/infra/.pixi/envs/default/bin/pyinfra
    Python: 3.14.0 (CPython, Clang 20.1.8 )
```

`pyinfra --debug` output with the proof that it did indeed set the quota on the parent dataset:

```sh
--> Starting operation: Ensure tank/home parent
    [umbra]  Skipped
    [shadow] Skipped
    [pyinfra.api.operations] Starting operation {'Ensure tank/home parent'} on epyca
    [pyinfra.api.facts] Getting fact: zfs.ZfsDatasets () (ensure_hosts: None)
    [pyinfra.api.facts] Getting fact: server.KernelModules () (ensure_hosts: None)
    [pyinfra.connectors.ssh] Running command on epyca: (pty=False) env SUDO_ASKPASS=/tmp/pyinfra-sudo-askpass-xST8FnMqbyzP *** sudo -H -A -k sh -c '! test -f /proc/modules || cat /proc/modules'
    [chasm]  Skipped
    [pyinfra.connectors.ssh] Waiting for exit status...
    [pyinfra.connectors.ssh] Command exit status: 0
    [pyinfra.api.facts] [epyca]  Loaded fact server.KernelModules
    [pyinfra.connectors.ssh] Running command on epyca: (pty=False) env SUDO_ASKPASS=/tmp/pyinfra-sudo-askpass-xST8FnMqbyzP *** sudo -H -A -k sh -c 'if command -v zfs >/dev/null 2>&1; then ( zfs get -H all ); else echo '"'"'##PYINFRA_NOCMD##'"'"'; fi'
    [pyinfra.connectors.ssh] Waiting for exit status...
    [pyinfra.connectors.ssh] Command exit status: 0
    [pyinfra.api.facts] [epyca]  Loaded fact zfs.ZfsDatasets
    [pyinfra.connectors.ssh] Running command on epyca: (pty=False) env SUDO_ASKPASS=/tmp/pyinfra-sudo-askpass-xST8FnMqbyzP *** sudo -H -A -k sh -c 'zfs set quota=500G tank/home'
    [pyinfra.connectors.ssh] Waiting for exit status...
    [pyinfra.connectors.ssh] Command exit status: 0
    [epyca]  Success
```

mvanhorn:
## Summary
ZFS operation `properties` no longer leak across calls. The mutable `{}` default argument is replaced with `None` plus a fresh dict per call.

## Why this matters
@sergpolly reported in #1781 that `zfs.dataset` properties leak across multiple calls. The operations used `properties={}` as a default argument, so every call without an explicit `properties` shared one dict, and `properties.update(extra_props)` mutated that shared object, carrying keys from one invocation into the next. This is the classic mutable-default-argument bug.

## Changes
- `dataset`, `snapshot`, `volume`, and `filesystem`: change `properties={}` to `properties=None` and build a fresh dict with `{**(properties or {}), **extra_props}` instead of mutating the default in place.

## Testing
Added a test fixture exercising extra props on `set` so the merge path is covered. `ruff check` passes; module imports cleanly.

Fixes #1781

AI was used for assistance.

### DIFF
--- a/src/pyinfra/operations/zfs.py
+++ b/src/pyinfra/operations/zfs.py
@@ -14,7 +14,7 @@ def dataset(
     recursive=False,
     sparse=None,
     volume_size=None,
-    properties={},
+    properties=None,
     **extra_props,
 ):
     """
@@ -47,7 +47,7 @@ def dataset(
 
     noop_msg = f"{dataset_name} is already {'present' if present else 'absent'}"
 
-    properties.update(extra_props)
+    properties = {**(properties or {}), **extra_props}
 
     datasets = host.get_fact(ZfsDatasets)
 
@@ -85,7 +85,7 @@ def dataset(
 
 
 @operation()
-def snapshot(snapshot_name, present=True, recursive=False, properties={}, **extra_props):
+def snapshot(snapshot_name, present=True, recursive=False, properties=None, **extra_props):
     """
     Create or destroy a ZFS snapshot, or modify its properties.
 
@@ -102,7 +102,7 @@ def snapshot(snapshot_name, present=True, recursive=False, properties={}, **extr
         zfs.snapshot("tank/home@weekly_backup")
 
     """
-    properties.update(extra_props)
+    properties = {**(properties or {}), **extra_props}
     snapshots = host.get_fact(ZfsSnapshots)
 
     if snapshot_name in snapshots or not present:
@@ -117,7 +117,7 @@ def snapshot(snapshot_name, present=True, recursive=False, properties={}, **extr
 
 @operation()
 def volume(
-    volume_name, size, sparse=False, present=True, recursive=False, properties={}, **extra_props
+    volume_name, size, sparse=False, present=True, recursive=False, properties=None, **extra_props
 ):
     """
     Create or destroy a ZFS volume, or modify its properties.
@@ -137,7 +137,7 @@ def volume(
         zfs.volume("tank/vm-disks/db_srv_04", "32G")
 
     """
-    properties.update(extra_props)
+    properties = {**(properties or {}), **extra_props}
     yield from dataset._inner(
         volume_name,
         volume_size=size,
@@ -149,7 +149,7 @@ def volume(
 
 
 @operation()
-def filesystem(fs_name, present=True, recursive=False, properties={}, **extra_props):
+def filesystem(fs_name, present=True, recursive=False, properties=None, **extra_props):
     """
     Create or destroy a ZFS filesystem, or modify its properties.
 
@@ -166,7 +166,7 @@ def filesystem(fs_name, present=True, recursive=False, properties={}, **extra_pr
         zfs.filesystem("tank/vm-disks/db_srv_04", "32G")
 
     """
-    properties.update(extra_props)
+    properties = {**(properties or {}), **extra_props}
     yield from dataset._inner(
         fs_name,
         present=present,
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
Nice catch, thanks @mvanhorn!

mvanhorn:
Thanks @Fizzadar for the merge on the ZFS fix. Swapping the mutable default properties dict stops the cross-call leak.

PR #1793, opened 2026-06-08
feat: preserve package extras and make pip/pipx/uv installs extras-aware

blenessy:
PkgInfo.from_pep508 parsed specs with packaging.Requirement but kept only reqt.name, silently dropping extras - so `foo[bar]` was installed as `foo`.

Keep extras on PkgInfo (as the raw frozenset, rendered only at point of use) and include them in the install command, while the installed-package lookup stays keyed by the bare name. The uninstall path uses a new extras-free `uninst_vers`, since extras are not separately uninstallable.

For idempotency, rather than reimplementing a resolver, ask each package manager's own resolver via --dry-run whether an already-installed bare package satisfies the requested extras:

  - pip / uv pip: `<tool> install --dry-run`
  - pipx:         `pipx runpip <app> install --dry-run`
  - uv tool:      probe the tool venv with `uv pip install --dry-run --python`
                  (uv tool install has no --dry-run); reinstall with --force

The dry-run check runs only when a spec carries extras, so plain installs are unaffected and stay offline. pip >= 22.2 is required on the extras path (for --dry-run/--report); pipx and uv tool reinstall missing extras with --force.

### DIFF
--- a/src/pyinfra/facts/pip.py
+++ b/src/pyinfra/facts/pip.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
+import re
+
 from typing_extensions import override
 
-from pyinfra.api import FactBase
+from pyinfra.api import FactBase, QuoteString, StringCommand
+from pyinfra.api.command import make_formatted_string_command
 
-from .util.packaging import parse_packages
+from .util.packaging import parse_packages, pip_report_is_satisfied
 
 PIP_REGEX = r"^([a-zA-Z0-9_\-\+\.]+)==([0-9\.]+[a-z0-9\-]*)$"
+PIP_VERSION_REGEX = r"^pip\s+(\S+)"
 
 
 class PipPackages(FactBase):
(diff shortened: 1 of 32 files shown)


wowi42 (commented):
Solid work overall: extras were genuinely dropped by `PkgInfo.from_pep508`, quoting is done properly through `QuoteString`, the dry-run resolver approach for idempotency is reasonable, and CI is green. A few things to address:

- When a spec has extras, the normal "installed" branch in `ensure_packages` is skipped entirely, so `pip.packages(["foo[bar]"], latest=True)` becomes a silent noop once extras are satisfied. Either handle the upgrade path or document the limitation.
- The pip >= 22.2 gate fires whenever any spec carries extras, even for a fresh install that old pip would handle natively. Please only gate when an extras spec hits an already-installed package (i.e. fetch `PipPackages` first).
- `UvToolVenvDryRun` probes `$(uv tool dir)/<name>/bin/python` with `pkg.name`, which is lowercased but not PEP 503 normalized. Names with `_`, `.` or uppercase (uv normalizes `Foo_Bar` to `foo-bar`) will probe the wrong path, fail the dry-run, and cause a `--force` reinstall on every run. Also the `--python` path is unquoted, so a tool dir with spaces breaks.
- The new fact classes leave `command()`/`process()` parameters unannotated; the project requires full type hints on FactBase methods.
- `PkgInfo.extras` was inserted as the 5th positional field before `inst_vers_format_fn`. In-repo callers are safe, but any out-of-tree positional use silently mis-assigns. Consider keyword-only or appending at the end.
- New fixture `extras_require_modern_pip.json` should be YAML (convention for new fixtures).

Heads up, not a blocker: #1823 also edits `ensure_packages` in `operations/util/packaging.py` (signature and present-loop), so whichever lands second will need a rebase.

PR #1794, opened 2026-06-09
fix(api): allow facts to process empty return values

Necavi:
Summary
========

Currently the example at [Writing Facts](https://docs.pyinfra.com/en/3.x/api/facts.html#example-getting-swap-status) fails silently. While this can be worked around using additional shell commands (e.g. using `swapon --show | wc -c` ), this seems like a more complicated solution that runs counter to the expected, and documented, way of interacting with the facts API.

Changes
=======

* Remove `stdout_lines` existence check from the `_get_fact` function in `pyinfra.api.facts`

Possible Issues
============

A very small number of facts may depend on the current contra-documented behavior and may throw errors if they are not expecting `output` to be empty.

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -348,20 +348,19 @@ def _get_fact(
     data = fact.default()
 
     if status:
-        if stdout_lines:
-            try:
-                data = fact.process(stdout_lines)
-            except FactProcessError as e:
-                log_error_or_warning(
-                    host,
-                    global_kwargs["_ignore_errors"],
-                    description=(f"could not process fact: {name} {get_kwargs_str(fact_kwargs)}"),
-                    exception=e,
-                )
+        try:
+            data = fact.process(stdout_lines)
+        except FactProcessError as e:
+            log_error_or_warning(
+                host,
+                global_kwargs["_ignore_errors"],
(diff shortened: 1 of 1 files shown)


wowi42 (changes requested):
The docs/code mismatch you found is real and worth fixing, but removing the `if stdout_lines:` guard as-is breaks the codebase. CI confirms it: every unit-test and end-to-end job fails.

Blocking issues:

1. Many facts index into output unconditionally (`output[0]` / `output[-1]` in `files.py`, `server.py`, `selinux.py`, `lxd.py`, `git.py`, `win_files.py`, ...). `_get_fact` only catches `FactProcessError`, so `process([])` on these raises `IndexError` and crashes the run. This is what the failing CI shows.
2. The default-value contract changes silently: facts with empty output now return whatever `process([])` produces instead of `default()`. For facts that depend on default-when-empty this changes return values without any error, which is worse than a crash.
3. No test covers the intended new behavior; the 8 failing tests are left red. A change to core fact semantics needs deliberate test updates in `tests/test_api/test_api_facts.py`, and an audit of every fact's `process()` for empty-input tolerance (or a different fix, e.g. changing the docs example to emit a marker line).

Happy to re-review after a rework, but this needs the full audit, not the one-line removal.

PR #1796, opened 2026-06-10
fix(facts.deb): distinguish package names from deb paths

### LINKED ISSUE #1795: Fact deb.DebPackage incorrectly interprets package names as paths (and vice versa)
Executing `host.get_fact(deb.DebPackage, 'somename')` will throw an exception if there happens to be a directory or file `somename` (that is not actually a deb package file) in the current working directory. It runs approximately the following code, which executes the `dpkg -I` branch:

    ( ! test -e not_a_package \
        && (dpkg -s not_a_package 2>/dev/null || true) \
        || dpkg -I not_a_package
    )

This in fact makes it _impossible_ to test an installed package if there happens to be a directory or file with the same name in the CWD of pytest when it accesses that host.

This also causes a (less likely) problem the other way around: if I e.g. download a package to `dpkg.deb` but typo the `get_fact` to check `'dpkg'` it will not tell me that I got the filename wrong but instead successfully return information for the installed `dpkg` package. (This is likely to be particularly confusing because it's returning package information, but the wrong information.)

Idempotent configuration systems should not rely on particular existing state but instead correctly do the requested thing regardless of system state. This API breaks that: instead of letting the user specify whether they're trying to examine an installed package or a package file, it changes which it examines based on the state of the host.

### Fix and Consequences

My suggested fix for this API is always to use `dpkg -s` if the name given does not have any slashes in it, and always use `dpkg -I` if the name does have slashes in it. If the user wants to specify a path in the current working directory, they can specify `./dpkg` instead of just `dpkg`.

This change will not affect the following common cases:
1. Anybody querying an installed package by name; it will still query the package by name. (Package names never have slashes in them.)
2. Anybody querying a package file that has a relative or absolute path (e.g., if a user is checking `'/tmp/foo.deb'` or whatever).

It will affect users who are trying to query a file in the CWD. (This might happen if they've downloaded a file to CWD instead of to `/tmp/`.) Most likely this will fail, as the file they're trying to query probably will end with `.deb` and so will never match an installed package. This isn't ideal, but given that this failure is exactly what we want the API to produce, there's no way around it. It is, however, fairly easy for the end user to fix, especially if we have the error message make it clear that it's querying an installed package and the user should prefix the name with `./` if they are trying to query a file.

### System Configuration

Pyinfra was installed into a virtual environment with the equivalent of `pip install pyinfra`, giving:

    $ pae . pip list | grep -i pyinfra
    pyinfra           3.9.2

`pyinfra --support` output:

    System: Linux
      Platform: Linux-7.0.11-100.fc43.x86_64-x86_64-with-glibc2.42
      Release: 7.0.11-100.fc43.x86_64
      Machine: x86_64
    pyinfra: v3.9.2
      click: v8.4.1
      distro: v1.9.0
      gevent: v26.5.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.15.0
    Executable: ./pyinfra
    Python: 3.14.5 (CPython, GCC 15.2.1 20260123 (Red Hat 15.2.1-7))

Jia-Ethan:
## Summary

`deb.DebPackage` now chooses between installed-package lookup and `.deb` archive inspection from the argument shape instead of remote filesystem state.

Package names without a slash use `dpkg -s`, while paths such as `/tmp/package.deb` or `./package.deb` use `dpkg -I`. This avoids treating an installed package name as a local archive just because a file or directory with the same name exists in the remote working directory.

## What changed

- `src/pyinfra/facts/deb.py`: use `dpkg -s` for slash-free package names and `dpkg -I` for path-like arguments.
- `tests/facts/deb.DebPackage/*`: update the existing command fixtures and add a `./local-package.deb` fixture for the explicit current-directory archive case.

## Testing

```bash
PYTHONPATH= scripts/dev-test.sh
# 1823 passed, 15 deselected, 7 warnings

PYTHONPATH= scripts/dev-lint.sh
# ruff, format, mypy, and arguments sync passed
```

Closes #1795

## AI usage disclosure

This pull request was prepared with AI assistance from OpenAI Codex. I reviewed the issue, repository guidance, code changes, tests, and verification output before submission.

### DIFF
--- a/src/pyinfra/facts/deb.py
+++ b/src/pyinfra/facts/deb.py
@@ -73,8 +73,14 @@ class DebPackage(FactBase):
 
     @override
     def command(self, package):
+        if "/" in package:
+            return make_formatted_string_command(
+                "dpkg -I {0}",
+                QuoteString(package),
+            )
+
         return make_formatted_string_command(
-            "! test -e {0} && (dpkg -s {0} 2>/dev/null || true) || dpkg -I {0}",
+            "dpkg -s {0} 2>/dev/null || true",
             QuoteString(package),
         )
 
--- a/tests/facts/deb.DebPackage/package.json
+++ b/tests/facts/deb.DebPackage/package.json
@@ -1,6 +1,6 @@
 {
     "arg": "package-name",
-    "command": "! test -e package-name && (dpkg -s package-name 2>/dev/null || true) || dpkg -I package-name",
+    "command": "dpkg -s package-name 2>/dev/null || true",
     "requires_command": "dpkg",
     "output": [
         "Package: package-name",
--- a/tests/facts/deb.DebPackage/package_needs_quotes.json
+++ b/tests/facts/deb.DebPackage/package_needs_quotes.json
@@ -1,6 +1,6 @@
 {
     "arg": "/tmp/my package; whoami.deb",
-    "command": "! test -e '/tmp/my package; whoami.deb' && (dpkg -s '/tmp/my package; whoami.deb' 2>/dev/null || true) || dpkg -I '/tmp/my package; whoami.deb'",
+    "command": "dpkg -I '/tmp/my package; whoami.deb'",
     "requires_command": "dpkg",
     "output": [
         "Package: my-package",
--- a/tests/facts/deb.DebPackage/relative_path.yaml
+++ b/tests/facts/deb.DebPackage/relative_path.yaml
@@ -1,0 +1,9 @@
+arg: ./local-package.deb
+command: dpkg -I ./local-package.deb
+requires_command: dpkg
+output:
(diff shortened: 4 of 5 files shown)


Fizzadar (commented):
Thanks for tackling #1795, @Jia-Ethan - closing in favour of #1829. The slash branch is a bare dpkg -I {0}, which exits non-zero when the  path doesn't exist yet and fails the host (state.fail_hosts). That breaks apt.deb(src="/tmp/x.deb") when the file's produced by an earlier op, 1829 fixes this whilst keeping the functionality.

PR #1797, opened 2026-06-10
fix(facts.deb): choose package command by argument path

### LINKED ISSUE #1795: Fact deb.DebPackage incorrectly interprets package names as paths (and vice versa)

Photon101:
## Summary

- Make `deb.DebPackage` choose `dpkg -s` for package-name lookups and `dpkg -I` for slash-bearing archive paths instead of branching on remote filesystem state.
- Preserve explicit current-directory archive lookup through `./package-name.deb` while treating bare `package-name.deb` as an installed package lookup.
- Add regression fixtures covering package names, absolute/quoted archive paths, and current-directory archive paths.

Closes #1795.

## Checks

- `uv run pytest tests/test_facts.py -k "deb.DebPackage"`
- `uv run pytest tests/test_facts.py`
- `uv run pytest tests/test_operations.py -k "apt.deb"`
- `scripts/dev-test.sh`
- `scripts/dev-lint.sh`

### DIFF
--- a/src/pyinfra/facts/deb.py
+++ b/src/pyinfra/facts/deb.py
@@ -73,8 +73,14 @@ class DebPackage(FactBase):
 
     @override
     def command(self, package):
+        if "/" in str(package):
+            return make_formatted_string_command(
+                "dpkg -I {0}",
+                QuoteString(package),
+            )
+
         return make_formatted_string_command(
-            "! test -e {0} && (dpkg -s {0} 2>/dev/null || true) || dpkg -I {0}",
+            "dpkg -s {0} 2>/dev/null || true",
             QuoteString(package),
         )
 
--- a/tests/facts/deb.DebPackage/archive_name_without_slash.json
+++ b/tests/facts/deb.DebPackage/archive_name_without_slash.json
@@ -1,0 +1,7 @@
+{
+    "arg": "package-name.deb",
+    "command": "dpkg -s package-name.deb 2>/dev/null || true",
+    "requires_command": "dpkg",
+    "output": [],
+    "fact": {}
+}
--- a/tests/facts/deb.DebPackage/current_directory_archive.json
+++ b/tests/facts/deb.DebPackage/current_directory_archive.json
@@ -1,6 +1,6 @@
 {
-    "arg": "package-name",
-    "command": "dpkg -s package-name 2>/dev/null || true",
+    "arg": "./package-name.deb",
+    "command": "dpkg -I ./package-name.deb",
     "requires_command": "dpkg",
     "output": [
         "Package: package-name",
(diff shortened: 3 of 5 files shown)


Fizzadar (commented):
Thanks @Photon101 - closing in favour of #1829. Same approach as #1796 and the same regression: the dpkg -I {0} slash branch exits non-zero  for a not-yet-existing path. 1829 solves this without that. Thank you for looking into this!

Issue #1760, opened 2026-05-18
Security audit: unquoted user input in shell commands (command injection hardening)


wowi42:
# Critical Security Issues - pyinfra Codebase

## Overview

Multiple instances where user-supplied values (reachable from public operation/fact arguments) are interpolated into shell commands via f-strings/`.format()` without `QuoteString` from `pyinfra.api.command`, enabling command injection.

> Verified against `pyinfra-dev/pyinfra@d3bc74b5` (branch `3.x`) on 2026-05-18. Line numbers below are current as of that commit. Items reclassified or downgraded during verification are annotated inline.

This issue tracks remediation as a checklist. An item is "done" when the user-supplied values in that file are composed via `StringCommand` + `QuoteString` / `MaskString` instead of raw interpolation.

## Shell command injection (confirmed)

- [ ] **1. `src/pyinfra/operations/util/docker.py`**
  - Scope: user values unquoted throughout (lines ~182-411)
  - Risk: container names, images, ports, volumes, env vars, networks, plugins, etc. (all from public `docker.*` operation args / `ContainerSpec`) are interpolated into `docker` commands via f-strings. `QuoteString` is never imported in this file.
  - Example: `f"docker container create --name {kwargs['container']}"` (**line 263**)

- [ ] **2. `src/pyinfra/facts/server.py`**
  - Scope: facts return unquoted commands (lines 41, 202, 365/372/376, 854-855, 1284-1285)
  - Risk: `user`, `command`, `port`, `protocol`, `path`, `pid` fact parameters are unquoted. No `QuoteString` in file.
  - Example: `return f"echo ~{user}"` (line 41)

- [ ] **3. `src/pyinfra/operations/server.py`**
  - Scope: lines 463/465/466 and 1203
  - Risk: `timezone`, `signal`, `pid` operation arguments interpolated raw (despite `QuoteString` being used elsewhere in this file).
  - Example: `yield f"timedatectl set-timezone {timezone}"` (line 463)

- [ ] **4. `src/pyinfra/operations/vzctl.py`**
  - Scope: lines 26, 39, 63, 85, 108, 137 (also line 74, not originally listed)
  - Risk: `ctid`, `template`, and `**settings` keys/values unquoted. No `QuoteString` in file.
  - Example: `yield f"vzctl start {' '.join(args)}"` (line 26)

- [ ] **5. `src/pyinfra/operations/lxd.py`**
  - Scope: lines 56, 59, 67
[...]

PR #1810, opened 2026-06-10
fix(operations.iptables): quote user input in iptables commands

wowi42:
## Describe the bug
`iptables.chain` and `iptables.rule` interpolate `table`, `chain`, `policy` and every rule field (protocol, source, destination, interfaces, jump, log_prefix, to_destination/source/ports) into the iptables/ip6tables commands with f-strings and `" ".join`.

## To Reproduce

+ Operation code & usage

```python
from pyinfra.operations import iptables

iptables.chain(chain='x; reboot')
```

## Expected behavior
Every user value is shell-escaped. `extras` stays raw as it is a multi-token passthrough of extension args. The crafted input is passed to the program as a single literal argument instead of being interpreted by the shell. A new injection fixture covers this.

## Meta

### DIFF
--- a/src/pyinfra/operations/iptables.py
+++ b/src/pyinfra/operations/iptables.py
@@ -5,7 +5,7 @@ The iptables modules handles iptables rules
 from __future__ import annotations
 
 from pyinfra import host
-from pyinfra.api import operation
+from pyinfra.api import QuoteString, StringCommand, operation
 from pyinfra.api.exceptions import OperationError
 from pyinfra.facts.iptables import Ip6tablesChains, Ip6tablesRules, IptablesChains, IptablesRules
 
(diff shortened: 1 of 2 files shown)


Fizzadar (changes requested):
The operation-side rewrite a great improvement. But the injection is still open - the `table` value goes through the facts, not the operation. Both ops call `host.get_fact(IptablesChains/IptablesRules, table=table)` before yielding any command, and all four facts still interpolate it raw:

```python
def command(self, table="filter"):
    return f"iptables-save -t {table}"
```

so `iptables.chain("INPUT", table="filter; reboot")` still runs the payload on the target - quoting `-t {table}` in the operation closes nothing for `table`.

(while we're there we might as well handle dport/sport as well)

PR #1815, opened 2026-06-10
fix(operations.sysvinit): quote user input in service enable/disable

wowi42:
## Describe the bug
`sysvinit.service` interpolates `service` into the update-rc.d, chkconfig and rc-update enable commands, and the discovered link path into `rm -f`, all with f-strings.

## To Reproduce

+ Operation code & usage

```python
from pyinfra.operations import sysvinit

sysvinit.service(service='x; reboot', enabled=True)
```

## Expected behavior
The `service` and link path are shell-escaped. The crafted input is passed to the program as a single literal argument instead of being interpreted by the shell. A new injection fixture covers this.

## Meta

### DIFF
--- a/src/pyinfra/facts/files.py
+++ b/src/pyinfra/facts/files.py
@@ -489,6 +489,7 @@ class FindFilesBase(FactBase):
         fname: str | None = None,
         iname: str | None = None,
         regex: str | None = None,
+        path_match: str | None = None,
         args: list[str] | None = None,
         quote_path=True,
     ):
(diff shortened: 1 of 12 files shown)


wowi42:
While hardening this @Fizzadar  found another hole in the same operation. The enable/disable path called FindLinks with quote_path=False so the glob could expand:

    start_links = host.get_fact(
        FindLinks,
        path=f"/etc/rc*.d/S*{service}",
        quote_path=False,  # enable path glob matching
    )

Turning off quoting to keep glob matching also lets the shell interpret a crafted service, so sysvinit.service("x; reboot", enabled=False) would run reboot.

Fixed by adding a path_match parameter to FindFilesBase that emits find -path with the pattern always shell-quoted. The shell passes it literally and find itself does the globbing, so glob matching stays but injection is gone:

    find /etc -type l -path '/etc/rc*.d/S*x; reboot' || true

This touches the shared FindFilesBase fact, so it is a bit wider than the original sysvinit-only change. Happy to split the fact parameter into its own PR if you prefer. Added fact and operation fixtures covering the injection. Tests and lint pass.

DonDebonair:
I was the one who found the hole in the operation ;)

Will review now. Did you do an end2end test with this, to see if the adjusted `FindLinks` fact still works as intended?

wowi42:
@DonDebonair my bad

DonDebonair (approved):

PR #1816, opened 2026-06-10
fix(operations.vzctl): quote user input in vzctl commands

wowi42:
## Describe the bug
`vzctl.*` interpolates `ctid`, `template` and `**settings` values into the vzctl commands with f-strings (no `QuoteString` in the file), so crafted values run arbitrary shell.

This PR fixes it by composing the command(s) with `StringCommand` and wrapping the user-supplied values in `QuoteString` (from `pyinfra.api`). Part of the #1760 shell-injection hardening (item 4).

## To Reproduce

+ Operation code & usage

```python
from pyinfra.operations import vzctl

vzctl.set(ctid='1', hostname='x; reboot')
```

## Expected behavior
The `ctid`, `template` and settings values are shell-escaped. Settings keys come from `**settings` so they are Python identifiers and need no quoting. The crafted input is passed to the program as a single literal argument instead of being interpreted by the shell. A new injection fixture covers this.

## Meta

### DIFF
--- a/src/pyinfra/operations/vzctl.py
+++ b/src/pyinfra/operations/vzctl.py
@@ -5,7 +5,7 @@ Manage OpenVZ containers with ``vzctl``.
 from __future__ import annotations
 
 from pyinfra import host
-from pyinfra.api import OperationError, operation
+from pyinfra.api import OperationError, QuoteString, StringCommand, operation
 from pyinfra.facts.vzctl import OpenvzContainers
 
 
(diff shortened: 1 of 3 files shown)


src/pyinfra/operations/vzctl.py
```diff
@@ -122,16 +120,19 @@ def set(ctid: str, save=True, **settings):
...
     """
 
-    args = [f"{ctid}"]
+    command: list[str | QuoteString] = ["vzctl set", QuoteString(str(ctid))]
 
     if save:
-        args.append("--save")
+        command.append("--save")
 
+    # `key` comes from **settings so it is a Python identifier and safe; only the
```
DonDebonair: I don't understand this. The keys in `settings` are user-constrolled as well. I could pass `{";reboot": "whatever"}` and it would render the command as `--;reboot 'whatever'`, which would still reboot the machine?

wowi42: Good catch, you are right. The keys are user-controlled too, dict unpacking does not constrain them to identifiers. I now wrap the flag as QuoteString(f"--{key}"), so {";reboot": "x"} renders as vzctl set 123 --save '--;reboot' x instead of breaking out. shlex.quote leaves normal flags like --hostname bare, so existing fixtures are unchanged. Added a fixture with a hostile key and dropped the wrong comment.

DonDebonair (approved):

PR #1817, opened 2026-06-10
fix(operations.yum): quote user input in yum.key

wowi42:
## Describe the bug
`yum.key` interpolates the user-supplied `src` into `rpm --import` with an f-string, so a crafted `src` runs arbitrary shell.

## To Reproduce

+ Operation code & usage

```python
from pyinfra.operations import yum

yum.key(src='https://example.com/key; reboot')
```

## Expected behavior
The `src` is shell-escaped. The crafted input is passed to the program as a single literal argument instead of being interpreted by the shell. A new injection fixture covers this.

## Meta


DonDebonair (approved):
LGTM

PR #1818, opened 2026-06-10
fix(operations.zfs): quote user input in zfs commands

wowi42:
## Describe the bug
`zfs.dataset` and `zfs.snapshot` interpolate dataset/snapshot names and `prop=value` property pairs into the `zfs create`/`set`/`destroy`/`snap` commands with f-strings, so crafted values run arbitrary shell.

## To Reproduce

+ Operation code & usage

```python
from pyinfra.operations import zfs

zfs.dataset('tank/x; reboot', compression='lz4')
```

## Expected behavior
The names and property pairs are shell-escaped, preserving the deterministic argument ordering. The crafted input is passed to the program as a single literal argument instead of being interpreted by the shell. A new injection fixture covers this.

## Meta

### DIFF
--- a/src/pyinfra/operations/zfs.py
+++ b/src/pyinfra/operations/zfs.py
@@ -54,31 +54,24 @@ def dataset(
     existing_dataset = datasets.get(dataset_name)
 
     if present and not existing_dataset:
-        # (sort_key, command bits) so we keep the deterministic ordering of the old
-        # joined-and-sorted string while quoting the user-supplied values.
-        arg_specs: list[tuple[str, list[str | QuoteString]]] = [
-            (f"-o {prop}={value}", ["-o", QuoteString(f"{prop}={value}")])
-            for prop, value in properties.items()
-        ]
-        if recursive:
-            arg_specs.append(("-p", ["-p"]))
-        if sparse:
-            arg_specs.append(("-s", ["-s"]))
-        if volume_size:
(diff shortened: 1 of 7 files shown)


src/pyinfra/operations/zfs.py
```diff
@@ -54,31 +54,46 @@ def dataset(
     existing_dataset = datasets.get(dataset_name)
 
     if present and not existing_dataset:
-        args = [f"-o {prop}={value}" for prop, value in properties.items()]
+        # (sort_key, command bits) so we keep the deterministic ordering of the old
+        # joined-and-sorted string while quoting the user-supplied values.
+        arg_specs: list[tuple[str, list[str | QuoteString]]] = [
```
DonDebonair: I don't understand the need for this complexity - including the sort action on L70. Dicts in Python are ordered, so you can make a list out of `dict.items()` which will be deterministic in order.

DonDebonair (changes requested):
I think the whole sorting thing for props and args is not needed.

src/pyinfra/operations/zfs.py
wowi42: Dropped the sort and the tuple machinery, the create branch now builds args straight from dict.items(). For zfs set I replaced the items() set difference with a dict-ordered comprehension, since iteration order of a set difference is not deterministic.

DonDebonair (approved):

PR #1822, opened 2026-06-10
feat: add supervisorctl operations and facts

### LINKED ISSUE #1216: Feature request: supervisor (supervisorctl) operations
supervisor (supervisorctl) operations

supervisor (supervisorctl) operations

wowi42:
Closes #1216.

Adds a `supervisorctl` module with operations and facts for processes managed by supervisord, based on the documented CLI surface at https://supervisord.org/running.html.

## Operations (`pyinfra.operations.supervisorctl`)

- `process`: idempotent start/stop/restart against the status fact. Supports `<name>`, `<gname>:*` and `all` targets, plus a custom action via `command`.
- `reread`, `update` (optional group list), `reload`, `add`, `remove`, `clear`, `signal`: stateless operations marked `is_idempotent=False`.

## Facts (`pyinfra.facts.supervisorctl`)

- `SupervisorctlStatus`: dict of process name (including `group:name`) to running state. `True` for RUNNING/STARTING, `False` for STOPPED/STOPPING/EXITED, `None` for BACKOFF/FATAL/UNKNOWN.
- `SupervisorctlPid`: PID of supervisord itself, or of a single process. `0` means present but not running, `None` means unreadable.
- `SupervisorctlPids`: `pid all` as a list of ints.

## Notes

- Every fact and operation accepts the global ctl options: `config_file` (`-c`), `server_url` (`-s`), `username` (`-u`), `password` (`-p`). The password is wrapped in `HiddenValue` so it renders as `*MASKED*` in logs.
- `operations.util.service.handle_service_control` is not reused on purpose: its format-string API cannot carry a `HiddenValue` password without leaking it. The `process` operation mirrors its semantics instead.
- Read-only and interactive actions are intentionally not operations: `status` and `pid` map to facts, `tail`/`fg`/`help` are out of scope.
- Both modules are registered in `pyinfra-metadata.toml` and the docs generators produce their pages.

## Tests

- YAML fixtures covering all operations (start/stop/noop/restart, group wildcard, `all`, custom command, auth options) and all three facts (states, group names, pid variants, error output).
- Full suite, ruff, mypy and the arguments sync lint pass.


src/pyinfra/operations/supervisorctl.py
```diff
@@ -0,0 +1,357 @@
...
+            username=username,
+            password=password,
+        ),
+        "reload",
+    )
+
+
+@operation(is_idempotent=False)
+def add(
+    name: str,
```
Fizzadar, line 225: This clashes w/the global `name` arg and will never be set (we should test that!)

src/pyinfra/operations/supervisorctl.py
```diff
@@ -0,0 +1,357 @@
...
+            password=password,
+        ),
+        "add",
+        QuoteString(name),
+    )
+
+
+@operation(is_idempotent=False)
+def remove(
+    name: str,
```
Fizzadar, line 255: Same issue as above

Fizzadar (changes requested):
Two arg changes then looks good

PR #1825, opened 2026-06-11
docs: render files fact import as code

### LINKED ISSUE #1790: Minor doc fix
Minor doc fix, couldn't find source

I wanted to fix the beginning of this page:

https://docs.pyinfra.com/en/3.x/facts/files.html

There is the intro section:

```
Facts need to be imported before use, eg

from pyinfra.facts.files import File
```

That second line should be preformatted text, code, etc.  

I was going to fix it myself, but looked all over your repos and couldn't find the file.  Found  multiple copies of rendered html and .rst files but its not clear which is the original.

Later found a promising one, only to find the docs repo has pull requests and issues disabled.

Himanshuagrawal4:
### Summary
- wrap the files fact import example in a Python fenced code block

### Why
The generated facts documentation should render `from pyinfra.facts.files import File` as code instead of plain paragraph text.

Fixes #1790.

### Checks
- `git diff --check`
- `python -m py_compile src\pyinfra\facts\files.py`


wowi42:
LGTM, just rebase when you can, thanks !!!

Himanshuagrawal4:
Thanks! I rebased this and found that current 3.x already contains the equivalent RST code block for this import example, so the branch ended up with no remaining diff. Closing this as superseded.

PR #1431, opened 2025-08-12
Add shellcheck support (workflow to detect/not fix) and fix scripts to pass checks

yarikoptic:
More about shellcheck: https://www.shellcheck.net/

I personally introduced it to a good number of my and projects of others
already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Individual commits might have more information.

On 3.x branch you would get now ([shellcheckit](https://github.com/yarikoptic/improveit/blob/main/shellcheckit) is just a little helper I have to prep this PR etc)

```shell
❯ shellcheckit doit
pyinfra/connectors/util.py:#!/bin/sh
scripts/build-public-docs.sh:#!/bin/bash
scripts/dev-lint.sh:#!/usr/bin/env bash
scripts/dev-test-e2e.sh:#!/usr/bin/env bash
scripts/dev-test.sh:#!/usr/bin/env bash
scripts/release.sh:#!/bin/sh
scripts/spellcheck.sh:#!/usr/bin/env bash

In scripts/build-public-docs.sh line 29:
    env DOCS_VERSION=$BRANCH_NAME sphinx-build -a docs/ docs/public/en/$BRANCH_NAME/
                     ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                       ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
    env DOCS_VERSION="$BRANCH_NAME" sphinx-build -a docs/ docs/public/en/"$BRANCH_NAME"/


In scripts/build-public-docs.sh line 33:
        env DOCS_VERSION=$BRANCH_NAME python scripts/generate_redirect_pages.py
                         ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
        env DOCS_VERSION="$BRANCH_NAME" python scripts/generate_redirect_pages.py


In scripts/dev-test.sh line 6:
pytest $@
       ^-- SC2068 (error): Double quote array expansions to avoid re-splitting elements.


In scripts/release.sh line 3:
set -euo pipefail
         ^------^ SC3040 (warning): In POSIX sh, set option pipefail is undefined.


In scripts/release.sh line 5:
VERSION=`python setup.py --version`
        ^-------------------------^ SC2006 (style): Use $(...) notation instead of legacy backticks `...`.

Did you mean: 
VERSION=$(python setup.py --version)


In scripts/release.sh line 6:
[...]

PR #1827, opened 2026-06-11
ci: add shellcheck workflow and fix scripts

wowi42:
## What this does

Adds a Shellcheck CI workflow and fixes the shell scripts so the tree passes it.

This revives #1431 by @yarikoptic, which went stale against `3.x`. The scripts have drifted since then, so the changes are re-derived against the current tree rather than a replay of the old diff.

## Changes

- `.github/workflows/shellcheck.yml`: installs shellcheck and runs it on every tracked file that starts with a shell shebang or a shellcheck directive. Matches the repo conventions (`actions/checkout@v5`, `ubuntu-24.04`, same branch triggers as `test.yml`), with `permissions: contents: read`.
- `scripts/dev-test.sh`: `uv run pytest $@` to `"$@"` (SC2068, real argument re-splitting bug).
- `scripts/release.sh`: shebang `#!/bin/sh` to `#!/bin/bash` (SC3040, `pipefail` is undefined in POSIX sh) and legacy backticks to `$(...)` (SC2006).

## Note on the file selection

The original detection also matched `src/pyinfra/connectors/util.py`, a Python file that embeds a `#!/bin/sh` heredoc as a string. Shellcheck cannot parse that file as shell and errors out, which would fail CI. The workflow grep now excludes Python files (`':!*.py'`), so only real shell scripts are checked. The earlier `build-public-docs.sh` SC2086 fix from the original PR is no longer needed: that script was rewritten upstream and is already clean.

## Verification

- `git grep -l '...' -- ':!*.py' | xargs shellcheck` exits 0 over all matched files.
- Workflow YAML parses, and both edited scripts pass `bash -n`.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,26 @@
+name: Shellcheck
+
+on:
+  push:
+    branches:
+      - 3.x
+      - 2.x
+      - 1.x
+      - 0.x
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    name: Check shell scripts
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v5
+      - name: Install shellcheck
+        run: sudo apt-get update && sudo apt-get install -y shellcheck
+      - name: Run shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' -- ':!*.py' \
+            | xargs shellcheck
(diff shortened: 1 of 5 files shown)


DonDebonair:
I think it would be nice to also add `shellcheck` to `scripts/dev-lint.sh`, but that would require contributors to install it. I don't see that as a problem. What do you think @Fizzadar ?

Fizzadar:
@DonDebonair @wowi42 seems reasonable to me, we can note that in the contributing doc.

wowi42:
@DonDebonair @Fizzadar done in 2433f2bf. `scripts/dev-lint.sh` now runs shellcheck using the same file detection as the CI workflow, and fails with a clear message pointing to the docs if shellcheck is not installed. I also noted the new dependency in `docs/contributing.md` with `apt-get` / `brew` install hints.

Fizzadar (approved):
LGTM!

PR #1829, opened 2026-06-12
fix(facts.deb): DebPackage failing when file/directory with same name exists in cwd

bendigg:
## Problem

`host.get_fact(deb.DebPackage, "somename")` throws an exception if a directory or regular file named `somename` exists in the current working directory.

The root cause is that `test -e` matches directories AND regular files, causing `dpkg -I` to be incorrectly invoked on non-.deb targets.

## Fix

- Changed `test -e` to `test -f` (only match regular files)
- Added `case` statement to only invoke `dpkg -I` on `*.deb` files
- Falls back to `dpkg -s` for all other cases (package name lookup)

## Tests Added

- `directory_same_name.json` — package name matches a directory in cwd
- `non_deb_file.json` — regular non-.deb file as argument

All 15 deb-related tests pass (7 DebPackage + 8 DebPackages).

### DIFF
--- a/src/pyinfra/facts/deb.py
+++ b/src/pyinfra/facts/deb.py
@@ -74,7 +74,7 @@ class DebPackage(FactBase):
     @override
     def command(self, package):
         return make_formatted_string_command(
-            "! test -e {0} && (dpkg -s {0} 2>/dev/null || true) || dpkg -I {0}",
+            "test -f {0} && case {0} in *.deb) dpkg -I {0} ;; *) dpkg -s {0} 2>/dev/null || true ;; esac || dpkg -s {0} 2>/dev/null || true",
             QuoteString(package),
         )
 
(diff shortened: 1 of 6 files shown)


Fizzadar (approved):
Thanks @bendigg this looks great, keeps all the existing functionality 👌

0cjs:
Sorry about not mentioning this a couple of weeks ago back when this was current, but my issue on this, #1795 (dating a week earlier than this) was never mentioned on this PR until after long after this had been merged.

> Thanks @bendigg this looks great, keeps all the existing functionality 👌

This does _not_ keep all existing functionality. For example, you used to be able to use a file called `somepackage.deb.old`; you can no longer do that. (In fact, you can no longer use any filename at all unless it ends in `.deb`.)

Further, ___this is not idempotent.___ It still depends on the particular contents of the filesystem.

Since doing it the way this PR does was going to break backwards compatibility anyway, that's why I suggested a way (arguably causing slightly more breakage, but hard to say) that completely clarifies whether the caller is asking to work on a package file or an installed package. Any chance we could reconsider that solution?

If not, I think we at least need to document the current behaviour, as lack of documentation on this was what led me to my issues anyway. I have filed #1860 about this.

Issue #1828, opened 2026-06-11
Brew tap - Add the option to trust tap/formula


vaz-ar:
## Is your feature request related to a problem? Please describe
Homebrew 6.0.0 just released, and it brings changes regarding how tap are treated (https://brew.sh/2026/06/11/homebrew-6.0.0/).
Tap now require to be trusted for upgrade to be possible.

## Describe the solution you'd like
It would be helpful to have an option to trust the tap when adding it, or another brew operation to trust/untrust tap or formulaes

PR #1836, opened 2026-06-16
feat(operations.brew, facts.brew): add support for brew trust

morrison12:
Added support for the [brew trust](https://docs.brew.sh/Tap-Trust) that is now mandatory with Homebrew 6.0 (and beyond) as noted in #1828.

Support is:
1. a new parameter to `brew.tap` as it seems unlikely one would want add a tap without trusting it
2. a new `brew.trust` operation given one can choose to trust narrower scopes than the full tap, e.g. a specific cask.
3. the related fact: `brew.BrewTrusted`.

The diffs are rather noisy given the changes required to appease the static checkers and my choice to change `brew.tap` to use `StringCommand` et al.

ps ignore the mess from committing in the wrong place; I (think) I fixed that.


morrison12:
Ugh.. Forgot about python 3.10.  Back in a bit.

src/pyinfra/operations/brew.py
```diff
@@ -213,20 +238,75 @@ def tap(src: str | None = None, present=True, url: str | None = None):
...
+    + kind: whether the item is a CASK, COMMAND, FORMULA or TAP (using BrewItemKind enum)
+    + trusted: whether this item should be trusted or not.  no default, must be specified
+
+    **Examples:**
+
+    .. code:: python
+
+        brew.trust(
+            name="Mark magic tap as trusted",
+            item="includeos/includeos",
```
wowi42: The docstring example uses `item=` (singular) but the parameter is `items=` (plural), and `trust=True` but the parameter is `trusted`. Copy-pasting this example would fail with TypeError. The `+ item:` docstring line above the example also needs to become `+ items:`.

src/pyinfra/operations/brew.py
```diff
@@ -213,20 +238,75 @@ def tap(src: str | None = None, present=True, url: str | None = None):
...
+    """
+    item_set = set(items if isinstance(items, list) else [items])
+    if any(len(item) < 1 for item in item_set):
+        raise OperationValueError("all items must have non-zero length names")
+    # TODO: remove this once the test infrastructure supports enums
+    if isinstance(kind, str):
+        try:
+            kind = BrewItemKind(kind)
+        except (TypeError, ValueError):
+            raise OperationValueError from None
```
wowi42: `raise OperationValueError from None` has no message argument. When `kind` is an invalid string, the user sees a bare `OperationValueError` with no clue what went wrong. Add a descriptive message.

src/pyinfra/facts/brew.py
```diff
@@ -127,5 +153,55 @@ def requires_command(self) -> str:
...
+                "formulae": [],
+                "casks": [],
+                "commands": []
+            }
+        }
+    """
+
+    @override
+    def command(self) -> str:
+        return "brew trust --json v1"
```
wowi42: The documented flag is `--json=v1` (Homebrew 6.0.0 release notes, man page). While Homebrew may accept the space form, the canonical `=` form is safer and matches upstream docs.

wowi42 (changes requested):
## Issues

- **trust() docstring example uses wrong parameter names.** The example passes `item=` (singular) but the parameter is `items=` (plural), and `trust=True` but the parameter is `trusted`. Copy-pasting the example would fail with TypeError. The `+ item:` docstring line also needs to become `+ items:`.

- **raise OperationValueError from None has no message.** When kind is an invalid string, the error is raised with no description. The user sees a bare OperationValueError with no clue what went wrong.

- **brew trust --json v1 uses space instead of =.** The documented flag is --json=v1 (Homebrew 6.0.0 release notes). The canonical = form is safer and matches the upstream docs.

## Notes

- trusted parameter docstring says "Default False" but the Python default is None. The `trusted = trusted or False` line makes it behave as documented, but the docstring and signature are inconsistent.

src/pyinfra/operations/brew.py
morrison12: fixed.

src/pyinfra/operations/brew.py
morrison12: I think the best course of action for this one is to wait until #1841 is merged and then I can update the tests and remove this per the TODO above it.

morrison12:
> * trusted parameter docstring says "Default False" but the Python default is None. The `trusted = trusted or False` line makes it behave as documented, but the docstring and signature are inconsistent.

Also fixed.. The line in the body of the function is still required for the case where the parameter is not provided (needed for backwards compatibility).

src/pyinfra/operations/brew.py
morrison12: code now removed.

Issue #1837, opened 2026-06-16
infra: tests infrastructure doesn't convert scalar value to Enum when needed


morrison12:
## Describe the bug
When the parameter to an operation or fact is an Enum, the test infrastructure doesn't not detect that and convert the value (e.g a string or number) to an instance of the Enum.  This can be worked around by the fact or operation but it would be safe to centralize the instantiation code in one place.

## To Reproduce
Take `brew.trust` as an example. The `kind` parameter has a type of `BrewTrustKind` which can't be represented in the fixture.  There is currently code in the operation that detects Enum arguments and instantiates an enum value from the scalar value provided by the fixture.

## Expected behavior
The test infrastructure detects enum parameters and converts scalar values supplied by the test fixtures to an instance of the enum before invoking the method.

## Meta
+ Include output of `pyinfra --support`.
```text
    System: Darwin
      Platform: macOS-26.5.1-arm64-arm-64bit-Mach-O
      Release: 25.5.0
      Machine: arm64
    pyinfra: v3.7.post147.dev0+9d383298
      click: v8.3.0
      distro: v1.9.0
      gevent: v25.9.1
      jinja2: v3.1.6
      packaging: v25.0
      paramiko: v3.5.1
      pydantic: v2.12.2
      python-dateutil: v2.9.0.post0
      typeguard: v4.4.4
      types-paramiko: v3.5.0.20250801
      typing-extensions: v4.15.0
    Executable: /Users/morrison12/.virtualenvs/pyinfra-dev/bin/pyinfra
    Python: 3.14.0 (CPython, Clang 20.1.4 
```
+ How was pyinfra installed (source/pip)?
+ Include pyinfra-debug.log (if one was created)
+ Consider including output with `-vv` and `--debug`.

PR #1841, opened 2026-06-17
test: Enable tests from fixtures to work with string-valued Enum params

morrison12:
To solve #1837, detect string values of the form `enum:<enum-type-name>:<rest-of-string>` in parameters to operations (`._inner`) and facts (`command`) and create an enum instance from the value `<rest-of-string>`.  Invalid values cause exceptions at test time.

Updated all but one of the `selinux.boolean` tests to use the new approach to ensure the code gets run regularly.  The next candidates are the operations tests for `brew.trust` once it is properly sorted.


wowi42:
Thanks for this, the direction makes sense. A few points from a review:

1. The rewritten `.json` fixtures are no longer valid JSON. They have trailing commas before the closing brace. They only load because `testgen` runs `yaml.safe_load` on every fixture, including `.json` files. A strict JSON parser rejects them. Since the repo prefers YAML for new fixtures, converting these to `.yaml` (or keeping them strict JSON) would be safer.

2. Silent fallthrough can mask failures. `parse_value` only converts when `enum_map` is non-empty. If annotation resolution fails, `"enum:Boolean:on"` is passed to the operation as a literal string, and the test may still pass through the op's legacy string-compat branch. The `enum:` prefix should always raise when it cannot be resolved.

3. `get_enum_map` relies on `getattr(module, str(v.annotation))`, which only works because these modules use `from __future__ import annotations`, so annotations are plain strings. It silently misses `Boolean | None` style annotations and enums not exposed as module attributes. `typing.get_type_hints` would resolve types properly, and it would also allow automatic conversion of scalar values without a special fixture syntax, which is closer to what #1837 asked for.

4. Minor: the `# noqa: SLF001` comments sit on lines with no private access, and `from inspect import _empty` could use the public `inspect.Parameter.empty` instead.

morrison12:
> Thanks for this, the direction makes sense. A few points from a review:
> 
> 1. The rewritten `.json` fixtures are no longer valid JSON. They have trailing commas before the closing brace. They only load because `testgen` runs `yaml.safe_load` on every fixture, including `.json` files. A strict JSON parser rejects them. Since the repo prefers YAML for new fixtures, converting these to `.yaml` (or keeping them strict JSON) would be safer.

Ugh. Fixed.

> 2. Silent fallthrough can mask failures. `parse_value` only converts when `enum_map` is non-empty. If annotation resolution fails, `"enum:Boolean:on"` is passed to the operation as a literal string, and the test may still pass through the op's legacy string-compat branch. The `enum:` prefix should always raise when it cannot be resolved.

Fixed.

> 3. `get_enum_map` relies on `getattr(module, str(v.annotation))`, which only works because these modules use `from __future__ import annotations`, so annotations are plain strings. It silently misses `Boolean | None` style annotations and enums not exposed as module attributes. `typing.get_type_hints` would resolve types properly, and it would also allow automatic conversion of scalar values without a special fixture syntax, which is closer to what [infra: tests infrastructure doesn't convert scalar value to Enum when needed #1837](https://github.com/pyinfra-dev/pyinfra/issues/1837) asked for.

Good generalization. Change made.  
I left removing the special fixture syntax and supporting `Enum` (and `datetime` and `path` and ...) as/in return values for another PR given the JSON dumps and round trips.

> 4. Minor: the `# noqa: SLF001` comments sit on lines with no private access, and `from inspect import _empty` could use the public `inspect.Parameter.empty` instead.

wowi42 (approved):
One minor copy-paste issue: in `test_set_of_simple_enum_works`, the parameter type is `list[TestGetEnumMap.ZipEnum]` instead of `set[...]`. Doesn't affect correctness of `get_enum_map` but the test name and annotation don't match.

PR #1843, opened 2026-06-18
fix(operations.git): update origin remote when src changes (#1763)

### LINKED ISSUE #1763: git.repo does not honor change of src

vjsai:
Closes #1763

## Description

`git.repo` handled two cases for an existing working copy — switching branch and pulling — but never reconciled the configured `origin` remote with the `src` argument. As a result, changing `src` for an already-cloned `dest` was silently ignored: `origin` kept pointing at the old URL, so commits from the new source were never fetched.

### Fix

In the existing-repo branch, read the current `remote.origin.url` via the `GitConfig` fact and compare it to `src`:

- If an `origin` URL is configured and differs from `src`, emit `git remote set-url origin <src>` before the fetch/pull so subsequent git commands operate against the new source.
- The pull-skip optimisation (added to keep the op idempotent when local tip == remote tip) is bypassed on a remote change, because that remote-tip fact was gathered against the *old* origin and would otherwise wrongly report the repo as up to date.
- If no `origin` is configured the operation leaves remote management alone (no behaviour change), and when `origin` already equals `src` nothing is emitted, so the operation stays idempotent.

### Tests

- New fixture `tests/operations/git.repo/change_src.json` asserts that a changed `src` produces `git remote set-url origin <src>` followed by `git pull` — even when the local commit already matches the (old) remote tip, proving the pull is not skipped.
- Existing existing-repo fixtures gain a `git.GitConfig` fact whose `remote.origin.url` equals `src`, exercising the unchanged-remote (idempotent) path and confirming no extra command is emitted.

All operation tests pass (`1023 passed`), and `ruff check`, `ruff format`, `mypy`, and the argument-sync lint are clean.

## Checklist

- [x] Pull request includes documentation for any new/updated operations/facts (behaviour-only fix; the existing `src` docstring already describes the intended semantics)

## AI usage disclosure

Per the project's [AI policy](../blob/3.x/AI_POLICY.md): this change was written with AI assistance (Claude Code, `claude-opus-4-8`). The diff is small and deliberate, and I have reviewed and understand the change, the `git.repo` control flow it sits in, and how it interacts with the `GitConfig` fact and the existing pull-skip logic.

### DIFF
--- a/src/pyinfra/operations/git.py
+++ b/src/pyinfra/operations/git.py
@@ -163,6 +163,16 @@ def repo(
 
     # Ensuring existing repo
     else:
+        # Reconcile the `origin` remote URL with `src`. If `src` has changed for
+        # an existing working copy, update `origin` so the fetch/pull below
+        # operate against the new source instead of silently continuing to track
+        # the old remote (see GH #1763). Only act when an `origin` URL already
+        # exists and differs - if it is missing we leave remote management alone.
+        existing_remote = host.get_fact(GitConfig, repo=dest).get("remote.origin.url")
+        remote_changed = existing_remote is not None and existing_remote != [src]
+        if remote_changed:
+            git_commands.append(StringCommand("remote", "set-url", "origin", QuoteString(src)))
+
         is_tag = False
         current_branch = host.get_fact(GitBranch, repo=dest)
         if branch is not None and current_branch != branch:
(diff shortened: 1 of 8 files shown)


Fizzadar (commented):
LGTM! Thank you @vjsai

Fizzadar (approved):

PR #1846, opened 2026-06-20
fix(operations.selinux): harden commands with QuoteString and remove asserts

### LINKED ISSUE #1845: [Security] SELinux Operations Allow Command Injection and Use Assertions for Runtime Validation
## Description

The SELinux operations module (`src/pyinfra/operations/selinux.py`) currently exposes two issues:

### 1. Command Injection Risk

Several SELinux operations construct shell commands using user-controlled values passed directly into `StringCommand` without wrapping them in `QuoteString`.

Affected operations and variables:

| Operation | Variables |
|------------|------------|
| `port` | `se_type`, `protocol`, `port_num` |
| `boolean` | `value_str` |
| `file_context` | `se_type` |
| `file_context_mapping` | `se_type` |

Because these values are not properly quoted, shell metacharacters may be interpreted by the shell, potentially allowing command injection.

### Example

```python
selinux.port(
    name="Add port",
    ports="80; rm -rf /",
    protocol="tcp",
    se_type="http_port_t",
)
```

Generated command:

```bash
semanage port -a -t http_port_t -p tcp 80; rm -rf /
```

The shell interprets `;` as a command separator, causing unintended commands to execute.

### Expected Behavior

All user-controlled values passed to `StringCommand` should be wrapped using `QuoteString` so they are treated as literal arguments.

Expected output:

```bash
semanage port -a -t 'http_port_t' -p 'tcp' '80; rm -rf /'
```

---

### 2. Runtime Validation Uses `assert`

The module also uses `assert` statements for type validation.

Example:

```python
assert isinstance(port_num, str)
```

Assertions can be removed when Python is executed with optimization enabled (`python -O`), resulting in validation being skipped entirely.

### Expected Behavior

Replace assertions used for runtime validation with explicit exceptions:

```python
if not isinstance(port_num, str):
    raise TypeError(
        f"Expected str, got {type(port_num).__name__}"
    )
```

This ensures validation always occurs regardless of Python optimization settings.

---

### Security

Potential command injection through SELinux operations when untrusted values are provided.

### Reliability

Input validation may be silently disabled when Python is executed with optimization flags.

---

## Suggested Fix

1. Wrap all user-controlled values passed into `StringCommand` with `QuoteString`.
2. Replace all runtime validation assertions with explicit exception handling (`TypeError` or `ValueError` as appropriate).
3. Add regression tests verifying malicious shell payloads are properly quoted and cannot alter generated commands.

## Affected File

```text
src/pyinfra/operations/selinux.py
```

## Additional Notes

This issue affects the security posture of SELinux-related operations because generated commands may execute unintended shell instructions when supplied with malicious input. Proper quoting and explicit runtime validation would align the implementation with secure command construction practices and project coding standards.

Muhammad-Ikhwan-Fathulloh:
## Summary

This PR addresses security and code quality issues in the SELinux operations module.

Several SELinux operations were constructing shell commands using user-controlled values without wrapping them in `QuoteString`, potentially allowing shell command injection when malicious input contains shell metacharacters.

Additionally, runtime input validation relied on `assert` statements, which can be stripped when Python is executed with optimization enabled (`python -O`). These assertions have been replaced with explicit type checks and exceptions.

## Linked Issue

Fixes #1845

### Security Hardening

Wrapped user-controlled values passed to `StringCommand` with `QuoteString` in the following operations:

* `selinux.port`

  * `se_type`
  * `protocol`
  * `port_num`

* `selinux.boolean`

  * `value_str`

* `selinux.file_context`

  * `se_type`

* `selinux.file_context_mapping`

  * `se_type`

This ensures shell metacharacters are treated as literal arguments and cannot alter command execution.

### Runtime Validation Improvements

Replaced runtime assertions such as:

```python
assert isinstance(value, str)
```

with explicit validation:

```python
if not isinstance(value, str):
    raise TypeError(...)
```

This guarantees validation remains active regardless of Python optimization settings.

### Tests

Added a regression test covering malicious SELinux input values to verify:

* Shell metacharacters are properly quoted.
* Generated commands cannot be altered through injected shell syntax.
* Existing SELinux functionality remains unchanged.

## Example

Before:

```bash
semanage port -a -t http_port_t -p tcp 80; rm -rf /
```

After:

```bash
semanage port -a -t 'http_port_t' -p 'tcp' '80; rm -rf /'
```

## Testing

Executed SELinux operation tests:

```bash
uv run pytest tests/test_operations.py -k "selinux" -v
```

Results:

* All existing SELinux tests passed.
* New security regression test passed.

## Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)
* [x] Security fix
* [x] Refactoring (internal code improvement)
* [ ] New feature
* [ ] Breaking change

## Checklist

* [x] Linked to an existing issue
* [x] Added/updated tests
* [x] Existing tests pass
* [x] No breaking API changes introduced
* [x] Follows project coding standards

### DIFF
--- a/src/pyinfra/operations/selinux.py
+++ b/src/pyinfra/operations/selinux.py
@@ -50,7 +50,8 @@ def boolean(bool_name: str, value: Boolean, persistent=False):
 
     value_str: str
     if value in ["on", "off"]:  # compatibility with the old version
-        assert isinstance(value, str)
+        if not isinstance(value, str):
+            raise TypeError("value must be a string")
         value_str = value
     elif value is Boolean.ON:
         value_str = "on"
@@ -64,7 +65,7 @@ def boolean(bool_name: str, value: Boolean, persistent=False):
         if persistent:
             command_bits.append("-P")
         command_bits.append(QuoteString(bool_name))
-        command_bits.append(value_str)
+        command_bits.append(QuoteString(value_str))
         yield StringCommand(*command_bits)
     else:
         host.noop(f"boolean '{bool_name}' already had the value '{value_str}'")
@@ -91,7 +92,7 @@ def file_context(path: str, se_type: str):
 
     current = host.get_fact(FileContext, path=path) or {}
     if se_type != current.get("type", ""):
-        yield StringCommand("chcon", "-t", se_type, QuoteString(path))
+        yield StringCommand("chcon", "-t", QuoteString(se_type), QuoteString(path))
     else:
         host.noop(f"file_context: '{path}' already had type '{se_type}'")
 
@@ -126,7 +127,9 @@ def file_context_mapping(target: str, se_type: str | None = None, present=True):
     if present:
         option = "-a" if len(current) == 0 else ("-m" if current.get("type") != se_type else "")
         if option != "":
-            yield StringCommand("semanage", "fcontext", option, "-t", se_type, QuoteString(target))
+            yield StringCommand(
+                "semanage", "fcontext", option, "-t", QuoteString(se_type), QuoteString(target)
+            )
         else:
             host.noop(f"mapping for '{target}' -> '{se_type}' already present")
     else:
@@ -161,7 +164,8 @@ def port(protocol: Protocol | str, port_num: int, se_type: str | None = None, pr
     """
 
     if protocol is Protocol:
-        assert isinstance(protocol, Protocol)
+        if not isinstance(protocol, Protocol):
+            raise TypeError("protocol must be a Protocol instance")
         protocol = protocol.value
 
     if present and (se_type is None):
@@ -178,12 +182,23 @@ def port(protocol: Protocol | str, port_num: int, se_type: str | None = None, pr
     if present:
         option = "-a" if current == "" else ("-m" if current != se_type else "")
         if option != "":
-            yield StringCommand("semanage", "port", option, "-t", se_type, "-p", protocol, port_num)
+            yield StringCommand(
+                "semanage",
+                "port",
+                option,
+                "-t",
+                QuoteString(se_type),
+                "-p",
+                QuoteString(protocol),
+                QuoteString(port_num),
+            )
         else:
             host.noop(f"setype for '{protocol}/{port_num}' is already '{se_type}'")
     else:
         if current != "":
-            yield StringCommand("semanage", "port", "-d", "-p", protocol, port_num)
+            yield StringCommand(
+                "semanage", "port", "-d", "-p", QuoteString(protocol), QuoteString(port_num)
+            )
         else:
             host.noop(f"setype for '{protocol}/{port_num}' is already unset")
 
--- a/src/pyinfra/operations/selinux.py
+++ b/src/pyinfra/operations/selinux.py
@@ -65,7 +65,7 @@ def boolean(bool_name: str, value: Boolean, persistent=False):
         if persistent:
             command_bits.append("-P")
         command_bits.append(QuoteString(bool_name))
-        command_bits.append(QuoteString(value_str))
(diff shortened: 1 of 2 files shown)


wowi42:
Thanks for tackling the SELinux hardening. The `QuoteString` wrapping is correct and the new regression fixture matches `shlex.quote` output. Two issues worth addressing before merge:

**1. `selinux.py:180` and `selinux.py:128` — use `is not None` instead of a truthy check.**

Both `port` and `file_context_mapping` changed `if option != "":` to `if option != "" and se_type:`. Since `se_type` is typed `str | None`, this should be `se_type is not None`. Per the repo convention: "when a parameter is `T | None`, branch on `if x is not None:` rather than truthy `if x:`. Empty strings, `0`, and empty containers are valid user input and a truthy check silently drops them." With `se_type=""` (which passes the earlier `is None` guard), the command that should run is silently routed to the `host.noop(...)` "already present" branch and nothing happens, with no error.

**2. `selinux.py:163` — `if protocol is Protocol:` is always False.**

This compares an enum member against the enum class, which is never true, so the branch never runs and the newly added `raise TypeError(...)` is unreachable. As a side effect, passing `Protocol.TCP` (as the docstring example shows) skips the `protocol = protocol.value` conversion, and `QuoteString(str(protocol))` then renders `-p 'Protocol.TCP'` instead of `-p tcp`. This is pre-existing, but since these lines are being touched it's a good time to fix it to `isinstance(protocol, Protocol)`. The new test only exercises string protocols, so it doesn't catch this.

**Minor:** wrapping `value_str` in `QuoteString` in `boolean` is harmless but unnecessary, since at that point `value_str` is always the literal `"on"` or `"off"`.

The rest looks good.

Muhammad-Ikhwan-Fathulloh:
> Thanks for tackling the SELinux hardening. The `QuoteString` wrapping is correct and the new regression fixture matches `shlex.quote` output. Two issues worth addressing before merge:
> 
> **1. `selinux.py:180` and `selinux.py:128` — use `is not None` instead of a truthy check.**
> 
> Both `port` and `file_context_mapping` changed `if option != "":` to `if option != "" and se_type:`. Since `se_type` is typed `str | None`, this should be `se_type is not None`. Per the repo convention: "when a parameter is `T | None`, branch on `if x is not None:` rather than truthy `if x:`. Empty strings, `0`, and empty containers are valid user input and a truthy check silently drops them." With `se_type=""` (which passes the earlier `is None` guard), the command that should run is silently routed to the `host.noop(...)` "already present" branch and nothing happens, with no error.
> 
> **2. `selinux.py:163` — `if protocol is Protocol:` is always False.**
> 
> This compares an enum member against the enum class, which is never true, so the branch never runs and the newly added `raise TypeError(...)` is unreachable. As a side effect, passing `Protocol.TCP` (as the docstring example shows) skips the `protocol = protocol.value` conversion, and `QuoteString(str(protocol))` then renders `-p 'Protocol.TCP'` instead of `-p tcp`. This is pre-existing, but since these lines are being touched it's a good time to fix it to `isinstance(protocol, Protocol)`. The new test only exercises string protocols, so it doesn't catch this.
> 
> **Minor:** wrapping `value_str` in `QuoteString` in `boolean` is harmless but unnecessary, since at that point `value_str` is always the literal `"on"` or `"off"`.
> 
> The rest looks good.

Thanks for the review! I’ve addressed the points you raised:

1. Updated se_type checks to is not None to align with project conventions.
2. Changed the Protocol check to isinstance(protocol, Protocol) to fix the enum handling.
3. Removed the redundant QuoteString wrapper for the boolean value.

I have pushed these changes to the branch. Thanks again for the guidance!

PR #1606, opened 2026-03-15
docs: generate fact/operation docs for modules (vs files)

morrison12:
Generate online docs for  facts and/or operations when they are implemented in a module (vs. in a file).
The immediate impact is to reveal ``freebsd`` in the sidebar of the [Operations Index](https://docs.pyinfra.com/en/3.x/operations.html) but can be used more generally.

I did not add `freebsd-ops` to `pyinfra-metadata.toml` as I am assuming it was not present for a reason but can do so if the consistency is helpful.

Related to this, I would have expected ``__init__.py`` for the ``freebsd`` operations to be as shown at /1/ (vs. the current ``from . import *``) to minimize duplication in the import lines but did _not_ change it in this PR as it would be a breaking change.

- [x] Pull request is based on the default branch (`3.x` at this time)
- [n/a] Pull request includes tests for any new/updated operations/facts
- [n/a] Pull request includes documentation for any new/updated operations/facts
- [x] Tests pass (see `scripts/dev-test.sh`)
- [x] Type checking & code style passes (see `scripts/dev-lint.sh`) ***ruff format and ruff check run manually***

/1/
```python
from . import pkg
from .freebsd_update import update
from .service import service
from .sysrc import sysrc
```

PR #1847, opened 2026-06-21
feat(facts, operations): add OpenWrt packages, move opkg underneath openwrt and add packages and update operations

morrison12:
Add OpenWrt facts and operations packages, move the existing `opkg` items there and add `packages` and `update` operations that select `apk` vs. `opkg` based on the running OpenWrt release.  

The rationale for the OpenWrt _package_ is that there are three OpenWrt -specific items (`opkg`, `procd`, `uci`) needed which don't need to be at the top-level.  `opkg` exists today (even if it is being supplanted by `apk`),  `procd` support will arrive soon and there are a couple of people working on `uci` support.

The expectation is that the top-level `opkg` will be removed after a suitable interval; the interval can probably be reasonably be short as the OpenWrt change from `opkg` to `ask` in their Release 25.12 means most users will switch to the release/package-management independent `openwrt.packages` once it is available.

Specifically changes here are:
1. add OpenWrt facts and operations packages:
 * add the directories themselves
 * adopt and then modify the deferred import mechanism used for `operations/freebsd` to allow for the equivalent of `import .sub-package.something as something` which is useful to avoiding unnecessarily levels in the namespace.  
 * modify the `operations/freebsd` implementation to be consistent with the `openwrt` implementations
 * change `test_facts.py` to allow for packages (vs. files) by making the same change as was made for `test_operations.py`
2. move the existing `opkg` facts and operations to `openwrt`
* move the implementation files to the relevant directory
* rename the test folders
3. replace the now-moved `opkg` facts and operations with deprecated "forwarders"
* provides backwards compatibility 
* add support for deprecated facts (both long and short) using same approach as taken for operations
* add limit of one warning _per `pyinfra` run per deprecated_ fact as some facts are called a large number of times
* add a **warning** to the top of the module-level doc string about the deprecation of the top-level `opkg` for both facts and operations
* copy the relevant tests from `openwrt.opkg`
5. create an `OpenWrtHasFeature` fact, a general version of check needed for 6
* currently decides based on OpenWrt release but does not need to be restricted to that
* a few lines of code can be removed when `test_facts.py` and `pyinfra_cli` support enum parameters
6. add release and package-manager independent `packages` and `update` operations for OpenWrt
* use`OpenWrtHasFeature` to decide whether to use `apk` or `opkg`
* no corresponding `Packages` fact: is it is only consulted prior to doing something which is already covered for the `packages` and `update` cases
7. add facts and operations entries in `pyinfra-metadata.toml` for (1)
8. fixed documentation generation to work for facts and operations in packages and not just files
* #1609  broke #1606  (for `operations/freebsd`)
* consolidated (and parametrized) code that generates lists of facts and operations
* depends on the exact deferred import mechanism implemented at (1) and will need to be reworked when when that is replaced by `lazy import` mechanism in `python 3.15`
9. changed sorting order of operations in `scripts/generation_operations_docs.py` to match sorting order used for imports; this affects not just the order on the page but also the order in the sidebar 

A couple of ***questions*** related to the deprecation at (3) above:
a)  there is a warning in the documentation for the deprecations but is there a specific `admonition` for deprecation that should have been used ?
b) does the future release need to be specified now (the documentation currently doesn't specify one)

One ***note***:  at (5), i.e. the `OpenWrtHasFeature` fact, there is a core mechanism (`ReleaseRange.contains`) for which it wasn't clear where to put the (non-fixture) tests. I opted for `tests/test_facts_other.py` but am happy to relocate them if there is a preferred location for such tests.


wowi42 (changes requested):
## Verdict: NO-GO

Well-structured PR with good deprecation handling and broad test coverage. Three issues need fixing before merge.

## Issues

### 1. Fixture bug: `has_dsa_is_false_for_19_07.json` tests the wrong feature

`tests/facts/openwrt.features.OpenWrtHasFeature/has_dsa_is_false_for_19_07.json` has `"arg": ["has_dsa"]` and `"command": "echo has_dsa && ..."`, but the first line of `output` is `"uses_apk"` instead of `"has_dsa"`. The test passes accidentally because both features return `false` for release 19.07, but the fixture is not testing what its name and args claim. First output line must match the echoed feature name: `"has_dsa"`.

### 2. Missing fixture: `openwrt.update` (i.e. `openwrt.packages.update`) has no tests

The `update` function in `src/pyinfra/operations/openwrt/packages.py` is a `@operation` that calls `host.get_fact(OpenWrtHasFeature, ...)`. There is no `tests/operations/openwrt.packages.update/` directory. Create at least two fixtures covering the apk and opkg branches.

### 3. Drive-by refactor of `freebsd/__init__.py` — out of scope

`src/pyinfra/operations/freebsd/__init__.py` replaces the glob-based module auto-discovery with an explicit `__ALL__` dict and a `getattr` that splits on dots for sub-module attribute access. This is unrelated to OpenWrt. The glob approach was more maintainable (drop a new `.py` under `freebsd/` and it's auto-discovered); the explicit dict requires manual registration. If this change is intentional, split it into its own PR with a rationale.

## Notes

- Copy-paste comment errors in `__getattr__` docstrings: `src/pyinfra/operations/freebsd/__init__.py` line 17, `src/pyinfra/operations/openwrt/__init__.py` line 19, and `src/pyinfra/facts/openwrt/__init__.py` line 19 all say "On-demand import of OpenWrt facts" but should reference the correct package (FreeBSD operations, OpenWrt operations, OpenWrt facts respectively).
- `openwrt.packages` has `packages: str | list[str] = ""` default. Per AGENTS.md new operations should use `None`. The `""` default passes through to `apk.packages._inner()` which treats `""` as a package name rather than nooping. Minor since the operation is always called with explicit package lists, but worth aligning with `None` convention.

morrison12:
> ## Verdict: NO-GO
> Well-structured PR with good deprecation handling and broad test coverage. Three issues need fixing before merge.
> 
> ## Issues
> ### 1. Fixture bug: `has_dsa_is_false_for_19_07.json` tests the wrong feature
> `tests/facts/openwrt.features.OpenWrtHasFeature/has_dsa_is_false_for_19_07.json` has `"arg": ["has_dsa"]` and `"command": "echo has_dsa && ..."`, but the first line of `output` is `"uses_apk"` instead of `"has_dsa"`. The test passes accidentally because both features return `false` for release 19.07, but the fixture is not testing what its name and args claim. First output line must match the echoed feature name: `"has_dsa"`.

Fixed.
> 
> ### 2. Missing fixture: `openwrt.update` (i.e. `openwrt.packages.update`) has no tests
> The `update` function in `src/pyinfra/operations/openwrt/packages.py` is a `@operation` that calls `host.get_fact(OpenWrtHasFeature, ...)`. There is no `tests/operations/openwrt.packages.update/` directory. Create at least two fixtures covering the apk and opkg branches.

Indeed needed.  Added.
> 
> ### 3. Drive-by refactor of `freebsd/__init__.py` — out of scope
> `src/pyinfra/operations/freebsd/__init__.py` replaces the glob-based module auto-discovery with an explicit `__ALL__` dict and a `getattr` that splits on dots for sub-module attribute access. This is unrelated to OpenWrt. The glob approach was more maintainable (drop a new `.py` under `freebsd/` and it's auto-discovered); the explicit dict requires manual registration. If this change is intentional, split it into its own PR with a rationale.

Removed.
I had changed `freebsd/__init__.py` in hopes of avoiding future duplicate package name elements  (e.g.  the current `freebsd.sysrc.sysrc`); it does take some effort to manually add packages but IMHO that was a nudge to think about the API.
Will put a separate PR for that on my list.

> 
> ## Notes
> * Copy-paste comment errors in `__getattr__` docstrings: `src/pyinfra/operations/freebsd/__init__.py` line 17, `src/pyinfra/operations/openwrt/__init__.py` line 19, and `src/pyinfra/facts/openwrt/__init__.py` line 19 all say "On-demand import of OpenWrt facts" but should reference the correct package (FreeBSD operations, OpenWrt operations, OpenWrt facts respectively).

Fixed.

> * `openwrt.packages` has `packages: str | list[str] = ""` default. Per AGENTS.md new operations should use `None`. The `""` default passes through to `apk.packages._inner()` which treats `""` as a package name rather than nooping. Minor since the operation is always called with explicit package lists, but worth aligning with `None` convention.

Fixed both `openwrt.packages` and `opkg.packages` (no backwards compatibility issue).
Not sure how many humans read `AGENTS.md`; we should look at adding some of what is desired (such changes, new fixtures in YAML, etc to CONTRIBUTING.md and/or the documents.

Also I converted the fixtures to YAML while I was here and added typing to `openwrt.opkg` (which predated typed pyinfra).

morrison12:
@wowi42: are further changes required ?

wowi42 (changes requested):
First, sorry for the long delay to come back on this. You answered the July review quickly and since then it has waited far too long, that is on me. Sorry also that the points below are new and were not seen in the first pass, they only came out with a fresh read and some experimentation this time.

The good news: all the points from the previous review are resolved. The `has_dsa` fixture is correct, `openwrt.update` has fixtures for both branches, the freebsd refactor is gone, and the two notes are fixed. Tests pass locally (1986) and CI is green. I have three new issues before merge, two are trivial and one is more about design.

## Issues

### 1. An accidentally committed file

`repros/brew-fixes-2026-02-repro.sh` looks like a local repro script about brew/openssl, not related to this PR. Could you remove it?

### 2. The `opkg.update` forwarder never shows the deprecation warning

In `src/pyinfra/operations/opkg.py`, `update()` has `@operation(is_idempotent=False, deprecated_for="openwrt.opkg.update or openwrt.update")` but, contrary to `packages()`, it misses `is_deprecated=True`. The warning only fires when `is_deprecated` is truthy (see `api/operation.py`), so `opkg.update` silently skips the deprecation that this PR is supposed to give.

### 3. The deferred-import `__getattr__` re-executes modules on every access

In both `operations/openwrt/__init__.py` and `facts/openwrt/__init__.py`, the `del sys.modules[module.__name__]` means every lookup of `openwrt.packages` or `OpenWrtFeature` re-imports and re-executes the submodule. I verified this on the branch:

- `openwrt.packages is openwrt.packages` gives `False` (a new function object at each access, the `@operation` decorators run again every time)
- `OpenWrtFeature is OpenWrtFeature` gives `False`, so enum identity is unstable and `==` comparisons between two accesses fail. The operation path is saved only by chance: `OpenWrtHasFeature.command/process` round-trips the feature through the echoed string and re-derives the enum locally. Any other enum comparison (future code, user deploys) will break in a subtle way.
- `del sys.modules[__name__].__dict__[pieces[0]]` can also raise `KeyError` if the attribute is absent.

Could this be simpler? Either cache the resolved attribute in `globals()` instead of deleting from `sys.modules`, or use classic eager imports like `from .packages import packages, update` in `__init__.py`. I do not see an import cycle that prevents this. One attention point: the docs-generation scripts now depend on this exact mechanism, so they must keep working with whatever replaces it.

## Notes (not blocking)

- Some docstring glitches: mismatched backticks and a stray `TBD -` in `openwrt/packages.py`, `.. note:` with a single colon, the meta-comment "note: this does _not_ show up in the online documentation..." left in three module docstrings, and the `+ packages:` description which reads "to that must/must not be present (default ``True``)" (grammar, and the default is `None`).
- `any((len(p) < 1) or (p is None) for p in ...)` in both `packages()` operations: `len(None)` raises `TypeError` for a `None` element before the `p is None` check has a chance to run. This path is not tested (fixtures only cover all-`""` lists).
- `already_logged_as_deprecated` in `api/facts.py` is a module-level mutable global with a public-looking name. I propose a `_` prefix, and the `global` statement is not necessary (it is only mutated).
- The deprecated fact wrappers (`facts/opkg.py`) lost all their fixture coverage (old `tests/facts/opkg.*` deleted), and the new warning emission in `get_fact` has no test.
- `logs: foo` in the null-list fixture is junk, but it is copied as-is from the old fixture, so pre-existing. Feel free to ignore.

Thanks again for the patience, and sorry once more for the delay.

PR #1851, opened 2026-06-22
fix(facts.hardware): use JSON from ip address show in NetworkDevices to cope with lack of prefix length and/or broadcast

### LINKED ISSUE #1848: IPv4 address discovery failure  (pointopoint-related?)
pyinfra v3.9.2 is unable to discover one of the IPv4 addresses  (actually the public one) of a virtual machine at Hetzner.

The target virtual machine, running Debian 13, has one Ethernet interface, `enp1s0`, set up with a public IPv4 address (redacted here as 88.198.X.Y) having as 'peer'/gateway a private address, 172.31.1.1 (unlike 'vanilla' cases in which the host and gateway IPv4 addresses are in a single CIDR block).  I mentioned Hetzner above because this is a common situation for virtual machines there (a 'legit' way to handle addresses/routing, not an anomaly).

Tried (remotely) both of these:
 
`pyinfra HOSTNAME fact hardware.Ipv4Addrs`
With no discovered IPv4 address, that network interface (`enp1s0`) is not included in the output.

`pyinfra HOSTNAME fact hardware.NetworkDevices`
The interface is listed in the output, but only IPv6 addresses are shown.

The machine also has 2 active WireGuard interfaces, and both of the above nicely report their interface names and addresses.  

I'm using ifupdown-ng to set up the interfaces there, and (with addresses redacted) this is the current setup for `enp1s0`:
```
auto  enp1s0
iface enp1s0
  address 88.198.X.Y/32
  point-to-point 172.31.1.1
  gateway 172.31.1.1
  address [IPv6 ADDRESS REDACTED]
  gateway fe80::1
  forward-ipv4 yes
  forward-ipv6 yes

```
and this is the output from `ip addr show enp1s0` there:
```
2: enp1s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 96:00:03:bf:0e:9f brd ff:ff:ff:ff:ff:ff
    altname enx960003bf0e9f
    altname HETZ
    inet 88.198.X.Y peer 172.31.1.1/32 scope global enp1s0
       valid_lft forever preferred_lft forever
    inet6 [IPv6 ADDRESS REDACTED] scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::9400:3ff:febf:e9f/64 scope link proto kernel_ll 
       valid_lft forever preferred_lft forever

```

The expected behavior would of course be for pyinfra to discover that 88.198.X.Y address of `enp1s0`.
Maybe the specific output format for that case — an interface address with no "/…" prefix length part, followed by "peer  172.31.1.1/32" — is not expected and pyinfra fails to parse it?

Information provided by `pyinfra --support` at my machine running pyinfra:
```
    System: Linux
      Platform: Linux-6.12.90+deb12.1-amd64-x86_64-with-glibc2.36
      Release: 6.12.90+deb12.1-amd64
      Machine: x86_64
    pyinfra: v3.9.2
      click: v8.4.1
      distro: v1.9.0
      gevent: v26.5.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.15.0
    Executable: /home/jmce/.local/bin/pyinfra
    Python: 3.11.2 (CPython, GCC 12.2.0)

```

Again with some redacting, here is the output of  `pyinfra -vv --debug HOSTNAME fact hardware.Ipv4Addrs`:

```
--> Loading config...
--> Loading inventory...
    [pyinfra_cli.inventory] Creating fake inventory...
    [pyinfra_cli.inventory] Checking possible group_data at: /home/jmce/group_data

--> Connecting to hosts...
    [pyinfra.connectors.ssh] Connecting to: HOSTNAME ({'allow_agent': True, 'look_for_keys': True, '_pyinfra_ssh_forward_agent': False, '_pyinfra_ssh_config_file': None, '_pyinfra_ssh_known_hosts_file': None, '_pyinfra_ssh_strict_host_key_checking': 'accept-new', '_pyinfra_ssh_paramiko_connect_kwargs': None, 'timeout': 10})
    [pyinfra.connectors.sshuserclient.client] Loading SSH config: None
    [HOSTNAME] Connected
    [pyinfra.api.state] Activating host: HOSTNAME

--> Gathering facts...
    [pyinfra.api.facts] Getting fact: hardware.NetworkDevices () (ensure_hosts: None)
    [pyinfra.connectors.ssh] Running command on HOSTNAME: (pty=False) sh -c 'ip addr show 2> /dev/null || ifconfig -a'
[HOSTNAME] >>> sh -c 'ip addr show 2> /dev/null || ifconfig -a'
    [pyinfra.connectors.ssh] Waiting for exit status...
    [pyinfra.connectors.ssh] Command exit status: 0
    [HOSTNAME] Loaded fact hardware.NetworkDevices

--> Fact data for: hardware.Ipv4Addrs
{
    "HOSTNAME": {
        "lo": [
            "127.0.0.1"
        ],
        "wgWAN0": [
            "100.64.0.1"
        ],
        "wgWAN1": [
            "100.65.0.1"
        ]
    }
}

```

morrison12:
To avoid (more) regex challenges, added support for `ip -j show addr` and extracted the required information from the resulting JSON output.  This should fix #1848 and hopefully head off problems due to other variations in output.

A few notes:
1) The existing `ip addr show` command and related regexes were left as an alternative in case a host does not support JSON output from `ip addr show` despite it being far [far from new](https://lwn.net/Articles/738897/); that might have been overly cautious.

2) If there is no `prefixlen` reported by the host, as is the case in #1848, the new code does _*not*_ invent one

3) There is now a `TypedDict` for the return type of `hardware.NetworkDevices`


wowi42:
Reviewed this. Switching `NetworkDevices` to `ip -j addr show` and parsing JSON is the right call, and the `mask()` octet math checks out (verified /16, /24, /25, /32, plus hex and dotted-quad inputs all produce correct netmasks). The happy path is well covered by the new fixtures. The issues are concentrated in the error handling of `ntwk_info_from_json`, most severe first:

1. The `error` flag is never reset, so one bad interface silently drops every interface after it. `error` starts `False` and is only ever set to `True`, never back. The per-interface store is gated on `if not (error := error or ...)`, so once any interface or any `addr_info` trips an error, `error` stays `True` for the rest of the loop and every subsequent valid interface is omitted from `result` (with a single `logger.error`). A host where the 2nd of 8 interfaces is malformed loses interfaces 2 through 8. Error handling should be per-interface (skip the bad one, keep the rest), not a sticky global flag.

2. An IPv6 address with no `prefixlen` sets `error=True`, which (via point 1) drops that interface and all following ones. This is asymmetric with IPv4, which simply omits `mask_bits` when `prefixlen` is absent and continues. It also contradicts the stated goal of coping with missing prefix length (#1848). IPv6 without `prefixlen` should degrade like IPv4, not poison the result.

3. `device` and `addr_blk` are function-scoped and reused across iterations, so a missing field causes use-before-assignment or stale-data corruption. When an `addr_info` lacks `"local"`, `addr_blk` is left unassigned but the IPv4/IPv6 branches still run `addr_blk["broadcast"] = ...` / `addr_blk["mask_bits"] = ...`. On the first interface this raises `UnboundLocalError` and crashes the whole fact; on a later interface `addr_blk` still references the previous interface's block (already stored in `result` by reference), so the assignment mutates already-emitted data. Same pattern for `device` when the required fields are missing. Each block should be built fresh and only merged once validated.

4. Minor: `all_devices[device_name] = cast("DevInfoType", cast(object, device_info))` in the text path. The double-cast through `object` works but signals that the text-path dict does not actually conform to `DevInfoType`. Worth tightening the construction so a single cast (or none) suffices.

The malformed-input handling is exactly where this fact gets exercised in the wild (that is what #1848 is), so the per-interface loop is worth restructuring before merge so a single odd interface or address cannot drop unrelated interfaces or crash.

morrison12:
Thanks for the review.
Some notes below and a new version

> Reviewed this. Switching `NetworkDevices` to `ip -j addr show` and parsing JSON is the right call, and the `mask()` octet math checks out (verified /16, /24, /25, /32, plus hex and dotted-quad inputs all produce correct netmasks). The happy path is well covered by the new fixtures. The issues are concentrated in the error handling of `ntwk_info_from_json`, most severe first:
> 
> 1. The `error` flag is never reset, so one bad interface silently drops every interface after it. `error` starts `False` and is only ever set to `True`, never back. The per-interface store is gated on `if not (error := error or ...)`, so once any interface or any `addr_info` trips an error, `error` stays `True` for the rest of the loop and every subsequent valid interface is omitted from `result` (with a single `logger.error`). A host where the 2nd of 8 interfaces is malformed loses interfaces 2 through 8. Error handling should be per-interface (skip the bad one, keep the rest), not a sticky global flag.

This was deliberate (i.e. fail on error).  I've now changed it to salvage as much (down to the address block level) as makes sense. In the more general case there should probably be guidance in [Writing Facts](https://docs.pyinfra.com/en/3.x/api/facts.html#writing-facts) on a) _never_ fail silently and b) whether the `pyinfra` way to to salvage as much as is self-consistent or to give up on the first error.  This code now does the former.

> 2. An IPv6 address with no `prefixlen` sets `error=True`, which (via point 1) drops that interface and all following ones. This is asymmetric with IPv4, which simply omits `mask_bits` when `prefixlen` is absent and continues. It also contradicts the stated goal of coping with missing prefix length ([IPv4 address discovery failure  (pointopoint-related?) #1848](https://github.com/pyinfra-dev/pyinfra/issues/1848)). IPv6 without `prefixlen` should degrade like IPv4, not poison the result.

IPv4 and IPv6 are now both the same, i.e. the field is just omitted
> 3. `device` and `addr_blk` are function-scoped and reused across iterations, so a missing field causes use-before-assignment or stale-data corruption. When an `addr_info` lacks `"local"`, `addr_blk` is left unassigned but the IPv4/IPv6 branches still run `addr_blk["broadcast"] = ...` / `addr_blk["mask_bits"] = ...`. On the first interface this raises `UnboundLocalError` and crashes the whole fact; on a later interface `addr_blk` still references the previous interface's block (already stored in `result` by reference), so the assignment mutates already-emitted data. Same pattern for `device` when the required fields are missing. Each block should be built fresh and only merged once validated.
> 4. Minor: `all_devices[device_name] = cast("DevInfoType", cast(object, device_info))` in the text path. The double-cast through `object` works but signals that the text-path dict does not actually conform to `DevInfoType`. Worth tightening the construction so a single cast (or none) suffices.
The regex parsing code now uses the same data types.
> 
> The malformed-input handling is exactly where this fact gets exercised in the wild (that is what #1848 is), so the per-interface loop is worth restructuring before merge so a single odd interface or address cannot drop unrelated interfaces or crash.

PR #1853, opened 2026-06-24
feat(connectors): allow HiddenValue in _env

### LINKED ISSUE #1798: `_env` in shell and script operations should accept HiddenValue

I am trying to pass some secret values via env variables. But it is getting rejected as the operator expects string values. I want to mask the secrets from getting printed in the logs.

```py
env = {"TOKEN": HiddenValue(os.getenv("..."))}
server.shell(
        commands=[
            "docker compose pull",
            "docker compose up",
        ],
        _chdir="/opt",
        _env=env,
    )
```

I want to be able to pass env dict with `HiddenValue` values. (related PR: https://github.com/pyinfra-dev/pyinfra/pull/1709)

Jesse-11:
## Summary

This PR is to allow _env values to accept the new `HiddenValue` so secret env variables can be passed into shell commands and not be exposed in the command logs.

The issue was that _env values were treated as strings when building the `export KEY=value` command. This meant that HiddenValue values would be converted into their masked representation too early, this subsequently caused the masked value to be used in the actual command instead of only the displayed command

This PR preserves the `HiddenValue` objects while constructing the env variables. Hence the raw value can be used during execution while keeping the masked value in logs. 

## Linked Issue

Fixes #1798 

## Changes

- Updates _env typing to support `HiddenValue` values:
```
EnvValue = str | HiddenValue

def generate_env(config: Config, value: Mapping[str, EnvValue] | None) -> dict[str, EnvValue]:
```

- Allow env value to be None as operation typing currently allow `_env=None` for safer handling
- Updates the environmental command construction to preserve `HiddenValue` until the command is rendered.
- Keeps the full `KEY=value` assignment quoted as one shell token.
- Adds tests covering:
  - `HiddenValue` being used as an _env value.
  - Raw command rendering using real secret
  - Masked command rendering hiding the secret
  - Shell-special characters remain safetly quoted
 

## Testing

Executed connectors tests:

`uv run pytest tests/test_connectors/test_util.py -k "env"`

Ran a tmp deploy (as shown in issues #1798) with full success:

`uv run pyinfra @local ../deploy.py -vv -y`

## Checklist

### DIFF
--- a/src/pyinfra/api/arguments.py
+++ b/src/pyinfra/api/arguments.py
@@ -16,6 +16,7 @@ from typing_extensions import TypedDict
 from pyinfra.api.exceptions import ArgumentTypeError
 from pyinfra.api.util import raise_if_bad_type
 from pyinfra.context import ctx_config
+from pyinfra.api.hiddenvalue import HiddenValue
 
 if TYPE_CHECKING:
     from pyinfra.api import Config, Host, State
@@ -23,6 +24,7 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 default_sentinel = object()
 
+EnvValue = str | HiddenValue
 
 class ArgumentMeta(Generic[T]):
     description: str
@@ -62,7 +64,7 @@ class ConnectorArguments(TypedDict, total=False):
     # Shell arguments
     _shell_executable: str
     _chdir: str
-    _env: Mapping[str, str]
+    _env: Mapping[str, EnvValue]
 
     # Connector control (outside of command generation)
     _success_exit_codes: Iterable[int]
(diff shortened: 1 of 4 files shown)


wowi42:
Reviewed this. The core fix is correct: swapping `QuoteString(f"{key}={value}")` for `QuoteString(StringCommand(key, value, _separator="="))` keeps the `HiddenValue` object intact until render time, so `get_raw_value` unmasks to the real secret while `get_masked_value` falls back to `HiddenValue.__str__` and shows `*MASKED*`. `_env` is only consumed in `make_unix_command` and all connectors route through it, so the change is well contained. Tests cover the raw, masked, and shell-char cases.

Two low severity nits, neither blocking:

1. `arguments_typed.py` hardcodes `None | Mapping[str, str | HiddenValue]` while `arguments.py` defines and uses the `EnvValue = str | HiddenValue` alias. The two files are meant to stay in sync. The sync linter only warns on type mismatches so CI still passes, but if `EnvValue` gains a member later this signature will silently drift. Using `EnvValue` in both keeps them coupled.

2. `generate_env` uses `if value:` rather than `if value is not None:`. Functionally harmless since the default is `{}` and `dict.update({})` is a no-op, but it is the truthy-check pattern the repo conventions call out (distinguish unset from empty).

Otherwise looks good.

Jesse-11:
Thanks @wowi42. I've pushed through changes addressing the small nits you brought up, should be good for merging 😄.

wowi42:
LGTM

PR #1856, opened 2026-06-28
style: fix ruff lint errors (PEP 585/604, f-strings, imports)

wowi42:
## What

Resolves the ruff 0.14 lint findings currently present on `3.x` (14 across 9 files), scoped to lint only with no behaviour change.

- PEP 604 unions (`X | None`, `X | Y`) in the runtime `FactBase[...]` subscripts of `facts.efibootmgr`, `facts.files` and `facts.server`, and in the `HostType` alias of `pyinfra_cli.inventory`.
- f-strings in place of `%` / `str.format` building in `connectors.scp.client`, `pyinfra_cli.virtualenv` and `progress`.
- Move the `.util.docker` import above the module-level `DOCKER_HUB_SERVER` constant in `operations.docker` (E402); the import does not depend on that constant.
- Mark the `HiddenValue` re-export in `pyinfra.api.__init__` with `# noqa: F401`, matching its sibling imports.
- Mark the two two-stage regex templates in `facts.files` with `# noqa: UP031`. There `%%s` is an intentional literal `%s` placeholder for a later substitution, and a `.format()` rewrite would need cryptic brace escaping.

## Notes

The `FactBase[...]` unions are evaluated at runtime (base-class subscripts, not annotations). They are valid because the project floor is Python 3.10 (`requires-python = ">=3.10,<4.0"`). ruff does not autofix these for that reason, so they were converted by hand and checked to construct at runtime.

## Verification

- `scripts/dev-lint.sh`: ruff check, ruff format, mypy (160 files) and the arguments sync check all pass.
- Full non-end-to-end test suite: 1855 passed.

### DIFF
--- a/src/pyinfra/api/__init__.py
+++ b/src/pyinfra/api/__init__.py
@@ -7,7 +7,7 @@ from .command import (  # noqa: F401
     StringCommand,
     MaskString,
 )
-from .hiddenvalue import HiddenValue
+from .hiddenvalue import HiddenValue  # noqa: F401 # pragma: no cover
 from .config import Config  # noqa: F401 # pragma: no cover
 from .deploy import deploy  # noqa: F401 # pragma: no cover
 from .exceptions import (  # noqa: F401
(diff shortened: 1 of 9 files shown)


DonDebonair:
LGTM! Always nice to keep the code base stylistically healthy!

PR #1858, opened 2026-07-01
fix: allow SSH connections when an ED25519-SK (FIDO2) key is present in the agent

### LINKED ISSUE #1242: SSH Key of type ED25519-SK work with pyinfra==3.1 & paramiko==3.1 but not with newer versions of paramiko (3.2+)
## Describe the bug

PyInfra cannot connect to host with ED25519-SK key and paramiko on 3.2+ version (works with 3.1)
ED25519-SK is the ssh key type use with a yubikey FA

## To Reproduce

+ setup a host with ED25519-SK (https://developers.yubico.com/SSH/Securing_SSH_with_FIDO2.html)
+ put the host in inventory.py (test= "XX.XX.XX.XX" ...)
+ pyinfra inventory.py exec --limit test -- hostname

## Expected behavior

pyinfra should connect to host but give this error:

`  File "/home/julien/.local/lib/python3.11/site-packages/paramiko/auth_handler.py", line 394, in _parse_service_accept
    key_type, bits = self._get_key_type_and_bits(self.private_key)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julien/.local/lib/python3.11/site-packages/paramiko/auth_handler.py", line 218, in _get_key_type_and_bits
    if key.public_blob:
       ^^^^^^^^^^^^^^^
  File "/home/julien/.local/lib/python3.11/site-packages/paramiko/agent.py", line 476, in __getattr__
    raise AttributeError(name)
AttributeError: public_blob
2024-11-20T09:21:46Z <Greenlet at 0x7362bfc313a0: <bound method Host.connect of Host(X.X.X.X)>> failed with AttributeError

--> An internal exception occurred:
`

mvanhorn:
## Summary

Add a small, idempotent, version-guarded shim so an ED25519-SK (FIDO2 / YubiKey) key in the SSH agent no longer breaks every connection on paramiko 3.2+. The shim wraps `AuthHandler._get_key_type_and_bits` so a key whose `public_blob` is missing falls back to `(key.get_name(), key)` instead of raising, mirroring the pending upstream paramiko fix. It is wired in `ssh.py` at connector import time, next to the existing `_retry_paramiko_agent_keys` workaround (added for paramiko/paramiko#1390).

Fixes #1242.

## Background

On paramiko 3.2+ (so pyinfra 3.2+), having an ED25519-SK key (the type `sk-ssh-ed25519@openssh.com`) loaded in the SSH agent broke connections to every host, even when that key was not the one being used.

The SK `AgentKey` paramiko builds for that type has no `inner_key` (paramiko cannot derive one), so `AgentKey.__getattr__` raises `AttributeError` on `public_blob` instead of returning `None`. Paramiko's `AuthHandler._get_key_type_and_bits` does `if key.public_blob:`, which then raises. Because it is an `AttributeError` rather than an `SSHException`/`OSError`, it slips past the connector's connect retry/except handling and surfaces as an uncaught greenlet failure ("An internal exception occurred").

This is paramiko/paramiko#2462; the upstream fix (paramiko/paramiko#2475) is not merged yet, so pyinfra needs to guard against it in the meantime. The issue thread shows this affecting several users since paramiko 3.2.

## Changes

The shim:

- probes the installed paramiko and only patches while the method still raises on a missing `public_blob`, so it stays a no-op once the fix ships upstream and never overrides a corrected implementation,
- is idempotent (a marker on the replacement makes a second install a no-op),
- preserves the original behaviour for keys that do have a `public_blob`.

## Testing

New tests in `tests/test_connectors/test_ssh.py` cover the SK fallback, the preserved `public_blob` path, idempotency, and that the patch is actually applied at connector import.

- [ ] Pull request includes documentation for any new/updated operations/facts (not applicable, no new operation/fact)

### DIFF
--- a/src/pyinfra/connectors/ssh.py
+++ b/src/pyinfra/connectors/ssh.py
@@ -20,7 +20,7 @@ from pyinfra.api.util import get_file_io, memoize
 
 from .base import BaseConnector, DataMeta
 from .scp import SCPClient
-from .ssh_util import get_private_key, raise_connect_error
+from .ssh_util import _patch_paramiko_sk_key_support, get_private_key, raise_connect_error
 from .sshuserclient import SSHClient
 from .util import (
     CommandOutput,
@@ -30,6 +30,9 @@ from .util import (
     run_local_process,
     write_stdin,
 )
+
+# Workaround for ED25519-SK agent keys until issue #1242 / paramiko#2462 is fixed.
+_patch_paramiko_sk_key_support()
 
 if TYPE_CHECKING:
     from pyinfra.api.arguments import ConnectorArguments
--- a/src/pyinfra/connectors/ssh.py
+++ b/src/pyinfra/connectors/ssh.py
@@ -31,9 +31,6 @@ from .util import (
     write_stdin,
 )
 
-# Workaround for ED25519-SK agent keys until issue #1242 / paramiko#2462 is fixed.
-_patch_paramiko_sk_key_support()
-
 if TYPE_CHECKING:
     from pyinfra.api.arguments import ConnectorArguments
 
@@ -219,6 +216,8 @@ class SSHConnector(BaseConnector):
 
     @override
     def connect(self) -> None:
+        _patch_paramiko_sk_key_support()
+
         retries = self.data["ssh_connect_retries"]
 
         try:
--- a/src/pyinfra/connectors/ssh_util.py
+++ b/src/pyinfra/connectors/ssh_util.py
@@ -1,6 +1,6 @@
 from getpass import getpass
 from os import path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from paramiko import (
     ECDSAKey,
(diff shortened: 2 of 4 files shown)


wowi42:
Thanks, the approach is sound. Paramiko offers no seam for this bug (it iterates agent keys internally, so per-connection filtering or error catching cannot fix the "breaks every connection" case), and the wrapper matches the semantics of the upstream fix in paramiko/paramiko#2475 (`hasattr(key, "public_blob") and key.public_blob`). Tests are hermetic and conventions are clean. A few points from a deeper review:

1. Patch scope and trigger. The shim fires as an import-time side effect of `connectors/ssh.py`, and the trigger is much broader than SSH use: connector entrypoint enumeration loads it eagerly, so merely building an `Inventory` in an embedding application imports `connectors.ssh` and permanently mutates paramiko's `AuthHandler` for the whole interpreter, including non-pyinfra paramiko sessions in the same process. There is also an asymmetric gap: code driving `pyinfra.connectors.sshuserclient.SSHClient` directly never imports `connectors.ssh`, so that path stays unpatched. Note the PR text cites `_retry_paramiko_agent_keys` as precedent, but that one is a per-connection retry helper, not a global patch (the closest real precedent is `sshuserclient/config.py`'s `paramiko.config.invoke = FakeInvoke`). Calling the shim from the connector's connect path or from `sshuserclient.SSHClient` would confine the mutation to actual SSH use.

2. Both failure directions of the version probe are silent. The probe calls the method with `self=None` and treats any `AttributeError` as "buggy" and any other exception as "leave it". A future implementation or a third-party wrapper that touches `self` gets either wrongly overwritten, or worse, makes the shim silently decline, resurfacing the SK bug with no log line. Since the wrapper is behaviorally identical to the upstream fix on both buggy and fixed paramiko, the probe buys little: always wrapping (keeping just the idempotency marker), or at least logging when the probe declines, removes the silent modes.

3. `getattr(AuthHandler, "_get_key_type_and_bits")` has no default and runs at import. All released paramiko versions in the allowed range (2.11 through 4.x) have the method, so this cannot fire today, but paramiko renamed the neighboring method once before (2.9 had `_get_algorithm_and_bits`), and a future 4.x rename would crash the whole SSH connector import, a worse failure than the bug being fixed. `getattr(..., None)` plus an early return is cheap hardening.

4. Minor, tests: `buggy_get_key_type_and_bits` and `FakeSkKey` are copies of paramiko's buggy method and the src `ProbeKey`. If the real SK key shape changes, the src probe and the test fakes can drift apart silently, leaving the guard's real-world behavior untested against the installed paramiko.

mvanhorn:
Thanks for the thorough review - all four points:

1. Confined the shim to actual SSH use: dropped the import-time call, now invoked from `SSHConnector.connect()` and sshuserclient `SSHClient.connect()`, so building an `Inventory` no longer mutates paramiko globally and the direct-SSHClient path is covered. Added a regression test.
2. Removed the probe's silent modes: the wrapper matches upstream on both buggy and fixed paramiko, so it always installs now (keeping just the idempotency marker).
3. Hardened the lookup with `getattr(..., None)` + early return.
4. Left the test-fake drift as a follow-up to keep this focused; happy to add a check against the installed paramiko shape if you'd prefer it here.

mvanhorn:
Thank you @wowi42! FIDO2 keys in the agent no longer block plain SSH connections.

mvanhorn:
Appreciate the merge, @wowi42! SSH connections working alongside ED25519-SK keys removes a confusing blocker for FIDO2 users.

PR #1859, opened 2026-07-01
fix(facts.server): resolve major.minor from distro release file

wlix13:

## Description
server.LinuxDistribution reports `minor: null` for releases whose `/etc/os-release` only exposes major version in VERSION_ID. CentOS 8 is example: `VERSION_ID="8"`, while precise version (8.4.2105) lives only in `/etc/centos-release`. Repos are frequently sliced by `major.minor`, so losing the minor version requires to write manual fact checks

## Fix

```python
"major": try_int(parsed.major_version(best=True)) or None,
"minor": try_int(parsed.minor_version(best=True)) or None,
```

### Tests

Add fact fixtures from live cloud instances captured output

- centos-8.json: CentOS Linux 8.4 -> asserts minor: 4 (was lost)
- centos-7.json: CentOS Linux 7.9 -> asserts minor: 9
- centos-stream-9.json: CentOS Stream 9 -> asserts minor: null
- fedora-40.json: Fedora 40 -> asserts minor: null


wowi42:
LGTM. The fix is correct and low-risk: `best=True` only changes the result where a distro release file is genuinely more precise than `VERSION_ID` (CentOS and similar EL derivatives). For Ubuntu, RHEL, Alpine and Fedora the sources tie or `os-release` already has full precision, so nothing changes there. The fact's command only collects `/etc/*-release` files, so `distro` cannot pick up sources like `/etc/debian_version`. Good fixture coverage of both the resolved and the still-null cases.

Two minor notes, neither blocking:

1. Pre-existing, not introduced here: `try_int(...) or None` turns a real minor of `0` into `None`, so an `X.0` release (e.g. RHEL 9.0) reports `minor: null`. `best=True` makes `.0` minors slightly more reachable via release files. Probably worth a separate fix with an `is None` check instead of `or`.
2. This changes fact output on existing hosts (CentOS 7 goes from `minor: null` to `minor: 9`), so it deserves a changelog mention.

wlix13:
Thanks for the review!
Changelog - added an entry. I couldn't find a documented convention, but I found example (#1714) so I followed that logic. Marked output change (CentOS 7: minor: null -> minor: 9) explicitly.

`try_int(...) or None` - good catch. I ended up fixing it in this PR rather than separately, since it's the exact two lines. One subtlety: plain `is None` check wasn't enough - try_int returns its input unchanged on failure, and distro returns "" when version part is missing. Went with an explicit int check instead.

Also added a RED OS 8.0 fixture from real host data - it's the only image with a .0 minor I could find in my cloud, and it was hitting exactly this bug (`VERSION_ID="8.0"` -> `minor: null`). Turns out there was no fixture for X.0 releases at all. As a bonus it covers the `parsed.name()` fallback and an oddly formatted release file that distro can't extract a version from. Happy to split this into a separate PR if you'd prefer.

PR #1863, opened 2026-07-04
feat(facts.server): Add OpenBSD support to server.Mounts

epicrazzmatazz:
`server.Mounts` handles Linux and FreeBSD, however, the fact fails on OpenBSD due to missing `/proc` (and it has different commandline options for `mount`). This PR adds a parser for OpenBSD's [`mount -v`](https://man.openbsd.org/mount#v).

### DIFF
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -6,8 +6,6 @@ import re
 import shutil
 from datetime import datetime
 from tempfile import mkdtemp
-from operator import methodcaller
-from itertools import takewhile
 from typing import Optional, Union
 from collections.abc import Iterable
 
@@ -336,13 +334,7 @@ class Mounts(FactBase[dict[str, MountsDict]]):
                     path = m[2]
                     device = m[1]
                     type_ = m[3]
-                    # assumes ctime flag is always last and drops it
-                    options = list(
-                        map(
-                            methodcaller("lstrip", " "),
-                            takewhile(lambda f: "ctime=" not in f, m[4].split(",")),
-                        )
-                    )
+                    options = [opt.strip(" ") for opt in m[4].split(",") if "ctime=" not in opt]
 
                     devices[path] = {"device": device, "type": type_, "options": options}
 
--- a/src/pyinfra/facts/server.py
+++ b/src/pyinfra/facts/server.py
@@ -6,6 +6,8 @@ import re
 import shutil
 from datetime import datetime
 from tempfile import mkdtemp
+from operator import methodcaller
+from itertools import takewhile
 from typing import Optional, Union
 from collections.abc import Iterable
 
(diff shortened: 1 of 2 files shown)


wowi42:
Thanks for adding OpenBSD support, the direction is good and the Linux/FreeBSD paths are untouched. A few issues in the parser, most of them share one root cause (the regex is too strict, and non-matching lines are silently skipped):

1. The device group `(/dev/[^ ]+)` can never match nfs, mfs or tmpfs mounts, even though the type alternation explicitly lists them. On OpenBSD those lines look like `server:/export on /mnt type nfs (...)`, `mfs:12345 on /tmp type mfs (...)` and `tmpfs on /tmp type tmpfs (rw)`. All three fail `fullmatch`, so the mounts silently vanish from the fact, and consumers like `server.mount` will treat them as unmounted.

2. The fs-type whitelist has a typo: `nfts` should be `ntfs` (OpenBSD ships `mount_ntfs`). More generally, any type not in the list is silently dropped with no error. Since the ` on <path> type <word> (<flags>)` structure is stable, capturing the device as `(\S+)` and the type as `(\w+)` fixes both this and point 1 without a whitelist to keep in sync with `/sbin/mount_*`.

3. `takewhile(lambda f: "ctime=" not in f, ...)` truncates the option list at ctime, but OpenBSD's `mount.c` prints filesystem-specific options after ctime for nfs/mfs/msdos (`v3`, `tcp`, `soft`, `wsize=`, ...), so for those types real options are silently discarded. A plain filter is order-independent and drops only the timestamp: `[o.strip() for o in m[4].split(",") if "ctime=" not in o]`. That also matches the FreeBSD branch's idiom just above and drops the `methodcaller`/`takewhile` imports (note `lstrip(" ")` also keeps trailing whitespace that `.strip()` would remove).

4. `command()` branches on `self._kernel.strip() == "OpenBSD"` but `process()` compares `self._kernel == "OpenBSD"` without strip. If the kernel value ever carries stray whitespace (e.g. `"OpenBSD\r"`, the connector only strips `\n`), `mount -v` runs but its output is fed to the Linux mountinfo parser, which raises ValueError. The FreeBSD branch has the same asymmetry today, so this just replicates it, but stripping once when storing `self._kernel` would fix both.

Smaller points:

- The regex is compiled inside `process()` on every call; the file's convention is module-level compiled patterns (see `_SAFE_USERNAME_RE`). And once compiled, `p.fullmatch(line)` instead of `re.fullmatch(p, line)`.
- The local variable `type` shadows the builtin; the FreeBSD branch uses `type_`.
- The fixture only covers ffs mounts, so none of the cases above are exercised. Adding one nfs or ntfs line would catch them.

One adjacent note, probably a follow-up issue rather than this PR: now that `Mounts` returns data on OpenBSD, `server.mount` with extra options on an already-mounted path emits `mount -o remount,... <path>`, and OpenBSD's mount has no `remount` (it uses `mount -u`). That branch was unreachable on OpenBSD before this fact worked.

epicrazzmatazz:
I appreciate your attention to detail. Regarding the `mount -u` point, I will eventually do something about it, as the project I've been working on that needed this fact also calls for the ability to remount a filesystem.

PR #1868, opened 2026-07-06
test: Move freezegun; `dev`→`test`

ScegfOd:
freezegun is required by test_facts.py

  - there are errors but none from this change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ repository = "https://github.com/pyinfra-dev/pyinfra"
 test = [
     "click>=8.2",
     "pytest>=8.3.5,<9",
+    "freezegun>=1.5.5",
     "coverage>=7.7.1,<8",
     "pytest-cov>=7.0.0,<8",
     "pytest-testinfra>=10.2.2,<11",
@@ -79,7 +80,6 @@ dev = [
     "ipdbplugin",
     # dev-only lint extras
     "typos>=1.36.2",
-    "freezegun>=1.5.5",
     "requests>=2.34.2",
 ]
 


0cjs:
This is failing because `(pyproject dependency-groups.test)` is an "unknown scope." I am not clear why; checking [the specification](https://www.conventionalcommits.org/en/v1.0.0/#specification), I see:

> 4. A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser)...

Ah, is it because this scope uses a noun _phrase,_ indicating the file and part of the file? I suggest we try using just `(pyproject)`. I feel this is less clear, but it's more important we follow the rules.

0cjs:
This is still failing because _this PR's title_ is not "linting." This is the first time I've ever seen a patch be bad because a PR title had an unwanted space in it, but anyway, removing the ` dependency-groups.test` string (note the leading space) should fix it, I think.

DonDebonair:
@0cjs can I ask you why you're being antagonistic in various different PR/issue discussions? I want to ask you to keep it civil. Keep sarcasm/cynicism to yourself please.

ScegfOd:
I can sincerely say that I've never had to figure out linting issues with a PR title either, though I admittedly have much less experience.

I found how to look at the error message from the linter!
The error message seems to explain a bit better than the docs, we have to match something more specific:

> Error: Unknown scope "pyproject" found in pull request title "test(pyproject): Move freezegun; `dev`→`test`". Scope must match one of: api, cli, operations(\.[A-Za-z0-9.\-]+)?, facts(\.[A-Za-z0-9.\-]+)?, connectors(\.[A-Za-z0-9.\-]+)?.

Which, for this change, all feel misleading. I'll try removing the parenthetical part of the locator entirely.

ScegfOd:
...and that seems to have green checks all the way down.
Perhaps we should use the above error message to update the docs, or use the docs to update the source of that error message and the probably-regex it uses?

DonDebonair:
You're right, we should make it more clear what the acceptable scopes are. I will open a pr for that.

DonDebonair (approved):
LGTM

0cjs:
> You're right, we should make it more clear what the acceptable scopes are.

From many go-arounds working out what was acceptable and not, my thoughts are:
1. Move "pull request title" to the top, because it's what needs the most focus. Including tests etc. with a PR is pretty obvious even to a mid developer, but it's unusual to need a particular PR title, so you want people to hit that before their eyes start to glaze over.
2. Explain that you replace the commit summaries with the PR title, as I think you do, and that you squash as well if you do that. It quite changes how I, at least, build commits if I know they're going to be changed in certain ways after.
3. Write up docs specific to this project on exactly what's allowed, because the [official full specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) does not mention anything about things that are apparently allowed, such as `chore:` and `style:`, nor does its specification of "A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser):" indicate that you're much more restrictive than that. (Yes, I know that the summary says that "_types_ other than fix: and feat: are allowed," but that doesn't really give any indication about what's allowed here, and besides, if you just click on "spec" you'll miss that completely.)
4. Maybe change the 'lint-XXX' names to something like 'pr-title-XXX' so it's a bit more obvious that this is not linting code but checking the PR title.

PR #1876, opened 2026-07-10
feat(connectors): fail fast when target uses sudo-rs

### LINKED ISSUE #1499: `sudo-rs` not supported

wowi42:
Closes #1499 (in a way)...

Detects when the target host uses `sudo-rs` instead of traditional sudo by running `sudo --version` on the first use of `_sudo`. If `sudo-rs` is detected, pyinfra fails fast with a clear error message rather than failing later with a cryptic authentication error.

- Detection result is cached per host in `host.connector_data`.
- Added tests for both SSH and local connectors.

### DIFF
--- a/src/pyinfra/connectors/util.py
+++ b/src/pyinfra/connectors/util.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from dataclasses import dataclass
 from getpass import getpass
 from queue import Queue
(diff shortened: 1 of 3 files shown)


0cjs:
This should not close #1499, and I'd rather not see this released because it will break things worse for me. (I currently have things working with sudo-rs and a monekypatch.) See that issue for details.

Sorry about not getting this information to you earlier.

wowi42:
Updated the branch. It now:

- Detects sudo-rs versions < 0.2.11 and fails fast with a clear message.
- Allows sudo-rs 0.2.11+ to run normally.
- Adds `sudo: interactive authentication is required` to the password-retry detection, which is the prefix sudo-rs 0.2.13 uses.
- Adds tests for unsupported sudo-rs, supported sudo-rs, and the new password-retry prefix.

@0cjs if you can pull this branch and test on your Ubuntu 26.04/26.10 hosts, that would confirm the fix in a real environment.

0cjs:
Unfortunately we've hit the last day of the week and I and my co-workers don't have time to do this today. We'll try for next week, but I'd suggest in the long term you spin up container hosts for testing these kinds of things. We've found that to be _very_ helpful, though I'm not in a position to release the code for that any time soon, unfortunately.

(Though thank you for reminding me; I now may have an idea why our container host didn't catch this when we tried it with Ubuntu 26.04.)

src/pyinfra/connectors/util.py
```diff
@@ -359,6 +399,7 @@ def make_unix_command_for_host(
 
     # Handle sudo password
     if command_arguments.get("_sudo"):
+        _fail_if_unsupported_sudo_rs(host)
```
Fizzadar, line 402: I think we need to make this check only if using sudo password. `sudo-rs` does work atm with `NOPASSWD` config. Should be safe to just move it down a to L408 or so.

src/pyinfra/connectors/util.py
```diff
@@ -346,6 +358,34 @@ def _ensure_askpass_set_for_host(
     return path
 
 
+def _fail_if_unsupported_sudo_rs(host: Host) -> None:
+    # Cache the check result so we only run `sudo --version` once per host.
+    if host.connector_data.get("sudo_version_checked"):
+        return
+
+    host.connector_data["sudo_version_checked"] = True
```
Fizzadar, line 366: We should move this after the run shell command call so we don't cache failures there (_ignore_errors might allow it).

Fizzadar (changes requested):
Couple of tweaks, this is an ideal solution to a shitty problem (thanks sudo-rs!).

PR #1881, opened 2026-07-11
fix(facts.git): GitLocalCommit fact failing on nonlocal ref

### LINKED ISSUE #1880: GitLocalCommit fact fails (and kills the host) when the ref doesn't exist locally, breaking git.repo on version bumps
## Describe the bug
GitLocalCommit promises in its docstring to return None "when the repository does not exist, the ref is unknown, or the command fails", and its process() is written accordingly. But the fact command doesn't guard the exit code:
```
# pyinfra/facts/git.py
return f"! test -d {repo} || (cd {repo} && git rev-parse {ref} 2>/dev/null)"
```

When the repo directory exists but ref is unknown, `git rev-parse` exits 128. `2>/dev/null` hides the message but not the exit code, so the whole fact command fails, and pyinfra aborts the host with `Error: could not load fact: git.GitLocalCommit - No hosts remaining!` instead of returning None.

## To Reproduce
1. Have an existing clone on the target whose local refs are stale (e.g. a new release tag not yet fetched).
2. Run a deploy containing:
```
git.repo(src="https://github.com/nextcloud/server.git",
          dest="/srv/nextcloud", branch="v34.0.1")
```
3. Deploy dies at fact-gathering time.

  This is easy to hit because git.repo's skip-pull check gathers GitLocalCommit for the target branch at plan time, before the git fetch it queued has run — so
  any version bump to a not-yet-fetched tag/branch triggers it.

## Expected behavior
It should work

## Meta
```
    System: Linux
      Platform: Linux-6.12.85+deb13-amd64-x86_64-with-glibc2.41
      Release: 6.12.85+deb13-amd64
      Machine: x86_64
    pyinfra: v0.0.0.post4460.dev0+a5fb8bde
      click: v8.4.0
      distro: v1.9.0
      gevent: v26.4.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.15.0
    Executable: /home/jan/ggu-infra/.venv/bin/pyinfra
    Python: 3.13.5 (CPython, GCC 14.2.0)
```

uv

yagarea:
Fixes: #1880 

---

Its docstring explicitly promises: "returns None when the repository does not exist, the ref is unknown, or the command fails." And the `process()` method is written for that — it validates the output against a SHA regex and returns None otherwise.

But the shell command doesn't hold up its end. When the directory exists and the ref is unknown, `git rev-parse` exits with code 128. The `2>/dev/null` hides the error text but does nothing about the exit code, so the whole fact command exits non-zero. pyinfra treats a failed fact command as fatal — `Error: could not load fact` — kicks the host out, and you get `No hosts remaining!`.

-------

As per AI policy I disclose that test were created with assistance of LLM. I have read and tested them manually so its human verified.

----

### DIFF
--- a/src/pyinfra/facts/git.py
+++ b/src/pyinfra/facts/git.py
@@ -83,7 +83,7 @@ class GitLocalCommit(GitFactBase):
 
     @override
     def command(self, repo: str, ref: str = "HEAD") -> str:
-        return f"! test -d {repo} || (cd {repo} && git rev-parse {ref} 2>/dev/null)"
+        return f"! test -d {repo} || (cd {repo} && git rev-parse {ref} 2>/dev/null) || true"
 
     @override
     def process(self, output: list[str]):
--- a/tests/facts/git.GitLocalCommit/branch_ref.json
+++ b/tests/facts/git.GitLocalCommit/branch_ref.json
@@ -3,7 +3,7 @@
         "repo": "myrepo",
         "ref": "mybranch"
     },
-    "command": "! test -d myrepo || (cd myrepo && git rev-parse mybranch 2>/dev/null)",
+    "command": "! test -d myrepo || (cd myrepo && git rev-parse mybranch 2>/dev/null) || true",
     "requires_command": "git",
     "output": [
         "1a2b3c4d5e6f7890abcdef1234567890abcdef12"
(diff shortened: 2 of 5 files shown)


wowi42 (approved):
One small note: the new fixture's `output` is `["v34.0.1"]` but on a real system `git rev-parse` outputs nothing to stdout when the ref is unknown — the error goes to stderr which is `/dev/null`. The fixture should be `"output": []`. The test passes either way (empty-list short-circuit and SHA-regex mismatch both return None), so this is cosmetic.

yagarea:
@wowi42 Thank you for review. The output field in `unknown_ref` test should be fixed now.

PR #1882, opened 2026-07-12
feat(operations.docker.volume): allow options to be passed when creating docker volumes

DonDebonair:
Options (`--opt`) are required when creating NFS volumes and other types of volumes. We make no attempt at modelling all possible options, because that would brittle in light of future Docker updates. Instead, a list of strings can be passed. Not all options are kv pairs, so we don't use a dict.


wowi42 (approved):
LGTM

PR #1884, opened 2026-07-14
fix(cli): preserve Windows paths in module errors

### LINKED ISSUE #1861: Cli "module not found" error fails if passed a full Windows path
## Describe the bug
When trying to use the cli with a full windows path of a non existing file, the error message is broken.  
For example when calling : `pyinfra inventory.py D:/non_existing_script.py`
The message printed will be: `pyinfra_cli.exceptions.CliError: No such module: D`

This happens because `try_import_module_attribute` in `pyinfra_cli/util.py` has a "Allow a.module.name:function syntax" logic that don't expect a `:` to be present in the path. 

## To Reproduce
Be on windows, then run `pyinfra inventory.py D:/non_existing_script.py`

## Expected behavior
Full module name in the error message, as passed to the cli.

## Meta
```
System: Windows
      Platform: Windows-11-10.0.26200-SP0
      Release: 11
      Machine: AMD64
    pyinfra: v3.7
      click: v8.3.2
      distro: v1.9.0
      gevent: v26.4.0
      jinja2: v3.1.6
      packaging: v26.0
      paramiko: v3.5.1
      pydantic: v2.13.0
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.1
      typing-extensions: v4.15.0
    Executable: D:\work\sgl-pyinfra\.venv\Scripts\pyinfra
    Python: 3.14.3 (CPython, MSC v.1944 64 bit (AMD64)
```

0xTaoZ:
## Summary
- preserve full Windows drive paths before parsing module:attribute syntax
- add regression coverage for forward-slash and backslash Windows paths

Fixes #1861.

## Validation
- uv run pytest tests/test_cli/test_cli_util.py -q
- uv run pytest tests/test_cli -q
- uv run ruff format --check src/pyinfra_cli/util.py tests/test_cli/test_cli_util.py
- uv run ruff check src/pyinfra_cli/util.py tests/test_cli/test_cli_util.py
- uv run mypy src/pyinfra_cli/util.py
- git diff --check HEAD~1..HEAD

## Risk
Low: the change only bypasses module/attribute parsing for explicit Windows drive paths such as `D:/...` and `D:\...`; existing dotted and `module:function` imports keep the previous flow.

### DIFF
--- a/src/pyinfra_cli/util.py
+++ b/src/pyinfra_cli/util.py
@@ -153,7 +153,16 @@ def parse_cli_arg(arg):
     return arg
 
 
+def _is_windows_drive_path(path: str) -> bool:
+    return len(path) >= 3 and path[1] == ":" and path[2] in ("/", "\\")
+
+
 def try_import_module_attribute(path, prefix=None, raise_for_none=True):
+    if _is_windows_drive_path(path):
+        if raise_for_none:
+            raise CliError(f"No such module: {path}")
+        return None
+
     if ":" in path:
         # Allow a.module.name:function syntax
         mod_path, attr_name = path.rsplit(":", 1)
--- a/tests/test_cli/test_cli_util.py
+++ b/tests/test_cli/test_cli_util.py
@@ -102,3 +102,11 @@ def test_try_import_module_attribute_falls_through_when_attr_missing(monkeypatch
     result = try_import_module_attribute("pip.Pip3Packages", prefix="pyinfra.facts")
 
     assert result is Pip3Packages
+
+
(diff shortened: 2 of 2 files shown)


wowi42 (approved):
The fix is correct and tight. It adds a 3-line check that spots Windows drive paths (`D:/...` or `D:\...`) before the `:` split logic runs, so the full path stays in the error message. Both slash styles are tested. All CI is green.

One small note: the helper matches any letter before `:`, not just `A-Z`, but that is harmless since no valid Python module name starts with a digit-colon-slash either.

PR #1891, opened 2026-07-19
chore: deduplicate shellcheck lint command

### LINKED ISSUE #1877: Duplicate shellcheck logic between dev-lint.sh and GitHub workflow
### Problem

The shellcheck invocation is duplicated between the developer script and the GitHub Actions workflow:

- `scripts/dev-lint.sh` lines 15-21
- `.github/workflows/shellcheck.yml`

This means the command, the file selection logic, and any future fixes need to be maintained in two places.

### Suggested fix

Move the shellcheck logic into a single script under `scripts/` and have both `scripts/dev-lint.sh` and `.github/workflows/shellcheck.yml` call it. This keeps local developer runs and CI in sync without changing the fail-fast behavior of `dev-lint.sh`.

0xTaoZ:
## Summary

- add `scripts/dev-shellcheck.sh` as the single shellcheck entrypoint
- have `scripts/dev-lint.sh` and the Shellcheck GitHub Actions workflow call that script

## Why

The shellcheck command and file-selection regex were duplicated between the local developer lint script and CI. Keeping the logic in one executable script keeps local and CI shell linting behavior aligned while preserving the existing fail-fast handling when shellcheck is missing.

Fixes #1877.

## Validation

- `scripts/dev-shellcheck.sh`
- `shellcheck scripts/dev-shellcheck.sh scripts/dev-lint.sh`
- `git diff --check`
- `scripts/dev-lint.sh`

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -21,6 +21,4 @@ jobs:
       - name: Install shellcheck
         run: sudo apt-get update && sudo apt-get install -y shellcheck
       - name: Run shellcheck
-        run: |
-          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' -- ':!*.py' \
-            | xargs shellcheck
+        run: scripts/dev-shellcheck.sh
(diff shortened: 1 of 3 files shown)


wowi42 (approved):
LGTM

Clean extraction of the shellcheck command into a single script, shared by CI and dev-lint.sh. No behavior change, all checks green, shellcheck workflow still passes with the new entrypoint.

PR #1892, opened 2026-07-20
fix(cli): parse literal list arguments

### LINKED ISSUE #1869: Cannot set operation argument as list from CLI
## Describe the bug
The operation `server.user` accepts an argument `groups` which needs to be a list of strings.
There is no way to define a list when calling this operation from the CLI - it always ends up as list of single characters.

## To Reproduce
This all does not work:
```shell
pyinfra INVENTORY.py server.user user=abc groups=[group1]
pyinfra INVENTORY.py server.user user=abc groups=["group1"]
pyinfra INVENTORY.py server.user user=abc 'groups=[group1]'
pyinfra INVENTORY.py server.user user=abc groups='["group1"]'
pyinfra INVENTORY.py server.user user=abc 'groups=["group1"]'
```
because the CLI does not interpret the arguments.

## Expected behavior
It should be possible to set arguments which require lists.
Just changing `kwargs[key]` = value to e.g.
```python
try:
    kwargs[key] = ast.literal_eval(value)
except (SyntaxError, ValueError):
    kwargs[key] = value
```
should be sufficient.

## Meta
pyinfra v3.9.2
Other facts are not relevant.

0xTaoZ:
## Summary

Fixes pyinfra-dev/pyinfra#1869.

CLI operation kwargs already accepted JSON containers, but Python-style literal containers such as `groups=['wheel', 'docker']` were left as raw strings. For list-valued operation kwargs like `server.user(..., groups=...)`, that meant the operation received a string and iterated it as characters.

This adds a safe `ast.literal_eval` fallback after the existing bool/int/JSON parsing and only returns container literals (`dict`, `list`, `tuple`). Scalar strings such as `'hello'` remain unchanged.

## Validation

- `uv run pytest tests/test_cli/test_cli_util.py -q`
- `uv run pytest tests/test_cli/test_cli.py -q`
- `uv run ruff format --check src/pyinfra_cli/util.py tests/test_cli/test_cli_util.py`
- `uv run ruff check src/pyinfra_cli/util.py tests/test_cli/test_cli_util.py`
- `uv run mypy src/pyinfra_cli/util.py`
- `git diff --check HEAD~1..HEAD`

## Risk

Low. The fallback is only reached after existing parsing fails, uses `ast.literal_eval` rather than evaluation, and only changes parsed container literals.

## AI assistance

OpenAI Codex was used to inspect the existing parsing path, implement the patch and regression tests, and run the validation commands above.

### DIFF
--- a/src/pyinfra_cli/util.py
+++ b/src/pyinfra_cli/util.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ast
 import json
 from datetime import datetime
 from importlib import import_module
@@ -150,6 +151,14 @@ def parse_cli_arg(arg):
         return json.loads(arg)
     except ValueError:
         pass
+
+    try:
+        parsed_arg = ast.literal_eval(arg)
+    except (SyntaxError, ValueError):
+        pass
+    else:
+        if isinstance(parsed_arg, (dict, list, tuple)):
+            return parsed_arg
 
     return arg
 
--- a/tests/test_cli/test_cli_util.py
+++ b/tests/test_cli/test_cli_util.py
@@ -54,6 +54,14 @@ class TestCliUtil(TestCase):
             (["one", "two"], {"hello": "world"}),
         )
 
+    def test_setup_op_and_python_literal_list_kwarg(self):
+        commands = ("server.user", "one", "groups=['wheel', 'docker']")
+
+        assert get_func_and_args(commands) == (
+            server.user,
+            (["one"], {"groups": ["wheel", "docker"]}),
+        )
+
 
 @pytest.fixture(scope="function")
 def user_sys_path():


wowi42:
LGTM, just fix the CI

0xTaoZ:
@wowi42 The original failures came from a temporary outage in the PR-title action rather than the patch. I rebased onto the current `3.x`; the focused tests pass locally and the fresh CI run is now 30/30 green.

src/pyinfra_cli/util.py
```diff
@@ -151,6 +152,14 @@ def parse_cli_arg(arg):
     except ValueError:
         pass
 
+    try:
+        parsed_arg = ast.literal_eval(arg)
```
Fizzadar, line 156: One issue here - this will parse `80,443` into a tuple now when it was previously a string, which some `--data` args might rely on. One option could be restricting the fallback to explicit containers only.

Fizzadar (changes requested):
This is great - one small issue I noticed which would be a breaking change, I think we can work around it.

src/pyinfra_cli/util.py
0xTaoZ: @Fizzadar I restricted the fallback to explicitly wrapped containers (`[]`, `{}`, or `()`), so values like `80,443` remain strings. I added a regression test and rebased onto the current `3.x` in 2a61a87d.

Fizzadar (approved):
LGTM, thank you @0xTaoZ!

PR #1894, opened 2026-07-20
refactor(cli): migrate from Click to Cyclopts + Rich

noirbizarre:
Migrates the pyinfra CLI from Click to [Cyclopts](https://cyclopts.readthedocs.io/) for argument parsing and adopts [Rich](https://rich.readthedocs.io/) for all human-facing output, while keeping the existing command syntax and behaviour.

Highlights:

- **Cyclopts-based CLI** with grouped `--help`, colourised usage, shell completion, and a machine-readable `--json` mode (stdout stays pure JSON; all human output goes to stderr).
- **Rich `debug-inventory`**: hosts rendered in a zebra-striped table with data shown as `key: value` lines (nested values as JSON), type-coloured to match the JSON highlighter.
- **Live hierarchical progress tree** during deploys, with per-host/per-operation nodes and informative error handling that attributes failures to the right host and operation.

Output is routed through a small pluggable layer in `pyinfra.api.output`, decoupled from Rich/Click so the core stays testable. Existing behaviour, exit codes and command syntax are preserved; a few intentional UX changes (usage-error exit code, `--version` output, confirm-prompt re-prompting) are documented in the commits.

See https://github.com/pyinfra-dev/pyinfra/discussions/1885 for context (includes recordings and screenshots).


wowi42 (approved):
LGTM

This migrates the CLI from Click to Cyclopts for argument parsing and Rich for human-facing output, while keeping the existing command syntax and all CLI flags intact. The new output layer in pyinfra.api.output stays decoupled from Rich/Click so the core remains testable, and the live progress tree during deploys is a nice improvement. All CI passes, including the full end-to-end matrix.

The shell injection surface is clean: user input still flows through the same operation/fact code paths, and Cyclopts handles argument parsing without introducing any new quoting concerns. The _lenient_bool converter for PYINFRA_YES preserves the legacy Click behavior including on/off and empty-string-as-false. The host.print_prefix change from [hostname] to hostname (with color and dimmed connector prefix) is an internal API change, not a public contract, and the e2e tests were updated to match. The operation_end callback reordering (before fail_hosts instead of after) is deliberate so the progress tree can finalize its display before error handling prompts or raises.

One small thing: the _support_command registered as @app.command(name="--support") and the support parameter in cli() both handle --support. The parameter body (if support: print_support_info(); return) looks unreachable since Cyclopts routes --support to the subcommand. It is harmless dead code but worth cleaning up in a follow-up.

Fizzadar (changes requested):
Hi @noirbizarre - I'm SUPER excited by this PR! But I think there's a few things to address:

- `--serial/--no-wait` loose all output under a TTY
- `--diff` also drops output unless `-v` is set with a TTY
- all host attributed logs from deploy code also dropped with TTY

This is also a significant change, I believe it would be better split into: click -> cyclopts, then Rich output / print and finally the live progress. This would make reviewing much easier as well if you don't mind 🙏

noirbizarre:
Hi @Fizzadar 

Thanks for the feedback; it was very useful. Especially the `--diff` case, which allowed me to start handling reusable prettified types.

So there it is; I split this PR into 3 separate stacked PRs (I'll keep this one as reference):
- https://github.com/pyinfra-dev/pyinfra/pull/1901
- https://github.com/pyinfra-dev/pyinfra/pull/1902
- https://github.com/pyinfra-dev/pyinfra/pull/1903

Those PRs include fixes for the case you both highlighted. I also provided some recordings, including more of those cases.

> [!IMPORTANT]
> I tried to have as stable and consistent as possible intermediate state for the PRs.
> But, given Click is handling both arg parsing and styling while Cyclopts only do parsing and rich only do styling/presentation, and given live+tree was my solution to handle by task/host isolation, the intermediate states are really created for review purposes and not truly intended to last, especially the rich without live+tree.

> [!NOTE]
> Also not that they are all targeting the `3.x` branch as I that can't target my intermediate branches which are on my fork. Review on #1902 will only make sense once #1901 is reviewed and merged and `refactor/rich` rebase on the updated `3.x`... I'll keep them in draft until then. I'll update the recordings each time on rebase too as addressing review comments might change them

Fizzadar:
@noirbizarre awesome, thank you! I think we can certainly merge Cyclopts migration on it's own first, then possibly the others together as you say. Going to focus on that PR first so we can start landing these changes ASAP!

wowi42:
Closing this PR: the proposal and its details have been moved to the discussion at https://github.com/pyinfra-dev/pyinfra/discussions/1885#discussioncomment-18143976 for further design discussion. The branch remains available if we decide to pick this up later.

Thanks @noirbizarre for the thorough work and the detailed write-up.

PR #1897, opened 2026-07-22
docs(facts.deb): clarify package resolution

### LINKED ISSUE #1860: Properly document get_fact(deb.DebPackage, 'somename')
When calling `get_fact(deb.DebPackage, 'somename')` , _somename_ may be either an installed package or a package file; how this is decided should be documented. (I filed #1795 in part due to the lack of this documentation.)

The documentation needs to state three things:

1. ___This call is not idempotent.___ It may behave differently depending on the contents of the directory that happens to be the CWD when it's called.

2. The rules for whether _somename_ is treated as a request for information about a package file or an installed package are as follows.

   a. If _somename_ does not match the shell glob `*.deb` the installed package database is queried.

   b. If _somename_ does match the shell glob `*.deb`:
      1. If a file in the CWD matching that name exists, that package file is queried.
      2. If a file in the CWD matching that name does not exist, the installed package database is queried.

3. This breaks backward compatibility from versions before [whatever version this is released in]: querying e.g. `somepackage.deb.old` when a package file with that name is present will now fail.

We should probably also note that this prohibits querying any package files that do not end in `.deb`, e.g., `somepackage.deb.old` or `somepackage.deb~`.

FWIW, I still believe that, since we're going to break backward compatibility anyway, using the scheme I suggested in #1795 (put a slash in any _somename_ that's supposed to reference a file) is the better solution because it makes this idempotent and gives an error when someone expects a file to be there that isn't, rather than misleadingly giving package information for a potentially different package.

Sanjays2402:
Closes #1860

Documents how `DebPackage` chooses between archive inspection and the installed package database, including the current-working-directory dependency and unsupported archive suffixes. Adds a focused documentation regression test.

### DIFF
--- a/src/pyinfra/facts/deb.py
+++ b/src/pyinfra/facts/deb.py
@@ -60,6 +60,19 @@ class DebPackages(FactBase):
 class DebPackage(FactBase):
     """
     Returns information on a .deb archive or installed package.
+
+    Package resolution depends on the remote current working directory and is
+    therefore not idempotent:
+
+    * Names that do not match the shell pattern ``*.deb`` query the installed
+      package database.
+    * Names that match ``*.deb`` inspect the archive only when that file exists
+      in the current working directory. Otherwise they query the installed
+      package database.
+
+    Consequently, archive names such as ``package.deb.old`` and
+    ``package.deb~`` cannot be queried as files. Older pyinfra versions tried
+    to inspect any existing path as an archive instead.
     """
 
     _regexes = {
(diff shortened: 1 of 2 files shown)


wowi42 (approved):
LGTM

Clean docs-only addition that explains how DebPackage resolves between archive inspection and the installed package database. The docstring is accurate against the shell command in command(), and the test ensures the documentation stays in place. All CI passes.

Issue #1136, opened 2024-06-30
Setting user SSH keys fails when authorized_keys file is a symlink


taliaferro:
## Describe the bug

Tasks like `server.user` and `server.user_authorized_keys` check to make sure the user's `~/.ssh/authorized_keys` file is indeed a regular file, and fail if it isn't. However, the assumption that `authorized_keys` is a regular file doesn't necessarily hold everywhere. 

On Proxmox, `/root/.ssh/authorized_keys` is a symlink to `/etc/pve/priv/authorized_keys` because `/etc/pve` is a filesystem that's synchronized across all nodes in a Proxmox cluster. In this scenario this operation will fail.

## To Reproduce

Make a file somewhere else and link `.ssh/authorized_keys` to it. Use the "user_authorized_keys" task to try to set that user's authorized SSH keys.

```bash
touch ~/.ssh/authorized_keys
mv ~/.ssh/authorized_keys ~/.ssh/real_authorized_keys
ln -s ~/.ssh/real_authorized_keys ~/.ssh/authorized_keys
pyinfra @local server.user_authorized_keys public_keys=~/.ssh/id_rsa.pub user=$USER
```

## Expected behavior

If `~/.ssh/authorized_keys` is a symlink, Pyinfra should instead check to see if the link target is a regular file, then proceed as normal.

## Meta
+ Include output of `pyinfra --support`.
```
    System: Linux
      Platform: Linux-6.9.6-arch1-1-x86_64-with-glibc2.39
      Release: 6.9.6-arch1-1
      Machine: x86_64
    pyinfra: v3.0b1
    Executable: /home/taliaferro/.local/share/virtualenvs/infra-gRqFrJWx/bin/pyinfra
    Python: 3.12.3 (CPython, GCC 13.2.1 20240417)
```
+ How was pyinfra installed (source/pip)? Pip

simonhammes:
Setting `authorized_key_filename`/`authorized_key_directory` to the values of the actual file fixes the issue for me (at least for `server.user_authorized_keys()`).

Unfortunately `server.user()` does not support those two keyword arguments.

---

> If ~/.ssh/authorized_keys is a symlink, Pyinfra should instead check to see if the link target is a regular file, then proceed as normal.

Of course this would be nicer for end users :slightly_smiling_face:

0cjs:
I just got hit by this, too. Here are my thoughts:

[...]

PR #1898, opened 2026-07-23
fix(operations.server): Allow symlinks

ScegfOd:
Plausible fix for #1136 

It seems that we generally wish to not allow symlinks to be followed for security reasons, but in the case of a user replacing their sensitive authorized_keys file with a symlink to said sensitive file's real location, it feels safe enough to follow at least that chain of symlinks.

  - Assuming that I shouldn't note that user_authorized_keys() accepts symlinks the same way `sshd` does, this is done
  - there are errors but none from this change

### DIFF
--- a/src/pyinfra/operations/server.py
+++ b/src/pyinfra/operations/server.py
@@ -949,6 +949,18 @@ def user_authorized_keys(
 
     public_keys = [key for key_or_file in public_keys for key in read_any_pub_key_file(key_or_file)]
 
+    # follow symlinks to the actual authorized_keys location:
+    while True:
+        auth_key_link = host.get_fact(Link, f"{authorized_key_directory}/{authorized_key_filename}")
+        if not auth_key_link:
+            break  # we're targeting the final file
+        link_target = auth_key_link["link_target"]
+        authorized_key_filename = os.path.basename(link_target)
+        if "/" == str(link_target)[0]:
+            authorized_key_directory = os.path.dirname(link_target)
+        else:
+            authorized_key_directory += "/" + os.path.dirname(link_target)
+
     # Ensure .ssh directory
     # note that this always outputs commands unless the SSH user has access to the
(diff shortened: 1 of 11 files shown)


wowi42 (changes requested):
Thanks for tackling #1136, the Proxmox case is real and worth fixing. Two blocking issues with the current approach, plus some smaller notes.

**Blocking**

1. Infinite loop on symlink cycles. The `while True` loop has no depth limit and no visited set. `ln -s a b; ln -s b a` (or a self-link) hangs operation generation forever. Please cap the chain (Linux gives up at 40, `ELOOP`) and raise `OperationError` when the limit is hit.

2. The resolution reassigns `authorized_key_directory`, and the "Ensure .ssh directory" block below then ensures the *resolved* directory with `mode=700` and `user`/`group` ownership. On Proxmox that directory is `/etc/pve/priv`, a cluster-shared filesystem directory: pyinfra would try to chown/chmod it as if it were `~/.ssh`. The directory ensure should keep targeting the original `~/.ssh` directory; only the file-level operations (and the `AuthorizedKeys`/`FindInFile` fact lookups) should use the resolved path.

**Non-blocking**

- A relative target with no directory component is the most common hand-made setup (`ln -s real_authorized_keys authorized_keys`). There `os.path.dirname` returns `""`, so the directory gains a trailing slash and every downstream path is double-slashed and unnormalized (`/home/u/.ssh//real_authorized_keys`). It works in the shell, but please normalize with `os.path.normpath`/`os.path.join` so fact keys and logged commands stay clean. A fixture for this case would be good too.
- Style: `if "/" == str(link_target)[0]` should be `if link_target.startswith("/")`. `Link` already returns a string.
- New fixtures should be YAML, not JSON (`follow_symlinks.json`). The edits to existing JSON fixtures are fine.

CI is green and the `Link`-fact mock additions to the existing fixtures look right.

PR #1899, opened 2026-07-24
fix(api): omit variadic kwargs from auto labels

### LINKED ISSUE #1896: Exclude **data from files.template automatic operation labels
When using `files.template` operation template values are output to stdout which can include secret information such as passwords. This is probably not a secure default, example:

```python
context = {"password": securely_get_password()}
files.template(
    src="C:/config.toml",
    dest="C:/config.toml",
    **context,
)
```

```shell
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [@winssh/my-server] Connected

    Loading: pyinfra/tasks/task.py
    [@winssh/my-server] Loaded fact pyinfra_windows.facts.files.File (path=C:/config.toml)

--> Detected changes:
    Operation                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  Change                                         Conditional Change
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  1 (@winssh/my-server)   -
    files.template (src=C:\config.toml.j2, dest=C:/config.toml, password='!CLEARTEXTPASSWORD')   1 (@winssh/my-server)   -

```

I think the best for the default behavior is to not use any `**data` args when building this label

0xTaoZ:
## Summary

Fixes #1896.

Automatic operation labels are generated from the call signature when `name=` is omitted. For operations such as `files.template`, arbitrary `**data` values can include template secrets and were being copied into the generated label metadata.

This changes auto-label argument capture to keep positional arguments and declared keyword parameters, while skipping extra variadic `**kwargs` values. The operation still receives the data and renders templates normally; only the generated label metadata changes.

## Validation

- `uv run pytest tests/test_api/test_api_operations.py -k template_op_auto_name_does_not_include_data_kwargs -q`
- `uv run pytest tests/test_api/test_api_operations.py -q`
- `uv run pytest tests/test_api -q`
- `uv run ruff format --check src/pyinfra/api/operation.py tests/test_api/test_api_operations.py`
- `uv run ruff check src/pyinfra/api/operation.py tests/test_api/test_api_operations.py`
- `uv run mypy src/pyinfra/api/operation.py`
- `git diff --check`

## Risk

Low. This only changes pyinfra-generated labels when an operation accepts arbitrary `**kwargs`; explicitly declared parameters and explicit `name=` labels are preserved.

### DIFF
--- a/src/pyinfra/api/operation.py
+++ b/src/pyinfra/api/operation.py
@@ -8,7 +8,7 @@ to the deploy state. This is then run later by pyinfra's ``__main__`` or the
 from __future__ import annotations
 
 from functools import wraps
-from inspect import signature
+from inspect import Parameter, signature
 from io import StringIO
 from types import FunctionType
 from typing import TYPE_CHECKING, Any, cast
@@ -300,7 +300,7 @@ def _wrap_operation(func: Callable[P, Generator], _set_in_op: bool = True) -> Py
 
         # Attach normal args, if we're auto-naming this operation
         if add_args:
-            op_meta = attach_args(op_meta, args, kwargs)
+            op_meta = attach_args(op_meta, func, args, kwargs)
 
         # Check if we're actually running the operation on this host
         # Run once and we've already added meta for this op? Stop here.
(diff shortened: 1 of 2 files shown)


wowi42 (commented):
The approach is correct and the dedup/execution paths are unaffected (`op_hash` comes from the call stack, `op_meta.args` is display-only). Two things before this can move:

- CI is red: `lint` failed because the branch sits on an old base (pre-`d4db499e`) and the ruff action installed ruff 0.16, which wants to reformat 19 unrelated files. Your files are not among them. Please rebase on current `3.x` and re-run; the unit tests were skipped because of the lint gate, so they have not actually run in CI yet.
- Since the motivation is keeping secrets out of auto labels, note that secrets passed via *declared* parameters still land in labels (e.g. `server.user(password=...)`). Pre-existing and arguably out of scope, but the leak surface remains after this PR.

Minor:

- `signature(func)` is computed per host per auto-named op in `attach_args`. On large inventories this repeats identical introspection; consider caching it per function.
- The new `func` parameter on `attach_args` is untyped.

0xTaoZ:
@wowi42 I rebased this onto the current `3.x` and moved signature inspection out of the per-host path, so it now runs once when the operation wrapper is built. `attach_args` now receives a typed `set[str]`; declared parameters intentionally remain visible because this PR is limited to arbitrary template data, as noted in the description. Fresh CI is running on 9ff7056f.

PR #1900, opened 2026-07-25
feat(facts.podman): add more basic facts

lsg551:
## Description

The facts for Podman as of now (v3.9.2) include two basic facts:

1. `podman.PodmanPs` to list the existing running/exited/... containers (equivalent to `podman ps -a`)
2. `podman.PodmanSystemInfo` to get Podman system information (=`podman system info`)

This PR extends this list with the following new facts:

1. `.PodmanImages` to get the output from `podman image ls` / list
    container image in the local storage
2. `.PodmanVolumes` similarly lists volumes
3. `.PodmanPods` lists pods
4. `.PodmanQuadlets` lists quadlets files (`podman quadlet list`, available since Podman v5.6.0)
5. `.PodmanNetworks` lists podman networks

Nothing fancy, it just (like the existing two) wraps the Podman CLI and forwards the JSON output.

## Checklist

## Caveats

- I wasn't sure how exceptions should be handled in fact processing. The existing code seems to be inconsistent and mostly lets the exception propagate / using `assert`s. I decided to wrap any exception raised while fact processing in the `FactProcessError`. Let me know what's the preference here.
- The `podman quadlet <cmd>` stuff was only recently added in Podman [v5.6.0](https://github.com/podman-container-tools/podman/releases/tag/v5.6.0) (published 2025-08-15). The new fact `podman.PodmanQuadlets` is using this and would fail if this version constraint is not met. I added a note in the docstring. If this is desired, I could also implement an internal version check based on the existing facts (which would still cause a failure but point the user to the exact root cause...). (But I don't know where/how exactly - yet.)

PS: sorry if I should have opened an issue before (the contributing guidelines did not mention this though)


wowi42 (changes requested):
Nice addition, the helper and the docstring examples are good. One blocker:

`PodmanNetworks.process` contains leftover debug code that writes `test_podman.json` to the current working directory on every fact call. I checked out the branch and ran the podman fact tests: the file lands in the repo root as an untracked artifact. In production this would write a file on the client machine every time the fact runs. Please remove it.

Smaller notes:

- `PodmanPs.process` still uses `assert isinstance(output, list)` plus `# type: ignore`. Now that `_parse_podman_json_list` exists, route `PodmanPs` through it too. (`assert` in `src/` is dropped under `python -O`, so the helper is also the safer pattern.)
- On your caveat about exception handling: wrapping parse failures in `FactProcessError` is the right call, keep it. The bare `except Exception` clause after `except json.JSONDecodeError` is redundant though; one or the other is enough.
- Typo in the helper docstring: "parasing".
- New fixtures should be YAML rather than JSON (project convention for new fixtures). Not blocking since the existing podman fixtures are JSON too, but worth doing.
- The Podman version note on `PodmanQuadlets` (>= 5.6.0, unchecked) is fine.

Tests and lint pass on the branch, docs examples look right.

lsg551:
> `PodmanNetworks.process` contains leftover debug code that writes `test_podman.json` to the current working directory on every fact call. I checked out the branch and ran the podman fact tests: the file lands in the repo root as an untracked artifact. In production this would write a file on the client machine every time the fact runs. Please remove it.

Yep, that's a testing artifact that should not have been included but slipped into the commit. Updated that commit accordingly, sorry/thanks for the catch.

> - `PodmanPs.process` still uses `assert isinstance(output, list)` plus `# type: ignore`. Now that `_parse_podman_json_list` exists, route `PodmanPs` through it too. (`assert` in `src/` is dropped under `python -O`, so the helper is also the safer pattern.)

Refactored accordingly.

> - On your caveat about exception handling: wrapping parse failures in `FactProcessError` is the right call, keep it. The bare `except Exception` clause after `except json.JSONDecodeError` is redundant though; one or the other is enough.

Dropped `json.JSONDecodeError`.

> - Typo in the helper docstring: "parasing".
> - New fixtures should be YAML rather than JSON (project convention for new fixtures). Not blocking since the existing podman fixtures are JSON too, but worth doing.

Typo fixed and testing fixtures migrated from JSON to YAML, including existing and new ones for Podman.

src/pyinfra/facts/podman.py
```diff
@@ -1,14 +1,32 @@
 from __future__ import annotations
 
+from pyinfra.api.exceptions import FactProcessError
```
wowi42: This import landed between `__future__` and the stdlib imports. `FactProcessError` is re-exported from `pyinfra.api`, so it can join the existing `from pyinfra.api import FactBase` below.

src/pyinfra/facts/podman.py
```diff
@@ -1,14 +1,32 @@
...
+def _parse_podman_json(value: Iterable[str], expected: type[PodmanJsonT]) -> PodmanJsonT:
+    """Parse podman output with `--format=json` option as JSON.
+
+    Raises:
+        FactProcessError: Exceptions captured during parsing are wrapped.
+    """
+    try:
+        output = json.loads("".join(value))
+    except Exception as e:
+        raise FactProcessError("process fact") from e
```
wowi42: Nit: `process fact` alone is vague when this surfaces in logs. Include the original error, e.g. `raise FactProcessError(f"failed to parse podman JSON output: {e}") from e`.

src/pyinfra/facts/podman.py
```diff
@@ -34,14 +52,49 @@ def command(self) -> str:
...
+        return _parse_podman_json(output, dict)
 
 
 class PodmanPs(PodmanFactBase[list[dict[str, Any]]]):
-    """
-    Output of 'podman ps'
+    """Returns a list of all containers.
+    This is equivalent to ``podman ps --all``.
+
+    .. code:: python
```
wowi42, line 63: Missing blank line after `.. code:: python`. Without it docutils does not parse the block as directive content, so the example will not render in the generated docs (see `zypper.py` or `rpm.py`, and `PodmanNetworks` below, for the convention). Same in `PodmanImages`, `PodmanVolumes`, `PodmanPods` and `PodmanQuadlets`.

src/pyinfra/facts/podman.py
```diff
@@ -50,6 +103,192 @@ def command(self) -> str:
...
+    @override
+    def process(self, output: Iterable[str]) -> list[dict[str, Any]]:
+        return _parse_podman_json(output, list)
+
+
+class PodmanQuadlets(PodmanFactBase[list[dict[str, Any]]]):
+    """Returns a list of podman quadlets.
+    This is equivalent to ``podman quadlet list``.
+
+    **Note**: Available since Podman [v5.6.0][release_notes]. This version
```
wowi42, line 231: This is Markdown link-reference syntax, but docstrings are rendered as rst, so the link will show up as literal text. Use an rst inline link instead, e.g. `` `Podman v5.6.0 <https://github.com/containers/podman/releases/tag/v5.6.0>`_ ``.

wowi42 (commented):
Thanks for the updates, everything from the previous round is addressed and CI is green. The facts themselves look good. A few small docstring issues left, see inline comments.

src/pyinfra/facts/podman.py
lsg551: My bad, I am used to ruff / isort doing this automatically for me. Moved the import line down.

I suggest enforcing this via ruff's isort rule. `ruff check --select I` currently detects 49 (automatically fixable) issues.

src/pyinfra/facts/podman.py
lsg551: Agree. Updated it.

src/pyinfra/facts/podman.py
lsg551: I could not replicate this locally, no error surfaced without the missing blank line.

Before pushing the first time, I ran `scripts/generate-docs.sh && uv run zensical serve` and inspected the docs manually. No issue whatsoever.

I updated it nonetheless for consistency.

src/pyinfra/facts/podman.py
lsg551: I tried RST first:

```
**Note**: Available since Podman
`v5.6.0 <https://github.com/containers/podman/releases/tag/v5.6.0>`_.
This version constraint is NOT checked automatically.
```

and it would not render correctly, see this screenshot:

hyperlink-rst

I am too unfamiliar with RST; am I doing something wrong? Also not working:

```
    **Note**: Available since Podman `v5.6.0`_. This version
    constraint is NOT checked automatically.

    .. _v5.6.0: https://github.com/podman-container-tools/podman/releases/tag/v5.6.0
```

The markdown variant is working though...

hyperlink-md

PS: sorry for the whitemode :D it did not correctly pick up my browser's preference

lsg551: Ohh, `scripts/generate-docs.sh` was also failing for me:

```
[...]
### Generating arguments snippet
### Building arguments doc
  File "/truncated/pyinfra/scripts/generate_arguments_doc.py", line 82, in <module>
    build_arguments_doc()
  File "/truncated/pyinfra/scripts/generate_arguments_doc.py", line 59, in build_arguments_doc
    type_args = ", ".join([arg.__name__ for arg in type_.__args__])
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/truncated/pyinfra/scripts/generate_arguments_doc.py", line 59, in <listcomp>
    type_args = ", ".join([arg.__name__ for arg in type_.__args__])
                           ^^^^^^^^^^^^
AttributeError: 'types.UnionType' object has no attribute '__name__'. Did you mean: '__ne__'?
```

Applied a subtle fix locally (not committed) so I could build the docs.

lsg551: > This is Markdown link-reference syntax, but docstrings are rendered as rst, so the link will show up as literal text. Use an rst inline link instead, e.g. `` `Podman v5.6.0 <https://github.com/containers/podman/releases/tag/v5.6.0>`_ ``.

~~Hmm, there are a lot of custom scripts handling formatting / docs generation, and I didn't want to dig too deep into the matter, but it seems like~~

- ~~mixed RST and Markdown,~~
- ~~`scripts/docs_utils.py:29 rst_to_md_docstring(text: str) -> str` renders *some* of this to Markdown, such as the code blocks, but not RST (inline) hyperlinks~~
   ```python
   def rst_to_md_docstring(text: str) -> str:
       """Best-effort RST→Markdown conversion for Python docstrings.

       Handles the directives pyinfra docstrings use in practice:
       - ``.. code:: LANG`` / ``.. code::`` → fenced code block.
       - ``.. note::``, ``.. caution::`` etc. → ``!!! note`` admonitions.
       - ``.. admonition:: Title`` → ``!!! note "Title"``.
       - Double-backtick inline literals (``` ``x`` ```) → single-backtick.
       """
       # ... implementation
   ```
- ~~Zensical, in the end, renders Markdown (?).~~

~~So my hypothesis is that the proposed RST hyperlink is actually interepreted as Markdown, and therefore rendered as inline code, as seen in the above screenshot.~~

👉 Actually, it seems like all such links are broken in the current production deployment.

Examples:

- [`operations.files.template`](https://docs.pyinfra.com/en/3.x/operations/files.html#files-template) has a broken RST hyperlink / renders it as inline code
- [`operations.uv.packages`](https://docs.pyinfra.com/en/3.x/operations/uv.html#uv-packages) ditto
- [`operations.opkg`](https://docs.pyinfra.com/en/3.x/operations/opkg.html) ditto
- and a few more

Apparently, probably all RST inline links are broken.

- I could not grep a single inline (`[]()`) or reference-syle (`[][]`) Markdown link (except mine)
- bare URLs render correctly (?) (via `pymdownx.magiclink`?), see [`facts.sysvinit`](https://docs.pyinfra.com/en/3.x/facts/sysvinit.html) 

Possibly a regression introduced in https://github.com/pyinfra-dev/pyinfra/commit/623adabeace0cb693d8520969ed0bc41924a7315 or simply not migrated to Markdown?

Fizzadar (approved):
This looks good to go, @lsg551! Thank you for the PR

Issue #785, opened 2022-04-04
Use rich


Fizzadar:
Drop the custom progress bar.

Can potentially show logs in pager, etc.

olafurnielsen:
I think having an option to enable verbose & pretty output on a per-operation/fact level would be a nice feature.

I created a little FactBase mixin for logging output in a Rich way.

I find it super useful for one-off fact runs as eye-balling some text/object in Pyinfra's output was hard and error-prone, especially for facts returning multiple lines/items.

Optionally it will display a syntax highlighted code block with information relevant to gathering the fact – e.g. an SQL query behind a fact or lines matched with FindInFile.

See some examples:
<img width="1269" alt="rich-fact" src="https://github.com/pyinfra-dev/pyinfra/assets/69799161/87fafd66-6979-4909-9c65-bef03152ac29">

<img width="1268" alt="rich-fact-2" src="https://github.com/pyinfra-dev/pyinfra/assets/69799161/9e61c6c0-381c-41e6-bc9b-2510640bbc55">

And for highlighting stdout from vital operations:
<img width="545" alt="Screenshot 2024-05-07 at 23 10 02" src="https://github.com/pyinfra-dev/pyinfra/assets/69799161/99345baf-91db-4349-9da6-84443e169dca">

Would be happy to help on this if there's interest in having something like this.

This is part of our collection of operations/facts/tasks for installing, configuring and managing Oracle Databases. Was thinking of open sourcing it once it's battle tested enough.

PR #1901, opened 2026-07-26
refactor(cli): migrate argument parsing from Click to Cyclopts

noirbizarre:
Part 1 of 3 of the CLI modernisation, split out of #1894 as requested for easier review.

This migrates command-line **argument parsing** from Click to [Cyclopts](https://cyclopts.readthedocs.io/), while keeping the existing command syntax, flags and behaviour. Output styling still goes through the pluggable `pyinfra.api.output` layer wired to Click (`set_formatter(click.style)` / `set_echo(click.echo)`), so this change is scoped to parsing only — Rich output arrives in the next PR.

- `cli.py` becomes a Cyclopts `App` with `@app.default` and typed `Annotated` parameters; `PYINFRA_YES` keeps its lenient boolean parsing (empty / on / off).
- `--support` is now a Cyclopts command; the previously duplicated `support` parameter and its unreachable handler are removed.
- `main.py` invokes the Cyclopts app and surfaces `click.ClickException` (CliError etc.) via their `.show()` / exit code.
- The confirmation prompt and all human output remain Click-based.
- pyproject: add `cyclopts`, keep `click` (now used for output only).
- Tests invoke the Cyclopts app through the updated `run_cli` helper.

---

### 📚 Stacked PR series

This is one of three PRs that split the combined CLI refactor (originally #1894) into reviewable layers. They are stacked and land in order; Parts 2 & 3 stay in **draft** until the preceding part is merged.

- #1901 — Part 1 — Click → Cyclopts argument parsing ← **this PR**
- #1902 — Part 2 — Rich-based output, prints & debug-inventory
- #1903 — Part 3 — live progress tree, TTY fixes & syntax-highlighted diffs

Original combined PR (kept for reference/discussion): #1894

Related issue: #785 (use Rich) — referenced, not closed.

---

### 🎥 Showcase

Those are here to show what visually changed and what did not (yet).

In this PR, `click` is still in charge of rendering/styling, `rich` is not yet introduced.

#### Help

help

#### Inventory
inventory current

#### Success

success current

#### Error and diff

diff

#### Multiline output

multiline


src/pyinfra_cli/cli.py
```diff
@@ -664,8 +691,8 @@ def _set_config(
 
     # Load up any config.py from the filesystem
     if state.cwd:
-        config_filename = str(Path(state.cwd) / config_filename)
-    if Path(config_filename).exists():
+        config_filename = path.join(state.cwd, config_filename)
```
Fizzadar: Reversion of #1753 changes here, let's put those back

src/pyinfra_cli/cli.py
```diff
@@ -40,258 +42,292 @@
...
+
+
+app = App(
+    name="pyinfra",
+    version=__version__,
+    version_flags=["--version"],
+    help_flags=["-h", "--help"],
+)
 
+# Enable ``pyinfra --install-completion`` for shell autocompletion.
```
Fizzadar, line 75: Much nicer! We need to update the doc accordingly in `docs/cli.md` and also remove `scripts/pyinfra-complete.[sh|zsh]`.

src/pyinfra_cli/cli.py
```diff
@@ -40,258 +42,292 @@
...
+    if value in _FALSY_BOOL_VALUES:
+        return False
+    if value in _TRUTHY_BOOL_VALUES:
+        return True
+    raise ValueError(f"invalid boolean value: {tokens[0].value!r}")
+
+
+app = App(
+    name="pyinfra",
+    version=__version__,
```
Fizzadar: Let's also set `version_format` so we match the current output in case anyone relies on that (`pyinfra: v3.9.2`).

Fizzadar (changes requested):
Couple of tweaks then we're good to go I think!

src/pyinfra_cli/cli.py
noirbizarre: Done — restored the pathlib forms to match #1753: `config_filename = str(Path(state.cwd) / config_filename)` and `Path(config_filename).exists()`. I'd also accidentally reverted #1753 in `_validate_operations` (`path.exists`/`path.relpath`/`path.join`), so those are back to `Path(...).exists()` / `os.path.relpath` / `str(Path(chdir) / filename)` too, and the `from os import ... path` import is dropped.

src/pyinfra_cli/cli.py
noirbizarre: Done — removed `scripts/pyinfra-complete.{sh,zsh}` and rewrote the *Shell Autocompletion* section of `docs/cli.md` to document the new `pyinfra --install-completion` command (auto-detects the shell, supports `bash`/`zsh`/`fish`, with `--shell`/`--output` overrides).
(that part was forgotten from the original PR)

src/pyinfra_cli/cli.py
noirbizarre: Fixed, `pyinfra --version` now prints `pyinfra: v3.9.2` again.

> [!NOTE]
> Cyclopts' `version_format` isn't a message template — it only selects the *markup* style (`plaintext`/`markdown`/`rst`), so it can't add the `pyinfra: v` prefix. The prefix is part of the version *string*, so I set it there instead: `version=f"pyinfra: v{__version__}"`.

noirbizarre:
All comments addressed. 👍🏼 

The version is now iso to the current format:
image

(Good to know for #1753, I love pathlib!)

wowi42 (approved):
Re-reviewed after the requested changes. Approving.

Verified locally on this branch (Python 3.13):

- `pyinfra --version` prints `pyinfra: v...` again (matches the old Click output).
- `pyinfra --support` works and exits 0.
- `-vvv` bundling still works for verbosity.
- Missing operations exits 1 with a clear usage message (Click exited 2 with a usage error; the new message is more helpful, fine by me).
- Unknown option exits 1 (was 2 under Click; acceptable).
- `PYINFRA_YES=off` lenient parsing works.
- Full unit suite passes (1951 passed), ruff and mypy clean.

Two cosmetic Cyclopts differences in `--help`, not blockers:

- Flags now show `[default: False]` and repeatable options show `[default: ()]`, which adds noise Click did not have. If Cyclopts has a flag to hide defaults for these, worth using in Part 2 or 3.
- The usage line reads `pyinfra COMMAND [OPTIONS] INVENTORY [ARGS...]`; the `COMMAND` token is a bit misleading for the default command.

The `scripts/pyinfra-complete.{sh,zsh}` removal is a breaking change for anyone sourcing those files, but it was requested in the earlier review and `--install-completion` (which also adds fish support) is a good replacement.

PR #1904, opened 2026-07-27
fix(cli): avoid local path validation for exec args

### LINKED ISSUE #1883: bug: exec verifies access to path-like arguments that exist locally as if they were task files
## Describe the bug
If I run an `exec` command that touches a filesystem path that in addition to potentially existing remotely, also exists locally, exec refuses to run if my local user does not have access to the local path.

This is a faulty check given that the local path and it's local permissions for the local user have nothing to do with whether the remote exec will succeed or not. My best guess is that any path-like-object on the cli is assumed to be a local task file, even with `exec` that does not take a local task file. If the local file doesn't exist, then the validation is ignored, but if it exists and is inaccessible, validation fails.

While this may seem like a somewhat niche problem, it becomes a constant issue if running pyinfra from a server that is part of the inventory set. Touching with exec cluster wide any path requiring root access becomes impossible, since it obviously exists "locally" but for good reason is not accessible to the current unprivileged user.

## To Reproduce
As a normal non-root user in any pyinfra context:
```shell
%> sudo touch /tmp/immovable.obj
%> sudo chmod go-rwx /tmp/immovable.obj
%> pyinfra inventory.py exec ls /tmp/immovable.obj
Usage: pyinfra [OPTIONS] INVENTORY OPERATIONS...
Try 'pyinfra --help' for help.

Error: Invalid value for 'OPERATIONS...': Path '/tmp/immovable.obj' is not readable.
%> pyinfra inventory.py exec ls /tmp/unstoppable.obj
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [demo]            Connected

--> Preparing exec operation...
    [demo]            Ready: shell

--> Starting operation: server.shell (ls /tmp/unstoppable.obj)
[demo]            ls: cannot access '/tmp/unstoppable.obj': No such file or directory
    [demo]            Error: executed 0 commands

```

Performing the `exec` instead as `exec -- ls /tmp/immovable.obj` does not change the issue, the path is still validated.

+ Operation code & usage
N/A

+ Target system information
N/A
N/A

## Expected behavior
Path checking locally with `exec` has no plausible use-case and it should be skipped when running `exec` commands.

## Meta
<details>

```
    System: Linux
      Platform: Linux-7.0.13+deb14-amd64-x86_64-with-glibc2.42
      Release: 7.0.13+deb14-amd64
      Machine: x86_64
    pyinfra: v3.9.2
      click: v8.4.2
      distro: v1.9.0
      gevent: v26.5.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.16.0
    Executable: /home/jinna/.local/share/mise/installs/pipx-pyinfra/latest/bin/pyinfra
    Python: 3.14.6 (CPython, Clang 22.1.3 )
```

</details>

   - pipx via mise
   - N/A
   - N/A, the validation triggers before either of these options is used and the output does not differ.

0xTaoZ:
## Summary

- stop Click from treating variadic operation arguments as local filesystem paths
- add a CLI regression test for `exec` with an unreadable local path argument

## Rationale

`pyinfra INVENTORY exec ...` sends the remaining arguments to the remote shell command. Those arguments may look like local paths, but local readability is irrelevant for `exec`. The existing `click.Path(exists=False)` type still rejected an argument when that path existed locally but was unreadable, before pyinfra could prepare the exec operation.

Deploy-file validation is still handled later in `_validate_operations`, so deploy files keep the existing explicit pyinfra checks.

Fixes pyinfra-dev/pyinfra#1883.

## Test plan

- `uv run pytest tests/test_cli/test_cli.py::TestExecCli::test_exec_command_does_not_validate_local_path_arguments -q`
- `uv run pytest tests/test_cli/test_cli.py::TestExecCli tests/test_cli/test_cli.py::TestOperationCli -q`
- `uv run ruff format --check src/pyinfra_cli/cli.py tests/test_cli/test_cli.py`
- `uv run ruff check src/pyinfra_cli/cli.py tests/test_cli/test_cli.py`
- `uv run mypy src/pyinfra_cli/cli.py`
- `git diff --check`

### DIFF
--- a/src/pyinfra_cli/cli.py
+++ b/src/pyinfra_cli/cli.py
@@ -61,7 +61,7 @@ CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
 @click.command(context_settings=CONTEXT_SETTINGS)
 @click.argument("inventory", nargs=1, type=click.Path(exists=False))
-@click.argument("operations", nargs=-1, required=True, type=click.Path(exists=False))
+@click.argument("operations", nargs=-1, required=True)
 @click.option(
     "verbosity",
     "-v",
--- a/tests/test_cli/test_cli.py
+++ b/tests/test_cli/test_cli.py
@@ -1,5 +1,7 @@
 import json
+import os
 from os import path
+import tempfile
 from unittest import TestCase
 
 from pyinfra.api import Host, Inventory
@@ -150,6 +152,23 @@ class TestExecCli(PatchSSHTestCase):
             "--",
             "echo hi",
         )
+        assert result.exit_code == 0, result.stderr
+
+    def test_exec_command_does_not_validate_local_path_arguments(self):
+        with tempfile.NamedTemporaryFile() as file:
+            os.chmod(file.name, 0o000)
+            try:
+                result = run_cli(
+                    path.join("tests", "test_cli", "deploy", "inventories", "inventory.py"),
+                    "exec",
+                    "--debug-operations",
+                    "--",
+                    "ls",
+                    file.name,
+                )
+            finally:
+                os.chmod(file.name, 0o600)
+
         assert result.exit_code == 0, result.stderr
 
 


wowi42 (commented):
Reviewed. This duplicates #1886, which fixes the same issue (#1883) with the same one-line change (dropping the `click.Path` type from the `operations` argument) plus an explicit `os.access(filename, R_OK)` check in `_validate_operations` so unreadable deploy files still produce a clean `CliError` instead of a raw `PermissionError` traceback at load time.

Since #1886 is the superset, I suggest we merge that one and close this. Still, thanks for the PR, and the regression test here is good.

One note if this one were to move forward instead: removing the Path type also removes all upfront validation for actual deploy files, so the `os.access` compensation from #1886 would need to come along.

PR #1912, opened 2026-08-04
fix(facts.files): read checksum targets via stdin to fix backslash escaping (#1911)

### LINKED ISSUE #1911: files.put is never idempotent when dest contains a backslash (systemd \x2d paths)
### Describe the bug

`files.put` / `files.template` are never idempotent when `dest` contains a
backslash. This is the usual case for systemd unit files, where `-` in a path
is escaped as `\x2d` (e.g. `mnt-data\x2dbig.mount`), so every run re-uploads
the file and reports a change.

As far as I can see the root cause is `Sha256File` fact parsing. GNU coreutils
`sha256sum` escapes filenames containing `\` (or a newline) and prefixes the
whole output line with a literal `\`:

```bash
touch '/tmp/mnt-data\x2dbig.mount'
sha256sum '/tmp/mnt-data\x2dbig.mount'
\e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /tmp/mnt-data\\x2dbig.mount
```

`pyinfra.facts.files.Sha256File.process()` appears to match the hash with a
regex anchored on the requested path, so neither the leading `\` nor the
doubled `\\` in the name match. The fact resolves to `None`, `files.put`
assumes the remote file is missing and uploads unconditionally.

The same likely applies to `Md5File`, `Sha1File` and `Sha384File`, and
therefore to `files.sync` and `files.download(sha256sum=...)`.

### To Reproduce

```python
from io import StringIO

files.put(
    name="systemd mount unit",
    src=StringIO("[Mount]\nWhat=/dev/sdb1\nWhere=/mnt/data-big\n"),
    dest=r"/etc/systemd/system/mnt-data\x2dbig.mount",
)
```

Run twice.

### Expected behavior

Second run is a no-op.

### Observed behavior

Every run uploads the file and is reported as changed. With `-vv` the fact is
loaded as `None` even though the file exists on the remote host.

### Possible fix

Avoid passing the filename to the checksum tool at all, so that no escaping
can occur:

```bash
{ sha256sum < '<path>'; } 2>/dev/null    # -> "<hash>  -" or empty string if <path> doesn't exist
```

Alternatively, detect the leading `\` in the output and unescape `\\` and `\n`
in the filename before matching, per coreutils' documented behaviour.

### Meta

- pyinfra: 3.9.2
- Target OS: Ubuntu 24.04 LTS
- Installed via pip

wowi42:
GNU coreutils' `sha256sum` / `md5sum` / etc escape filenames containing backslashes (and newlines), prefixing the output line with a literal \. This broke `HashFileFactBase.process()`, which matched the hash against the unescaped path, so `files.put` / `files.template` / `files.sync` / `files.download` with such paths were never idempotent.

Fix by feeding the file to each checksum tool via stdin so the filename never appears in the output. Update the parser to handle both the GNU/shasum stdin style (`"<hash>  -"`) and the BSD digest style (`"<hash>"`).

Fixes #1911.


Fizzadar (approved):
Neat fix @wowi42!

PR #1915, opened 2026-08-07
feat(operations.files): Ensure file ops don't fail if source missing in prepare phase

aktiur:
Updated the move, copy and unarchive operations so that they won't raise an operation error when the source file (or target directory) is missing while in the prepare phase. Instead, they simply log an info message.

During the execution phase, the operations raise errors as before.

This commit is based on previous work by @vaz-ar in PR #1787.

All tests pass. However, the `FakeState` used by `pyinfra-testing` always has `is_executing` set to false. This means only the prepare phase code path is actually tested. On a similar note, I haven't seen a way to assert what is logged during the operation. How should I proceed to make the tests more complete?

### DIFF
--- a/src/pyinfra/operations/files.py
+++ b/src/pyinfra/operations/files.py
@@ -18,7 +18,6 @@ from typing import IO, Any
 from jinja2 import TemplateRuntimeError, TemplateSyntaxError, UndefinedError
 
 from pyinfra import host, logger, state
-from pyinfra.api.output import format_text
 from pyinfra.api import (
     FileDownloadCommand,
     FileUploadCommand,
@@ -31,6 +30,7 @@ from pyinfra.api import (
     operation,
 )
 from pyinfra.api.command import make_formatted_string_command
+from pyinfra.api.output import format_text
 from pyinfra.api.util import (
     get_call_location,
     get_file_io,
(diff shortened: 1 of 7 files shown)


wowi42 (approved):
LGTM

The core logic is solid. During the prepare phase, `move`, `copy`, and `unarchive` now log an info message and proceed when the source or destination directory doesn't exist yet, rather than raising `OperationError`. The execution phase still raises errors as before. The `False` vs `None` distinction in the fact return values is handled correctly: `False` (path exists but is wrong type) always raises, `None` (doesn't exist yet) is lenient during prepare.

Two small things:

First, the `host.log` → `logger.info` change in the `line` function and the import reordering of `format_text` are unrelated to this PR. Functionally identical, just a style change. Please split these out into a separate PR.

Second, `move/invalid_dest.json` was changed from testing `null` (dest doesn't exist) to `false` (dest is not a directory). Both are useful, but the "doesn't exist" prepare-phase path for `move`'s dest is now untested. A follow-up fixture would close that gap — the `copy` equivalent is already covered by `copy/invalid_dest.json`.

On execution-phase testing: you're right that `FakeState` always has `is_executing=False`. The `is_executing` branches are simple enough to verify by inspection, so this isn't a blocker. If you want to add coverage later, a small patch to `pyinfra-testing` to allow setting `is_executing` on the fake state would be the cleanest approach.

aktiur:
Thanks for your review! I updated the PR to address your two remarks:

> First, the `host.log` → `logger.info` change in the `line` function and the import reordering of `format_text` are unrelated to this PR. Functionally identical, just a style change. Please split these out into a separate PR.

My mistake. I originally used `host.log` in the prepare phase code path of the operations I modified, but realised I could not, because `log` is not defined on the `FakeHost` used during tests. I was not careful and/or too zealous when I did some find/replace to handle that. It's gone from the PR.

> Second, `move/invalid_dest.json` was changed from testing `null` (dest doesn't exist) to `false` (dest is not a directory). Both are useful, but the "doesn't exist" prepare-phase path for `move`'s dest is now untested. A follow-up fixture would close that gap — the `copy` equivalent is already covered by `copy/invalid_dest.json`.

I created two additional tests for both `files.copy` and `files.move` named `missing_dest` that cover the case where the destination directory is missing. The original tests (named `invalid_test`) both cover the case where the destination is not a directory.

> On execution-phase testing: you're right that `FakeState` always has `is_executing=False`. The `is_executing` branches are simple enough to verify by inspection, so this isn't a blocker. If you want to add coverage later, a small patch to `pyinfra-testing` to allow setting `is_executing` on the fake state would be the cleanest approach.

I'm going to look into that.

Fizzadar (approved):
This is great, thank you @aktiur!

PR #1918, opened 2026-08-08
docs: fix links to connector and fact guides

### LINKED ISSUE #1889: Documentation contains a few dead links
The docs over at https://docs.pyinfra.com contain a few dead links:

- https://docs.pyinfra.com/en/3.x/connectors.html -> "Check out the writing connectors guide" -> 404
- https://docs.pyinfra.com/en/3.x/facts.html -> "You can leverage facts within operations like this:" -> 404
- https://docs.pyinfra.com/en/3.x/facts.html -> "Check out the writing facts guide" -> 404

There might be a few more.

I tried to fix these myself, but got stuck since I could not reproduce the 404 errors locally (since the URL paths are different) and was unsure whether my fixes would work when deploying to docs.pyinfra.com...

Himanshuagrawal4:
## Summary

- fix the connector overview link to the writing-connectors guide
- fix the facts overview links to operations and the writing-facts guide

Closes #1889

## Validation

- generated the documentation site successfully
- verified all seven local link targets in the modified pages

## AI disclosure

OpenAI Codex was used to help investigate the broken links, prepare the patch, and run validation. I manually reviewed and approved the final diff before committing it.

### DIFF
--- a/docs/connectors.md
+++ b/docs/connectors.md
@@ -8,7 +8,7 @@ Connectors enable pyinfra to integrate with other tools out of the box. Connecto
 
 Each connector page is listed below and contains examples as well as a list of available data that can be used to configure the connector.
 
-**Want a new connector?** Check out [the writing connectors guide](../api/connectors.md).
+**Want a new connector?** Check out [the writing connectors guide](api/connectors.md).
 
 ### Popular connectors
 
--- a/docs/facts.md
+++ b/docs/facts.md
@@ -22,7 +22,7 @@ Multiple facts with arguments may be called like so:
 pyinfra @local fact files.File path=setup.py files.File path=anotherfile.txt
 ```
 
-You can leverage facts within [operations](../using-operations.md) like this:
+You can leverage facts within [operations](using-operations.md) like this:
 
 ```python
 from pyinfra import host
@@ -32,6 +32,6 @@ if host.get_fact(LinuxName) == 'Ubuntu':
     apt.packages(...)
 ```
 
-**Want a new fact?** Check out [the writing facts guide](../api/facts.md).
+**Want a new fact?** Check out [the writing facts guide](api/facts.md).
 
 --8<-- "facts-cards.html"


wowi42 (approved):
LGTM. The `../` prefix was wrong for pages at the docs root: `../api/connectors.md` escapes the docs source tree, so the built page at `/en/3.x/connectors.html` linked to `/en/api/connectors.html`, hence the 404s in #1889.

The new links match the convention already used by the other root-level pages (`docs/index.md`, `docs/operations.md`, `docs/contributing.md`), and I confirmed these three were the last remaining `../` links from root-level pages, so this fixes the whole class of bug, not just the reported ones.

AI disclosure present and compliant with AI_POLICY.md.

Note for the record: I could not complete a local `zensical build` to double-check, because `scripts/generate_arguments_doc.py` crashes on Python 3.13 (`types.UnionType` has no `__name__`). That is an unrelated pre-existing issue; the docs-build CI on this PR passes.

simonhammes:
Hmm, I'm not sure if this PR actually fixed the issue.

I'm assuming that the changes have already been deployed to docs.pyinfra.com via https://github.com/pyinfra-dev/docs.pyinfra.com/commit/84a797bc7b1d5573dfc47a24e5225415bee64f02 (?)

Clicking on the "Check out the writing connectors guide" link on https://docs.pyinfra.com/en/3.x/connectors.html still yields a 404

wowi42:
@simonhammes The fix in this PR does work. It is live on https://docs.pyinfra.com/en/latest/connectors.html and https://docs.pyinfra.com/en/next/connectors.html (both now link to `api/connectors.html` correctly).

The `/en/3.x` pages are versioned docs and are only rebuilt from version tags (see `scripts/build-public-docs.sh`), while branch builds only refresh `en/next` and `en/latest`. So `/en/3.x/connectors.html` will keep the broken link until the next v3.x release is tagged.

simonhammes:
> The `/en/3.x` pages are versioned docs and are only rebuilt from version tags (see `scripts/build-public-docs.sh`), while branch builds only refresh `en/next` and `en/latest`.

Ah, that makes sense! Thanks for the explanation.

PR #1855, opened 2026-06-28
fix(connectors.ssh): respect ssh_key_password over ssh_config IdentityFile

wowi42:
Fixes #1852.

## Problem

The ssh_config certificate auto-load added in #1750 (shipped in 3.9.0) loads the ssh_config `IdentityFile` eagerly during `parse_config`, but without a passphrase. With an encrypted key this triggers a `getpass` prompt even when the host already sets `ssh_key` and `ssh_key_password`, so the configured password is ignored:

```
Enter password for private key: /home/user/.ssh/foo.example.com:
```

It only triggers when the user's `~/.ssh/config` has an `IdentityFile` matching the host and that key is encrypted.

## Fix

1. An explicit pyinfra `ssh_key`, already loaded into a `pkey` using `ssh_key_password`, now wins over any ssh_config `IdentityFile`, the same way `ssh -i` overrides config. The config identity is no longer loaded when a `pkey` is already present, so no prompt fires.
2. The config identity load is now non-interactive (`allow_prompt=False`). An encrypted identity with no known passphrase falls through to the legacy `key_filename` flow instead of blocking on a prompt.

## Tests

- New `ssh_encrypted_key` fixture (real passphrase-protected key via `ssh-keygen`).
- `test_parse_config_explicit_pkey_wins_over_encrypted_identityfile`: explicit pkey preserved, no prompt.
- `test_parse_config_encrypted_identityfile_does_not_prompt`: encrypted config identity falls through silently, no prompt.

Full non-e2e suite passes (1857), ruff/format/mypy clean.

PR #1921, opened 2026-08-15
fix(connectors.ssh): prompt for encrypted ssh_config IdentityFile passphrase (#1917)

### LINKED ISSUE #1852: Version 3.9.0 through 3.9.2 doesn't use `ssh_key_password`
## Describe the bug

Upgrading to Pyinfra 3.9.0 from 3.8.0 breaks `ssh_key_password` in inventory files. Instead of decrypting and using the SSH key with `ssh_key_password`, Pyinfra prompts for a password to decrypt the key. Versions 3.9.1 and 3.9.2 don't fix this.

## To Reproduce

Create an inventory file like the following:

```python
import subprocess as sp
from pathlib import Path

host = "foo.example.com"
# pago is my CLI password manager similar to `pass`.
# You could use a literal password.
password = sp.check_output(["pago", "show", "ssh-foo"], text=True).strip()

inventory = [
    (
        host,
        {
            "ssh_key": str(Path.home() / ".ssh" / host),
            "ssh_key_password": password,
        },
    ),
]
```

Try it with Pyinfra 3.8.0 and 3.9.0:

```none
> uv tool install pyinfra@v3.8.0
> pyinfra inventory.py deploy.py
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [foo.example.com] Connected

    Loading: deploy.py
...
> uv tool install pyinfra@v3.9.0
> pyinfra inventory.py deploy.py
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
Enter password for private key: /home/dbohdan/.ssh/foo.example.com:
```

The password is still used at some point. Assigning `password = "asdf"` instead of the correct password results in an error:

```none
  File "src/gevent/greenlet.py", line 912, in gevent._gevent_cgreenlet.Greenlet.run
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/api/host.py", line 390, in connect
    self.connector.connect()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/connectors/ssh.py", line 224, in connect
    return self._connect()
           ~~~~~~~~~~~~~^^
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/connectors/ssh.py", line 247, in _connect
    kwargs = self.make_paramiko_kwargs()
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/connectors/ssh.py", line 203, in make_paramiko_kwargs
    kwargs["pkey"] = get_private_key(
                     ~~~~~~~~~~~~~~~^
        self.state,
        ^^^^^^^^^^^
        key_filename=ssh_key,
        ^^^^^^^^^^^^^^^^^^^^^
        key_password=self.data["ssh_key_password"],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/connectors/ssh_util.py", line 137, in get_private_key
    key = load_key_with_certificate(
        key_filename=key_filename,
        key_password=key_password,
        cwd=state.cwd,
    )
  File "/home/dbohdan/.local/share/uv/tools/pyinfra/lib/python3.13/site-packages/pyinfra/connectors/ssh_util.py", line 112, in load_key_with_certificate
    raise PyinfraError(f"Invalid private key file: {key_filename}")
pyinfra.api.exceptions.PyinfraError: Invalid private key file: /home/dbohdan/.ssh/foo.example.com
2026-06-23T15:27:15Z <Greenlet at 0x7823655f6340: <bound method Host.connect of Host(foo.example.com)>> failed with PyinfraError
```

## Expected behavior

Pyinfra using `ssh_key_password` and not showing a password prompt.

## Meta

```none
> pyinfra --support

    System: Linux
      Platform: Linux-6.8.0-124-generic-x86_64-with-glibc2.39
      Release: 6.8.0-124-generic
      Machine: x86_64
    pyinfra: v3.9.2
      click: v8.4.1
      distro: v1.9.0
      gevent: v26.5.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.15.0
    Executable: /home/dbohdan/.local/bin/pyinfra
    Python: 3.13.9 (CPython, Clang 21.1.4 )

```

Context: I started using `ssh_key_password` because Pyinfra didn't use the ssh-agent password with a jumphost. This is an upstream issue with Paramiko.

### LINKED ISSUE #1917: Regression:  pyinfra 3.10.0 no longer prompts for passphrases of encrypted SSH private keys defined in `IdentityFile`
## Describe the bug
This is a regression in pyinfra 3.10.0 compared with pyinfra 3.9.2.

When the SSH connector uses an encrypted private key supplied through an SSH config `IdentityFile`, pyinfra 3.9.2 prompts interactively for the private-key passphrase. In pyinfra 3.10.0, the passphrase prompt is not shown and authentication fails with "Authentication error () (Private key file is encrypted)".
## To Reproduce
You can use the script I provided to reproduce this bug.
```
mkdir pyinfra-bug-test
cd pyinfra-bug-test
wget https://github.com/user-attachments/files/30856557/setup.sh
# it's a simple script, you can verify it easily before executing it.
bash setup.sh

# enter password 123, success
source .venv-3.9.2/bin/activate
pyinfra -vv -y inventory.py deploy.py
deactivate

# fail
# get "Authentication error () (Private key file is encrypted)"
source .venv-3.10.0/bin/activate
pyinfra -vv -y inventory.py deploy.py
deactivate

docker compose down -v --remove-orphans



```
[setup.sh](https://github.com/user-attachments/files/30856557/setup.sh)
The script:
1. Creates `.venv-3.9.2` and `.venv-3.10.0`.
2. Installs the corresponding pyinfra versions from PyPI.
3. Generates an encrypted Ed25519 key pair with passphrase `123`.
4. Creates an SSH config using that key as `IdentityFile`.
5. Starts an OpenSSH server container.
6. Mounts `authorized_keys` into the server container.
7. Prints the exact test commands.

The operation is:

```python

server.shell(commands=["echo connected-from-pyinfra"])
```

The inventory is:

```python
hosts = [
    ("pyinfra-repro", {"ssh_config_file": "ssh_config", "ssh_strict_host_key_checking": "no"}}),
]
```
The SSH config is:

```sshconfig
Host pyinfra-repro
    HostName 127.0.0.1
    Port 2222
    User debug
    IdentityFile ./keys/id_ed25519
    IdentitiesOnly yes
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
```

Run the 3.9.2 test:

```bash
source .venv-3.9.2/bin/activate
pyinfra -vv -y inventory.py deploy.py
deactivate
```

Run the 3.10.0 test:

```bash
source .venv-3.10.0/bin/activate
pyinfra -vv -y inventory.py deploy.py
deactivate
```

The private-key passphrase is:

```text
123
```

## Expected behavior

The 3.10.0 SSH connector should preserve the 3.9.2 behavior and prompt for the encrypted `IdentityFile` passphrase when running interactively.

Alternatively, if prompting is intentionally disabled for SSH-config identities, pyinfra should produce a clear actionable error explaining that the private key is encrypted and that `ssh_key_password` must be supplied. It should not silently skip the identity or fail with only a generic authentication error.

A possible explicit interface would be a dedicated option such as:

```text
--ssh-key-password-prompt
```

## Meta
This bug is likely related to #1855 

### pyinfra --support
    System: Linux
      Platform: Linux-7.0.0-29-generic-x86_64-with-glibc2.43
      Release: 7.0.0-29-generic
      Machine: x86_64
    pyinfra: v3.10.0
      click: v8.4.2
      distro: v1.9.0
      gevent: v26.7.0
      jinja2: v3.1.6
      packaging: v26.3
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.6.0
      typing-extensions: v4.16.0
    Executable: /home/xubuntu-test/pyinfra-debug/test/.venv-3.10.0/bin/pyinfra
    Python: 3.14.4 (CPython, GCC 15.2.0)

NoiceHax:
Closes #1917

## Description

Since 3.10.0 pyinfra no longer asks for the passphrase of an encrypted private key that comes from an `IdentityFile` in ssh_config. 3.9.2 prompted for it, 3.10.0 just fails with "Authentication error () (Private key file is encrypted)".

The cause is #1855. That change did two things to fix #1852: it added a `"pkey" not in cfg` guard in `parse_config` so an explicit pyinfra `ssh_key` wins over the ssh_config `IdentityFile`, and it passed `allow_prompt=False` to `load_key_with_certificate` in `_attach_identity_with_certificate`. The guard on its own is what fixes #1852, because with an explicit `ssh_key` the config identity is never loaded and so can never prompt. The `allow_prompt=False` was not needed for that, and it is what broke everyone else: any encrypted `IdentityFile` now falls through to the old `key_filename` path without asking for the passphrase, and paramiko cannot do anything useful with it. I checked the 3.9.2 source and `_attach_identity_with_certificate` was already there, just without that argument.

So I dropped `allow_prompt=False` from that one call. The prompt inside `_load_private_key_file` is already gated on `pyinfra.is_cli`, so API mode keeps raising the "set ssh_key_password" error instead of blocking on stdin. The `allow_prompt` keyword stays on both helpers with its default of `True`, and I fixed its docstring, which claimed the ssh_config path had to be non-interactive because of #1852. On the test side, `test_parse_config_explicit_pkey_wins_over_encrypted_identityfile` is untouched and still passes, which is the proof that #1852 stays fixed. I turned `test_parse_config_encrypted_identityfile_does_not_prompt` around into a regression test for this issue, and added one more checking that a wrong passphrase still falls back to `key_filename` rather than raising out of `parse_config`.

One case I left alone on purpose: if `ssh_password` is set and ssh_config also names an encrypted `IdentityFile`, you get asked for a key passphrase you do not need. That was true in 3.9.2 as well, so it is not part of this regression and felt better as a separate change.

- [x] Pull request includes documentation for any new/updated operations/facts (no operation or fact changed, the two affected docstrings are updated)

## AI usage disclosure

Per the project's [AI policy](../blob/3.x/AI_POLICY.md): I wrote this with AI assistance (Claude Code), which drafted the one line source change, the docstring edits and the two tests. I have reviewed all of it and I can explain how `parse_config`, `_attach_identity_with_certificate` and `_load_private_key_file` fit together, and why removing `allow_prompt=False` does not undo the #1852 fix.

### DIFF
--- a/src/pyinfra/connectors/ssh_util.py
+++ b/src/pyinfra/connectors/ssh_util.py
@@ -133,8 +133,8 @@ def load_key_with_certificate(
       ``<key>-cert.pub`` is used if present, falling back to ``<key>.pub``.
     + cwd: optional working directory used to resolve relative key paths.
     + allow_prompt: when False, an encrypted key with no known passphrase raises
-      instead of prompting. Used by non-interactive callers like the ssh_config
-      identity path so they fall through rather than block (issue #1852).
+      instead of prompting. Only for callers that must never block on input;
+      prompting is already limited to CLI runs, so API mode always raises.
     """
 
     resolved_path: str | None = None
--- a/src/pyinfra/connectors/sshuserclient/client.py
+++ b/src/pyinfra/connectors/sshuserclient/client.py
@@ -104,9 +104,9 @@ def _attach_identity_with_certificate(cfg: dict, host_config: dict) -> None:
     falls back to the implicit ``<key>-cert.pub`` lookup so OpenSSH-style CA
     auth works without explicit pyinfra ``ssh_key`` configuration (issue #1569).
     Silent fall-through when no identity exists keeps the legacy paramiko
-    ``key_filename`` flow for missing or otherwise unloadable files. The load is
-    non-interactive (``allow_prompt=False``): an encrypted identity with no known
-    passphrase falls through instead of blocking on a prompt (issue #1852).
+    ``key_filename`` flow for missing or otherwise unloadable files. An encrypted
+    identity prompts for its passphrase under the CLI and raises in API mode,
+    both handled inside :func:`load_key_with_certificate` (issue #1917).
     """
 
     identity_files = host_config.get("identityfile") or []
@@ -126,7 +126,6 @@ def _attach_identity_with_certificate(cfg: dict, host_config: dict) -> None:
             cfg["pkey"] = load_key_with_certificate(
                 key_filename=identity_file,
                 certificate_filename=certificate_filename,
-                allow_prompt=False,
             )
         except (PyinfraError, SSHException, OSError) as e:
             logger.debug("Could not load identity %s with certificate: %s", identity_file, e)
--- a/tests/test_connectors/test_sshuserclient.py
+++ b/tests/test_connectors/test_sshuserclient.py
@@ -554,12 +554,11 @@ def test_parse_config_explicit_pkey_wins_over_encrypted_identityfile(
     assert cfg["pkey"] is sentinel
 
 
-def test_parse_config_encrypted_identityfile_does_not_prompt(
+def test_parse_config_encrypted_identityfile_prompts_in_cli(
     ssh_encrypted_key, tmp_path, _clear_ssh_config_cache
 ):
-    # Regression for #1852: an encrypted ssh_config IdentityFile with no known
-    # passphrase must fall through to the legacy key_filename flow instead of
-    # blocking on an interactive passphrase prompt.
+    # Regression for #1917: under the CLI an encrypted ssh_config IdentityFile
+    # must prompt for its passphrase, the way it did before 3.10.0.
     config_path = tmp_path / "ssh_config"
     _write_ssh_config(
         config_path,
@@ -568,11 +567,35 @@ def test_parse_config_encrypted_identityfile_does_not_prompt(
     )
 
     client = SSHClient()
-    with patch("pyinfra.connectors.ssh_util.getpass", return_value="wrong") as fake_getpass:
+    with patch(
+        "pyinfra.connectors.ssh_util.getpass",
+        return_value=ssh_encrypted_key["passphrase"],
+    ) as fake_getpass:
         with patch("pyinfra.is_cli", True):
             _, cfg, *_ = client.parse_config("myhost", ssh_config_file=str(config_path))
 
-    fake_getpass.assert_not_called()
+    fake_getpass.assert_called_once()
+    assert isinstance(cfg["pkey"], PKey)
+    assert "key_filename" not in cfg
+
+
+def test_parse_config_encrypted_identityfile_wrong_passphrase_falls_through(
+    ssh_encrypted_key, tmp_path, _clear_ssh_config_cache
+):
+    # A wrong passphrase must not blow up parse_config: fall back to the legacy
+    # key_filename flow and let paramiko report the failure.
+    config_path = tmp_path / "ssh_config"
+    _write_ssh_config(
+        config_path,
+        host="myhost",
+        IdentityFile=str(ssh_encrypted_key["key"]),
+    )
+
+    client = SSHClient()
+    with patch("pyinfra.connectors.ssh_util.getpass", return_value="wrong"):
+        with patch("pyinfra.is_cli", True):
+            _, cfg, *_ = client.parse_config("myhost", ssh_config_file=str(config_path))
+
     assert "pkey" not in cfg
     assert cfg["key_filename"] == [str(ssh_encrypted_key["key"])]
 


wowi42 (approved):
LGTM

This restores the passphrase prompt for encrypted ssh_config IdentityFile keys that regressed in 3.10.0. The analysis is correct: the `"pkey" not in cfg` guard in `parse_config` (line 329) already prevents `_attach_identity_with_certificate` from being called when an explicit `ssh_key` is set, so `allow_prompt=False` was redundant for #1852. Removing it lets `_load_private_key_file` prompt in CLI mode (gated on `pyinfra.is_cli`) and raise `PyinfraError` in API mode, which the existing except clause catches and falls through to `key_filename`. The `test_parse_config_explicit_pkey_wins_over_encrypted_identityfile` test is untouched and still passes, confirming #1852 stays fixed. The renamed test now asserts the prompt fires and the key loads, and the new wrong-passphrase test covers the fallback path. Both docstrings are updated to match the new behavior.

PR #1922, opened 2026-08-15
fix(connectors.ssh): treat ProxyJump none as no jump host (#1445)

### LINKED ISSUE #1445: Fails when ssh_config for host contains `ProxyJump none`
From [ssh_config manual](https://man.openbsd.org/ssh_config#ProxyJump):

> Setting the host to none disables this option entirely.

However pyinfra will attempt to proxy-jump through the literal hostname "none" and fail on the name resolution step.

The underlying paramiko SSHConfig library has custom handling for `ProxyCommand none` but not `ProxyJump none`. It might make sense for pyinfra to handle it itself because it already has a whole bunch of customization on top of paramiko.

There is a paramiko bug open for this: https://github.com/paramiko/paramiko/issues/2285

NoiceHax:
Closes #1445

## Description

OpenSSH reads `ProxyJump none` as switching the option off. pyinfra instead split the value on commas and used the result as a list of hops, so a host with `ProxyJump none` in its ssh_config made pyinfra try to connect to a machine literally called "none" and fail on name resolution. Paramiko has a special case for `ProxyCommand none` but not for this one, so it needs handling here.

The fix skips the jump when the value is `none`, compared without case. `sock` is then never set and pyinfra connects straight to the target, which is what ssh does. The check is against the whole value on purpose, since OpenSSH only accepts `none` on its own and never as one entry in a comma separated hop list. It reads the same way as the `IdentityAgent none` check a few lines above.

New test `test_load_ssh_config_proxyjump_none` covers both `none` and `NONE`, and asserts no hop connection is made and no `sock` ends up in the config. It fails before the change and passes after.

## Checklist

- [x] Pull request includes documentation for any new/updated operations/facts (no new operations or facts, this only changes ssh_config handling)

## AI usage disclosure

Per the project's [AI policy](../blob/3.x/AI_POLICY.md): I wrote this with AI assistance (Claude Code), which drafted the two line source change and the new test, and I have read the diff and understand how `parse_config` builds the ProxyJump chain and why skipping it leaves a plain direct connection.

### DIFF
--- a/src/pyinfra/connectors/sshuserclient/client.py
+++ b/src/pyinfra/connectors/sshuserclient/client.py
@@ -347,11 +347,15 @@ class SSHClient(ParamikoClient):
             if agent_path.lower() != "none":
                 identity_agent = os.path.expanduser(agent_path)
 
+        proxyjump = host_config.get("proxyjump")
+
         if "proxycommand" in host_config:
             cfg["sock"] = ProxyCommand(host_config["proxycommand"])
 
-        elif "proxyjump" in host_config:
-            hops = host_config["proxyjump"].split(",")
+        # ``ProxyJump none`` disables jumping entirely (issue #1445), the same as
+        # OpenSSH - without this we'd jump via a host literally called "none".
+        elif proxyjump and proxyjump.strip().lower() != "none":
+            hops = proxyjump.split(",")
             sock = None
             # Propagate the target's timeout down so hop connections and the
             # direct-tcpip channel don't hang forever when the network misbehaves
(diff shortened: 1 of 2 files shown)


wowi42 (approved):
LGTM

Fixes the bug where `ProxyJump none` in ssh_config caused pyinfra to try connecting to a host literally called "none" instead of disabling the jump. The two-line guard checks the whole value case-insensitively (matching OpenSSH behavior) and the new test covers both `none` and `NONE` with assertions that no hop connection or gateway is created. Pattern matches the existing `IdentityAgent none` check a few lines above. Clean, focused, good.

PR #1923, opened 2026-08-15
fix(facts.sysvinit): allow underscores in service names

### LINKED ISSUE #1844: `bsdinit.service(running=True)` is not idempotent for non-base-install services
## Describe the bug

`bsdinit.service(running=True)` is not idempotent for non-base-install services on OpenBSD.

The following code will always execute, regardless if the service is already running:
```python
    bsdinit.service(
        service="node_exporter",
        running=True,
        _sudo=True,
    )
```

However, this seems to only be a problem with services that are are not part of the base install. The above code *is* idempotent with the service "unbound", which is part of the base install. I have not exhaustively tested this theory, but it holds true for 3 base-install services (`unbound`, `tftpd`, `ntpd`) and 3 services installed via `pkg_add` (`node_exporter`, `unbound_exporter`, `isc_dhcpd`).

Looking at [bsdinit's source](https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/operations/bsdinit.py), I suspected that the issue is with the [RcdStatus](https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/facts/bsdinit.py) fact; however, when I run the shell code on the OpenBSD machine, everything returns as I would expect:
```bash
for SERVICE in `find /etc/rc.d /usr/local/etc/rc.d -type f`; do $SERVICE status 2> /dev/null || $SERVICE check 2> /dev/null ; echo "`basename $SERVICE`=$?" ; done
...
unbound(ok)
unbound=0
...
node_exporter(ok)
node_exporter=0
...
```

Looking at the permissions of the rc files themselves:
```
rtr5$ ls -lah /etc/rc.d/{tftpd,unbound,ntpd,node_exporter,unbound_exporter,isc_dhcpd} 
-rwxr-xr-x  1 root  bin     137B Oct  9  2025 /etc/rc.d/isc_dhcpd
-rwxr-xr-x  1 root  bin     181B Oct  9  2025 /etc/rc.d/node_exporter
-r-xr-xr-x  1 root  wheel   137B Oct 12  2025 /etc/rc.d/ntpd
-r-xr-xr-x  1 root  wheel   164B Oct 12  2025 /etc/rc.d/tftpd
-r-xr-xr-x  1 root  wheel   372B Oct 12  2025 /etc/rc.d/unbound
-rwxr-xr-x  1 root  bin     251B Oct  9  2025 /etc/rc.d/unbound_exporter
```

The permissions do not look problematic, but just in case, I altered `node_exporter` to match to base-install services. It did not change the behavior.

So I'm left with opening a bug report for now, without an accompanying PR.

Any ideas for further investigation is most welcome. :-) 

## To Reproduce

+ Operation code & usage

see above

+ Target system information

OpenBSD 7.8

## Expected behavior

I expect `bsdinit.service(running=True)` to behave consistently idempotent across all services.

## Meta

```
❱ pyinfra --support 

    System: Linux
      Platform: Linux-7.0.10+deb14-amd64-x86_64-with-glibc2.42
      Release: 7.0.10+deb14-amd64
      Machine: x86_64
    pyinfra: v3.9.2
      click: v8.4.1
      distro: v1.9.0
      gevent: v26.5.0
      jinja2: v3.1.6
      packaging: v26.2
      paramiko: v4.0.0
      pydantic: v2.13.4
      python-dateutil: v2.9.0.post0
      typeguard: v4.5.2
      typing-extensions: v4.15.0
    Executable: /home/aqw/.venvs/pyinfra/bin/pyinfra
    Python: 3.13.14 (CPython, GCC 15.2.0)
```

```
uv pip install pyinfra
```

none

If you would like the output, please let me know. I've reviewed it, and I was unable to find anything informative.

NoiceHax:
Fixes #1844.

`InitdStatus.regex` was `([a-zA-Z0-9\-]+)=([0-9]+)`, which has no underscore in the character class. Because `re.match` anchors at the start of the line, a line like `node_exporter=0` does not match at all, so the service is dropped from the dict the fact returns. `bsdinit.RcdStatus` inherits the regex, so `bsdinit.service(running=True)` found no entry for the service, treated it as stopped, and issued a start command on every run. Underscores are common in rc.d and init.d script names, for example `node_exporter` and `isc_dhcpd`.

This widens the character class to `([a-zA-Z0-9_\-.]+)`, which also covers dots, another character these script names use. I added a fact fixture for `bsdinit.RcdStatus` with underscored service names, and added one underscored service to the existing `sysvinit.InitdStatus` fixture so the Linux path is covered too. Both fixtures fail before the change and pass after it.

*AI disclosure (per `AI_POLICY.md`):* prepared with Claude Code, which traced the regex and drafted the test fixtures, and I reviewed the change and ran the tests before submitting.

### DIFF
--- a/src/pyinfra/facts/sysvinit.py
+++ b/src/pyinfra/facts/sysvinit.py
@@ -32,7 +32,7 @@ class InitdStatus(FactBase):
         done
     """
 
-    regex = r"([a-zA-Z0-9\-]+)=([0-9]+)"
+    regex = r"([a-zA-Z0-9_\-.]+)=([0-9]+)"
     default = dict
 
     @override
--- a/tests/facts/bsdinit.RcdStatus/services.yaml
+++ b/tests/facts/bsdinit.RcdStatus/services.yaml
@@ -1,0 +1,11 @@
+command: "\n            for SERVICE in `find /etc/rc.d /usr/local/etc/rc.d -type f`; do\n                $SERVICE status 2> /dev/null || $SERVICE check 2> /dev/null\n                echo \"`basename $SERVICE`=$?\"\n            done\n        "
+output: |
+  node_exporter=0
+  unbound=0
+  isc_dhcpd=1
+  sshd=0
+fact:
+  node_exporter: true
+  unbound: true
+  isc_dhcpd: false
+  sshd: true
--- a/tests/facts/sysvinit.InitdStatus/services.json
+++ b/tests/facts/sysvinit.InitdStatus/services.json
@@ -3,11 +3,13 @@
     "output": [
         "rpcsvcgssd=0",
         "rsyslog=4",
-        "saslauthd=3"
+        "saslauthd=3",
+        "node_exporter=0"
     ],
     "fact": {
         "rpcsvcgssd": true,
         "rsyslog": null,
-        "saslauthd": false
+        "saslauthd": false,
+        "node_exporter": true
     }
 }


wowi42 (approved):
LGTM

Fixes #1844: `InitdStatus.regex` had no underscore in its character class, so `re.match` dropped any init.d/rc.d service with an underscore in its name (node_exporter, isc_dhcpd). Since `bsdinit.RcdStatus` inherits the regex, `bsdinit.service` saw those services as missing, treated them as stopped, and re-issued a start on every run. The fix widens the class to `[a-zA-Z0-9_\-.]`, also covering dots. I verified the old regex drops underscored and dotted names and the new one catches them, and that the widened class is strictly additive, so existing behavior for already-matched names is unchanged.

Both fixtures collect and pass on the PR branch (`RcdStatus or InitdStatus`: 2 passed). The new `tests/facts/bsdinit.RcdStatus/services.yaml` fills a real gap, since that fact previously had no fixture at all, and its command string matches the rendered `bsdinit.py` command exactly. The edit to the existing `sysvinit.InitdStatus` fixture is additive only: one underscored service appended, no existing expected values touched. The hyphen in the class is escaped, so no accidental range. Scope is atomic, CI is fully green, and the body explains the root cause well. Nothing to change.

Issue #1440, opened 2025-08-27
files.get with sudo  fail with "Permission denied"


KarolBedkowski:
Hi,

## Describe the bug
Using files.get(..., _sudo=True) fail with PermissionError; example:

```
--> Starting operation: download etc files | get  
[rockpi.home.arpa]  >>> sudo -H -n sh -c '! (test -e /etc/nftables.conf || test -L /etc/nftables.conf ) || ( stat -c '"'"'user=%U group=%G mode=%A atime=%X mtime=%Y ctime=%Z size=%s %N'"'"' /etc/nftables.conf 2> /dev/null || stat -f '"'"'user=%Su group=%Sg mode=%Sp atime=%a mtime=%m ctime=%c size=%z %N%SY'"'"' /etc/nftables.conf )'
[rockpi.home.arpa]  user=root group=adm mode=-rwxr-x--- atime=1712484978 mtime=1756193730 ctime=1756193730 size=8185 '/etc/nftables.conf'
    [rockpi.home.arpa]  Loaded fact files.File (path=/etc/nftables.conf)
[rockpi.home.arpa]  >>> sudo -H -n sh -c 'cp /etc/nftables.conf /tmp/user/0/pyinfra-a302f6cd8526066f2904619397ee54c2660d795a && chmod +r /tmp/user/0/pyinfra-a302f6cd8526066f2904619397ee54c2660d795a'
[rockpi.home.arpa]  >>> sudo -H -n sh -c 'rm -f /tmp/user/0/pyinfra-a302f6cd8526066f2904619397ee54c2660d795a'
    [rockpi.home.arpa]  Command socket/SSH error: PermissionError(13, 'Permission denied')
    [rockpi.home.arpa]  Error: executed 0 commands

```

File is copied to tmp dir but instead of use tmp of user that is used for connection, is used root's tmp dir (/tmp/user/0/) and other users don't have access to it (all dirs in /tmp/user are 700).

On error local file is truncated.

Not sure how to fix it without breaking other things...

## To Reproduce

``` 
files.get(
        name="get ",
        src="/etc/nftables.conf",
        dest="nftables.conf",
        force=True,
        _sudo=True,
    )
```

Target os: Debian 13.

## Expected behavior

With sudo there should be possible to get any file.

## Meta

+ Include output of `pyinfra --support`.
``` 
    If you are having issues with pyinfra or wish to make feature requests, please
[...]

PR #1924, opened 2026-08-15
fix(connectors.ssh): don't truncate local file when download fails (#1440)

NoiceHax:
Part of #1440, not a full fix.

`SSHConnector._get_file` opened the local destination with `get_file_io(..., "wb")`, which truncates it, and only then started the transfer. If the transfer failed, the user was left with an empty local file and no way back to the original contents. In #1440 the download fails with `PermissionError`, so `files.get` with `_sudo=True` wipes the local copy.

The download now goes through a `SpooledTemporaryFile` first, and the destination is only opened once the remote file has arrived in full. Spooling keeps small downloads in memory and spills larger ones to disk, so this does not pull whole files into RAM. Both transfer backends write into the file object they are given, so `sftp` and `scp` behave the same.

This addresses the truncation half of #1440 that @Fizzadar flagged. The sudo temp dir resolution is left open, so please don't auto-close the issue on merge. `docker.py` and `chroot.py` have the same shape in their download paths and are untouched here, they belong in a separate PR.

New test `test_get_file_failure_leaves_local_file_alone` writes a real file to a temp dir, makes the mocked `getfo` raise `PermissionError`, and asserts the bytes on disk are unchanged. It fails on `3.x` (reads `b''`) and passes with the fix.

## Checklist

- [x] Pull request includes documentation for any new/updated operations/facts (no operation, fact or argument changed, so nothing to document)
- [x] Tests pass (`scripts/dev-test.sh`, 1943 passed / 1 skipped)

## AI usage disclosure

Per [AI_POLICY.md](../blob/3.x/AI_POLICY.md): written with AI assistance using Claude Code, which located the truncating write, drafted the buffered version and the regression test, and ran the tests and linters. I reviewed and understand the change, including why the spooled buffer keeps the `sftp` and `scp` paths identical and why the sudo temp dir half of the issue is out of scope here.

### DIFF
--- a/src/pyinfra/connectors/ssh.py
+++ b/src/pyinfra/connectors/ssh.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import os
 from random import uniform
-from shutil import which
+from shutil import copyfileobj, which
 from socket import gaierror
+from tempfile import SpooledTemporaryFile
 from time import sleep
 from typing import IO, TYPE_CHECKING, Any, Protocol
 from collections.abc import Iterable
@@ -33,6 +34,10 @@ from .util import (
 
 if TYPE_CHECKING:
     from pyinfra.api.arguments import ConnectorArguments
+
+# Downloads are buffered in memory up to this size before spilling over onto disk, so that
+# the local destination is only written once the remote file has arrived in full.
+GET_FILE_BUFFER_MAX_SIZE = 1024 * 1024  # 1MB
 
 
 class ConnectorData(TypedDict):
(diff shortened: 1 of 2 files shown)


wowi42 (approved):
LGTM

This fixes the truncation half of #1440 by buffering the SFTP download into a SpooledTemporaryFile before opening the local destination. The fix is minimal, the test is real (uses the actual filesystem, no mock_open), and the failed-transfer path is the only behavior change. Scope is clean: only ssh.py, with docker and chroot left for follow-ups as the body says. No shell surface, no compat break, no new API.

PR #1925, opened 2026-08-15
fix(connectors.docker): strip sha256 prefix instead of slicing image ID

### LINKED ISSUE #1905: Podman connector: The reported truncated image id of build is wrong (or at least unexpected)
Building an image using pyinfra returns:

```
[@podman/docker.io/fedora:44] podman build complete, image ID: 3b9bc6d8d5fc
```

Using podman (5.8.4):

```
❯ podman images --format "table {{.ID}} {{.Id}}" | head -n 2
IMAGE ID      <no value>
0d05fac3b9bc  0d05fac3b9bc6d8d5fc7b323a35a23daf0ab61fc38685de569ce69f6feb7fbe6
```

The returned image id from pyinfra is starting 7 chars in into the full-length image id, I'm guessing this is the culprit:

```
https://github.com/pyinfra-dev/pyinfra/blob/3.x/src/pyinfra/connectors/docker.py#L179
```

Unsure whether this is specifically podman or also a docker connector problem.

NoiceHax:
Fixes #1905.

`docker commit` prints the new image ID as `sha256:<digest>`, so `disconnect()` took a fixed `[7:19]` slice to log the short ID. `podman commit` prints the bare digest with no prefix, so that same slice cut characters 7 to 19 out of the digest itself. The podman connector then reported a short image ID that does not match any real image.

The fix removes the `sha256:` prefix when it is there and takes the first 12 characters of what is left, which is correct for both tools. The `run -d` path is untouched, since the container ID it returns was never sliced.

On the test side, the podman fake shell now returns a bare 64 character digest like podman really does, the docker fake keeps the prefixed form, and a new test asserts the logged ID is the first 12 characters of the digest. It runs for both connectors through the shared base class.

I used Claude Code to write this patch and its test. I reviewed the change, ran the tests, and understand what it does.

### DIFF
--- a/src/pyinfra/connectors/docker.py
+++ b/src/pyinfra/connectors/docker.py
@@ -174,9 +174,13 @@ class DockerConnector(BaseConnector):
         quoted_container_id = shlex.quote(container_id)
 
         with progress_spinner({f"{self.docker_cmd} commit"}):
-            image_id = local.shell(
-                f"{self.docker_cmd} commit {quoted_container_id}", splitlines=True
-            )[-1][7:19]  # last line is the image ID, get sha256:[XXXXXXXXXX]...
+            # last line is the image ID, optionally prefixed with the digest algorithm
+            # (docker prints "sha256:XXXX...", podman prints the bare digest)
+            image_id = (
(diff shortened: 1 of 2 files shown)


wowi42 (approved):
LGTM

PR #1926, opened 2026-08-15
fix(connectors.ssh): skip unparsable known_hosts lines instead of dropping the file

### LINKED ISSUE #1339: Duplicate entries in known_hosts
## Describe the bug

Pyinfra creates duplicate entries in `~/.ssh/known_hosts` when unable to connect.

## To Reproduce

```
$ cat ~/.ssh/known_hosts
@cert-authority *.internal.example.com ssh-ed25519 <key> <comment>

$ pyinfra $host exec -- echo "hello world"
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    Failed to load host keys from /Users/<user>/.ssh/known_hosts: ('@cert-authority *.internal.example.com ssh-ed25519 <key> <comment>', Error('Incorrect padding'))
    No host key for <host> found in known_hosts, accepting & adding to host keys file
    Added host key for <host> to known_hosts
    [<host>] Authentication error () (Authentication failed.)

$ pyinfra $host exec -- echo "hello world"
<same output>

$ pyinfra $host exec -- echo "hello world"
<same output>

$ cat ~/.ssh/known_hosts
@cert-authority *.internal.example.com ssh-ed25519 <key> <comment>
<host> ssh-ed25519 <key>
<host> ssh-ed25519 <key>
<host> ssh-ed25519 <key>
```

## Expected behavior
Pyinfra should only add the key once. This is likely an edge case related to https://github.com/pyinfra-dev/pyinfra/issues/1209, I ran into this while using pyinfra to inspect a server that was not yet signed by the CA.

NoiceHax:
Fixes #1339.

`get_host_keys` calls paramiko's `HostKeys.load`, which only skips lines that raise `SSHException`. A `@cert-authority` line raises `InvalidHostKey`, which is a plain `Exception`, so the load stops there and the rest of the file is thrown away. pyinfra then sees no known hosts, treats hosts it already knows as new, and with `StrictHostKeyChecking accept-new` writes their keys to `known_hosts` again on every run. That is where the duplicate entries come from.

This reads the file line by line, skips blank lines and comments the same way paramiko does, and drops only the lines that fail to parse, logging a warning with the file name and line number. Keys after a bad line are kept.

The test writes a `known_hosts` whose first entry is an unparsable `@cert-authority` line followed by a valid ecdsa key, then checks that key still resolves. It fails before the change and passes after.

*AI disclosure (per `AI_POLICY.md`):* the patch and test were written with Claude Code, then reviewed and tested by me before submitting.

### DIFF
--- a/src/pyinfra/connectors/sshuserclient/client.py
+++ b/src/pyinfra/connectors/sshuserclient/client.py
@@ -161,6 +161,38 @@ def get_ssh_config(user_config_file=None):
             return ssh_config
 
 
+def _load_host_keys_file(host_keys: HostKeys, filename: str) -> None:
+    """
+    Load a single known_hosts file, skipping any line paramiko cannot parse.
+
+    ``HostKeys.load`` only ignores lines that raise ``SSHException``, so a line
+    it fails on for any other reason (``@cert-authority``/``@revoked`` markers,
+    a truncated key) aborts the whole file and drops every valid key in it. That
+    makes pyinfra treat known hosts as unknown and append them again (issue
+    #1339). See: https://github.com/paramiko/paramiko/pull/1990
+    """
+
+    with open(filename) as f:
+        for lineno, line in enumerate(f, 1):
+            line = line.strip()
+            if not line or line.startswith("#"):
+                continue
+
+            try:
+                entry = HostKeyEntry.from_line(line, lineno)
+            # Broad by necessity: paramiko raises a bare ``Exception`` subclass
+            # (``InvalidHostKey``) for undecodable key data.
+            except Exception as e:
+                logger.warning("Skipping bad host keys line %s:%i: %s", filename, lineno, e)
+                continue
+
+            if entry is None:
+                continue
+
+            for hostname in entry.hostnames:
+                host_keys.add(hostname, entry.key.get_name(), entry.key)
+
+
 @memoize
 def get_host_keys(filenames):
     """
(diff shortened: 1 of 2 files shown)


wowi42 (approved):
LGTM

This fixes #1339. The root cause is real: paramiko's `HostKeys.load` only skips lines that raise `SSHException`, but a `@cert-authority` line raises `InvalidHostKey`, which is a plain `Exception`, so the load aborts and the rest of the file is thrown away. The new `_load_host_keys_file` helper mirrors paramiko's `load` exactly (verified against 3.5.1 source) but catches broad `Exception` per line, logs a warning with the file and line number, and keeps every valid key that follows a bad line. The wider `except Exception` is justified and documented.

The regression test is solid. It collects, passes with the fix, and I confirmed it fails without it by temporarily switching back to `host_keys.load` - the whole file gets dropped and `lookup` returns None. The example key genuinely triggers the bug: paramiko mis-reads the four-field line, so the key type lands in the base64 field and decode fails. All 19 tests in the file pass, and ruff check and format are clean. Only this one call site uses `HostKeys.load`, so nothing else needs the same change. Scope is tight and AI usage is disclosed with real validation.

Nit: the new helper's docstring is plain prose while `get_host_keys` above it uses the `Args:` style, but both are internal helpers, not operations or facts, so docs generation is unaffected. Not worth a change.

PR #1175, opened 2024-08-12
Fix tilde expansion (#1025)

simonhammes:
Idea: https://unix.stackexchange.com/questions/151850/why-doesnt-the-tilde-expand-inside-double-quotes#comment247796_151865

Fixes #1025

PR #1931, opened 2026-08-24
fix(facts.files): fix tilde expansion for HashFiles (#1235)

### LINKED ISSUE #1235: `files.put()` does not expand `~`
## Describe the bug
`files.put()` does not expand `~` to the value of `$HOME`

## To Reproduce

```python

files.put(name='Copy file', src='README.md', dest='~/pyinfra-README.md')
```

```bash
pyinfra @local test.py
```

## Observed behavior

A directory named `~` is created inside the current working directory.

## Expected behavior
The file is copied to `~/pyinfra-README.md`.

## Meta

<details><summary>System Information</summary>
<p>

```
System: Linux
      Platform: Linux-6.1.0-26-amd64-x86_64-with-glibc2.36
      Release: 6.1.0-26-amd64
      Machine: x86_64
    pyinfra: v3.1.1
      black: v24.4.2
      black: v24.4.2
      click: v8.1.7
      configparser: v7.1.0
      coverage: v7.5.1
      coverage: v7.5.1
      distro: v1.9.0
      flake8: v7.0.0
      flake8: v7.0.0
      flake8-black: v0.3.6
      flake8-black: v0.3.6
      flake8-isort: v6.1.1
      flake8-isort: v6.1.1
      flake8-spellcheck: v0.12.1
      gevent: v24.10.3
      importlib-metadata: v8.5.0
      ipdb: v0.13.13
      ipdbplugin: v1.5.0
      ipython: v8.29.0
      isort: v5.13.2
      isort: v5.13.2
      jinja2: v3.1.4
      mypy: v1.13.0
      mypy: v1.13.0
      myst-parser: v2.0.0
      myst-parser: v2.0.0
      packaging: v24.1
      paramiko: v3.5.0
      pyinfra-guzzle-sphinx-theme: v0.16
      pyinfra-guzzle-sphinx-theme: v0.16
      pytest: v8.2.1
      pytest: v8.2.1
      pytest-cov: v5.0.0
      pytest-cov: v5.0.0
      python-dateutil: v2.9.0.post0
      pywinrm: v0.5.0
      redbaron: v0.9.2
      setuptools: v66.1.1
      sphinx: v6.2.1
      sphinx: v6.2.1
      twine: v5.1.1
      typeguard: v4.4.0
      types-PyYAML: v6.0.12.20240917
      types-PyYAML: v6.0.12.20240917
      types-cryptography: v3.3.23.2
      types-cryptography: v3.3.23.2
      types-paramiko: v3.5.0.20240928
      types-paramiko: v3.5.0.20240928
      types-python-dateutil: v2.9.0.20241003
      types-python-dateutil: v2.9.0.20241003
      types-setuptools: v75.2.0.20241025
      types-setuptools: v75.2.0.20241025
      typing-extensions: v4.12.2
      wheel: v0.44.0
    Executable: /data/Code/pyinfra/venv/bin/pyinfra
    Python: 3.11.2 (CPython, GCC 12.2.0)
```

</p>
</details>

SecT0uch:
Fixes #1235
Heavily inspired by #1175 

Edit:

Tested from my branch.
Now with:

```python

files.put(
    src="files/editors.conf",
    dest= "~/.config/environment.d/editors.conf"
)
```

Running `pyinfra @local deploy.py --dry --diff` correctly returns no change (assuming the file is already there)

### DIFF
--- a/src/pyinfra/facts/files.py
+++ b/src/pyinfra/facts/files.py
@@ -398,15 +398,21 @@ class HashFileFactBase(FactBaseOptionalStr):
 
     @override
     def command(self, path):
-        self.path = path
-        return make_formatted_string_command(self._raw_cmd, QuoteString(path))
+        if path.startswith("~/"):
+            # Insert .* to take into account string expansion
+            self.escaped_path = ".*" + re.escape(path.removeprefix("~"))
+            # Do not quote leading tilde to ensure that it gets properly expanded by the shell
+            path = StringCommand("~/", QuoteString(path.removeprefix("~/")), _separator="")
+        else:
+            self.escaped_path = re.escape(path)
+            path = QuoteString(path)
+        return make_formatted_string_command(self._raw_cmd, path)
 
     @override
     def process(self, output) -> str | None:
         output = output[0]
-        escaped_path = re.escape(self.path)
         for regex in self._regexes:
-            matches = re.match(regex % escaped_path, output)
+            matches = re.match(regex % self.escaped_path, output)
             if matches:
                 return matches.group(1)
         return None
--- a/tests/facts/files.Md5File/tilde.yaml
+++ b/tests/facts/files.Md5File/tilde.yaml
@@ -1,0 +1,5 @@
+"arg": "~/myfile"
+"command": "test -e ~/myfile && ( md5sum ~/myfile 2> /dev/null || md5 ~/myfile 2> /dev/null ) || true"
+"output":
+  - "c10ba97d7c9078a006d26b5db01d8ee7 /home/pyinfra/myfile"
+"fact": "c10ba97d7c9078a006d26b5db01d8ee7"
--- a/tests/facts/files.Md5File/tilde_bsd_style.yaml
+++ b/tests/facts/files.Md5File/tilde_bsd_style.yaml
@@ -1,0 +1,5 @@
+"arg": "~/myfile"
+"command": "test -e ~/myfile && ( md5sum ~/myfile 2> /dev/null || md5 ~/myfile 2> /dev/null ) || true"
+"output":
+  - "MD5 (/home/pyinfra/myfile) = c10ba97d7c9078a006d26b5db01d8ee7"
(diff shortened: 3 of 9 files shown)


wowi42 (commented):
Holding comment before the technical points: tilde expansion in pyinfra is still under discussion on Matrix, and no decision has been taken on whether pyinfra should support it at all. I am personally against it. Until that discussion reaches a conclusion, this PR should not be merged.

Technical review, in case the decision goes in favor of tilde support:

1. The regex change in `__init_subclass__` weakens path validation for every hash fact call, not just tilde paths. `^([a-fA-F0-9]{%d})\s+.*%%s$` turns an exact path match into a suffix match. If this lands, the wildcard should only apply to the tilde case; plain paths should keep the strict regex.

2. The root problem is that `~` is expanded by the remote shell, so the client cannot know the expanded path at parse time, hence the `.*` wildcard. That trade-off (silently accepting any path ending in the same suffix) is exactly the kind of thing being weighed in the Matrix discussion. An alternative worth considering there: drop the filename check entirely for tilde paths instead of weakening it for all paths, since the command only ever hashes the one requested file.

3. Only `~/` is handled, not `~user/`. Partial tilde support is arguably worse than none, since `files.put(dest="~other/file")` would still silently create a literal `~other` directory.

4. The change to the `File` fact (`path[2:]` -> `removeprefix("~/")`) is unrelated to this fix and should be split out.

5. New fixtures should be YAML, not JSON (project convention). Also, the fixtures only cover GNU coreutils output; BSD output with an expanded home path is untested, and the BSD regex has the same suffix-match problem.

SecT0uch:
While waiting for the decision I did the following:
- the `.*` is added only for paths with a tilde and unchanged for the rest
- removed the commit changing the `File` fact (`path[2:]` -> `removeprefix("~/")`)
- Changed the fixtures to YAML format
- Added the fixtures for BSD

That leaves only the `~user/` feature, where we'd have to implement it for both classes `File` and `HashFileFactBase`
Do you think this should come separately ?

PR #1932, opened 2026-08-25
feat: add files.rename operation

### LINKED ISSUE #1867: Add rename operation
Currently it is not possible to rename a file via the `files` operations. There is an operation called `files.move`, but it only moves the file to a new directory and explicitly checks if the target is a directory and merges the source filename with the target directory name.

```python
@operation()
def move(src: str, dest: str, overwrite=False)
...

if not host.get_fact(Directory, dest):
    raise OperationError(f"dest {dest} is not an existing directory")

full_dest_path = os.path.join(dest, os.path.basename(src))
```

Either remove the restriction of `files.move` or create new operation called `files.rename` which can be used to rename files.

wowi42:
Adds a `files.rename` operation that renames a remote file, directory or link to a new path. Unlike `files.move`, the destination is the full new path, not a directory to move into.

- Errors if `src` does not exist
- Errors if `dest` already exists, unless `overwrite=True` (which removes it first)

Closes #1867


src/pyinfra/operations/files.py
```diff
@@ -1499,6 +1499,28 @@ def move(src: str, dest: str, overwrite=False):
...
+    + dest: new path for `src`
+    + overwrite: whether to overwrite dest, if present
+    """
+
+    if host.get_fact(File, src) is None:
+        raise OperationError(f"src {src} does not exist")
+
+    if host.get_fact(File, dest) is not None:
+        if overwrite:
+            yield StringCommand("rm", "-rf", QuoteString(dest))
```
evilham, line 1517: This **really** feels like we ought to have more safeguards around `dest` not being `/` due to e.g. a wrong string interpolation on pyinfra's user side.

~~OTOH, `files.move` has the exact same issue~~ (not really because it uses the file name at least): https://github.com/pyinfra-dev/pyinfra/blob/d525dcb3990e86974d0a0fe1125e9dabff262f9f/src/pyinfra/operations/files.py#L1495

While there though, wouldn't it make sense to refactor `files.move` so these two share the same code?

Kind of:

```
def move(...):
    full_dest_path = posixpath.join(dest, PurePosixPath(src).name)
    rename(src, full_dest_path, overwrite=overwrite)
```

Fizzadar (approved):

PR #1934, opened 2026-08-25
fix(operations.git.repo): pass depth argument to fetch and pull

evilham:
Previously, when cloning a repo with the `depth` keyword, if the remote directory changed, fetch and/or pull operations would unshallow the repository, given that the `depth` keyword was not being passed.

In turn, this ensures that if the value for the `depth` keyword changes, it will be updated in the deployment, though only when there is a change in the remote repository.

- [X] Pull request title follows the 

### DIFF
--- a/src/pyinfra/operations/git.py
+++ b/src/pyinfra/operations/git.py
@@ -118,7 +118,7 @@ def repo(
     + ssh_keyscan: keyscan the remote host if not in known_hosts before clone/pull
     + update_submodules: update any git submodules
     + recursive_submodules: update git submodules recursively
-    + depth: create a shallow clone with a history truncated to the specified number of commits
+    + depth: truncate clone, fetch and pull history to the specified number of commits
     + fetch_tags: Whether all tags should be fetched prior to attempting to
       check out the specified revision.
     + force: Execute ``fetch``, ``pull`` and ``checkout`` commands with ``--force``.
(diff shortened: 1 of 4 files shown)


wowi42 (approved):
Approving: the change is correct and well-scoped, and CI is green. The `depth_options` reuse across clone/fetch/pull is clean, and the flag order in the generated commands is valid (`git pull --depth 1 --rebase`, `git fetch --depth 1 --tags`). The note in the PR body about depth changes only applying on remote changes matches the code.

Minor notes, none blocking:

1. The new fixtures are JSON. The repo convention (AGENTS.md) is to prefer YAML for new fixtures, so `branch_pull_depth.yaml` / `clone_depth.yaml` would be better.
2. `branch_pull_depth.json` carries `second_output_commands`, `idempotent` and `disable_idempotent_warning_reason` over from `branch_pull.json`. The current testgen harness silently ignores those keys, so the claimed second-run pull is never actually tested. Pre-existing repo-wide cruft, but copying it spreads it.
3. `str(depth)` goes into the `StringCommand` unquoted. Pre-existing style in this function and safe since `depth` is typed `int | None`, but the repo convention is to wrap everything in `QuoteString`.

Fizzadar (approved):
Nice, thank you @evilham! I rebased to fix a conflict, merging!

Fizzadar (approved):
Nice, thank you @evilham! I rebased to fix a conflict, merging!

PR #1938, opened 2026-08-27
feat(connectors.docker): add docker_user connector data option

### LINKED ISSUE #661: Allowing docker user to be specified
Allow the docker user to be specified when building a docker image. At the moment, we can already specify the ssh user (`--user, --ssh-user`) and winrm user (`--winrm-username`), so it would make sense to have a `--docker-user` or something similar that will allow us to build an image using a specific user that is not root. This of course is dependent on that user being available in the image.

wowi42:
Closes #661.

Adds a `docker_user` host data option to the docker connector (inherited by the podman connector). When set, commands run as that user inside the container:

```
docker exec --user <user> -i <container> sh -c ...
```

Usage, following the existing connector data convention (docker options have no dedicated CLI flags):

```shell
pyinfra @docker/alpine:3.8 --data docker_user=someuser ...
```

```python
@docker/myimage, {"docker_user": "someuser"}
```

Notes:

- The user must exist in the image, as the issue states.
- File up/downloads use `docker cp`, which is daemon-side and unaffected by the exec user.
- Connector docs pick the option up automatically from `connector_data_meta`.

Tests: new `test_run_shell_command_with_user` in `tests/test_connectors/test_docker.py`, run for both the docker and podman test classes.

### DIFF
--- a/src/pyinfra/connectors/docker.py
+++ b/src/pyinfra/connectors/docker.py
@@ -29,12 +29,14 @@ class ConnectorData(TypedDict):
     docker_identifier: str
     docker_platform: str
     docker_architecture: str
+    docker_user: str
 
 
 connector_data_meta: dict[str, DataMeta] = {
     "docker_identifier": DataMeta("ID of container or image to start from"),
     "docker_platform": DataMeta("Platform to use for Docker image (e.g., linux/amd64)"),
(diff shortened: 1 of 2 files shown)


Fizzadar (approved):
LGTM!

PR #1939, opened 2026-08-27
refactor(facts): no sudo to get pacman facts

SecT0uch:
It seems that facts requested while evaluating an operation inherit the operation's `_sudo` argument.

For example, `pacman.packages(_sudo=True)` currently collects `PacmanPackages` with sudo.

However, `pacman -Q` is a read-only query and does not require elevated privileges. As a result, a package operation can require sudo authentication even when no package changes are necessary.

I wonder if this should become a convention for operations more generally.

For facts which only perform read-only, unprivileged state discovery, operations could explicitly request them with `_sudo=False`, even when the operation itself requires sudo.

This would have two benefits:

1. read-only fact collection doesn't unnecessarily require elevated privileges.
2. an operation which makes no changes should not need sudo credentials just to determine that no changes are required.

This could also be extended to `doas` and `_dzdo`

- [x] Pull request title follows the  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format

### DIFF
--- a/src/pyinfra/operations/pacman.py
+++ b/src/pyinfra/operations/pacman.py
@@ -74,9 +74,11 @@ def packages(
     yield from ensure_packages(
         host,
         packages,
-        host.get_fact(PacmanPackages),
+        host.get_fact(PacmanPackages, _sudo=False),
         present,
         install_command="pacman --noconfirm -S",
         uninstall_command="pacman --noconfirm -R",
-        expand_package_fact=lambda package: host.get_fact(PacmanUnpackGroup, package=package),
+        expand_package_fact=lambda package: host.get_fact(
+            PacmanUnpackGroup, package=package, _sudo=False
+        ),
     )


wowi42:
Thanks, this is correct and safe. Verified:

- The override works: `_handle_fact_kwargs` (`src/pyinfra/api/facts.py:148-156`) applies call-site kwargs on top of the operation context kwargs, so `_sudo=False` here wins over an operation-level `_sudo=True`, and it flows through `global_kwargs` into the connector executor.
- No cache poisoning: the fact cache key is built from `fact_kwargs`, which excludes global arguments, so a sudo-less gather and a sudo gather share one cache entry. Harmless here since `pacman -Q` and `pacman -S --print-format` return identical data either way.
- Both facts are genuinely unprivileged: `pacman -Q` is a plain db query and `pacman -S --print-format` is a print-only transaction simulation (same form AUR helpers use unprivileged). Neither takes the db lock.

A few comments:

1. Please drop the module docstring addition ("The facts called by these operations will be gathered with `_sudo=False`."). Module docstrings are rendered into the user-facing operations docs page, this is an internal implementation detail, and it overclaims: of the three operations in the module, only `packages` gathers facts. If you want it documented, put it in the `packages` docstring, but I would just remove it.

2. Nothing tests the new behavior: operation fixtures mock facts, so the suite still passes if the `_sudo=False` kwargs are removed. Not blocking since fixtures can't easily express this, just worth knowing.

3. On the convention question in the PR body: call-site `_sudo=False` works, but a declarative attribute on the fact class (something like `FactBase.requires_root = False`) would scale better than auditing every operation. That's a larger API discussion and should not block this change. If the convention spreads, the apt/yum/apk/dpkg package-list facts are the obvious next candidates.

SecT0uch:
Thanks for your review!

I've removed the change on the docstring.

I agree that an attribute like `FactBase.requires_root` sounds like a better idea.

SecT0uch:
I drafted this PR, waiting for the review of #1954, since it would be a better implementation.

Fizzadar:
Closing for #1954 which is in discussion

PR #1945, opened 2026-09-05
fix(connectors.ssh): trust known_hosts certificate authorities

### LINKED ISSUE #1944: openssh host ca with cert-auth and wildcard not working
## Describe the bug
Pyinfra doesn't accept @cert-authority with wild cards like 192.168.1.* in the known_hosts 
 "    Failed to load host keys from /home/martijn/.ssh/known_hosts: ('@cert-authority 192.168.1.* ssh-ed25519 THESECRETKEYWASHERE', Error('Incorrect padding'))
    No host key for 192.168.1.236 found in known_hosts, ... "

## To Reproduce
See the comment below

## Expected behavior
If direct ssh to the host works from the CLI then pyinfra should also work with ssh host ca and cert-auth for the keys

## Meta
+ How was pyinfra installed (source/pip)? pip (pyinfra v3.9) venv (but also in a container is doesn't work)

0xTaoZ:
## Summary

Fixes #1944.

pyinfra already skipped `@cert-authority` known_hosts lines because Paramiko cannot parse them as normal host keys, but that also dropped the CA trust rule. With strict host key checking enabled, a server presenting an OpenSSH host certificate signed by that CA still looked like an unknown host.

This change keeps `@cert-authority` entries separately while loading known_hosts files, then accepts a missing host key when the presented server key has a valid OpenSSH host certificate that:

- is signed by a CA trusted for the target hostname pattern
- includes the target hostname as a certificate principal
- is currently within its validity window

Normal host-key loading and the existing handling for bad known_hosts lines are unchanged.

## Validation

- `uv run pytest tests/test_connectors/test_sshuserclient.py -q`
- `uv run pytest tests/test_connectors/test_sshuserclient.py tests/test_connectors/test_ssh.py -q`
- `uv run ruff check src/pyinfra/connectors/sshuserclient/client.py tests/test_connectors/conftest.py tests/test_connectors/test_sshuserclient.py`
- `uv run ruff format --check src/pyinfra/connectors/sshuserclient/client.py tests/test_connectors/conftest.py tests/test_connectors/test_sshuserclient.py`
- `uv run mypy src/pyinfra/connectors/sshuserclient/client.py`
- `git diff --check`

### DIFF
--- a/src/pyinfra/connectors/sshuserclient/client.py
+++ b/src/pyinfra/connectors/sshuserclient/client.py
@@ -4,7 +4,12 @@ source has now vanished (https://github.com/tobald/sshuserclient).
 """
 
 import os
+from base64 import b64decode
+from binascii import Error as BinasciiError
+from collections.abc import Iterator
+from fnmatch import fnmatchcase
 from pathlib import Path
+from time import time
 
 from gevent.lock import BoundedSemaphore
 from paramiko import (
@@ -16,6 +21,8 @@ from paramiko import (
 )
 from paramiko.agent import AgentRequestHandler
 from paramiko.hostkeys import HostKeyEntry
+from paramiko.message import Message
+from paramiko.pkey import PKey, UnknownKeyType
 from typing_extensions import override
 
 from pyinfra import logger
@@ -30,10 +37,15 @@ from .config import SSHConfig
 
 HOST_KEYS_LOCK = BoundedSemaphore()
 
+CertAuthority = tuple[tuple[str, ...], PKey]
+
 
 class StrictPolicy(MissingHostKeyPolicy):
     @override
     def missing_host_key(self, client, hostname, key):
+        if _host_certificate_is_trusted(client, hostname, key):
+            return
+
         logger.error("No host key for %s found in known_hosts", hostname)
         raise SSHException(f"StrictPolicy: No host key for {hostname} found in known_hosts")
 
@@ -65,6 +77,9 @@ def append_hostkey(client, hostname, key):
 class AcceptNewPolicy(MissingHostKeyPolicy):
     @override
     def missing_host_key(self, client, hostname, key):
+        if _host_certificate_is_trusted(client, hostname, key):
+            return
+
         logger.warning(
             (
                 f"No host key for {hostname} found in known_hosts, "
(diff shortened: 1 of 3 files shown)


Fizzadar (approved):
LGTM! Thank you @0xTaoZ 

(I can't wait to be rid of paramiko!)

martijnvanveelen:
Paramiko —> asyncio

martijnvanveelen:
Hi 0xTaoZ, 
I've tested the new feature. And now at least for specific hosts there is a partial fix for the @cert-authority with pyinfra which is to replace @cert-authority 192.168.1.* ... with an entry per host e.g. @cert-authority 192.168.1.32 it is not perfect, but possibly workable. **Thanks!**.  Though in the end, I would want my known_hosts to be kept clean, and be able to use `192.168.1.*`. May be there is something in between strict host checking and  "no host checking and adding to known_hosts what ever pyinfra connects to. I would love to hear if you could realize something like that, and keep consistent with bare ssh functionality.

0xTaoZ:
@martijnvanveelen thanks for testing this. `192.168.1.*` is meant to work and is covered by the merged test, so could you share whether these hosts use a non-default port or a `HostName` alias, plus the exact hostname shown in pyinfra’s `No host key for ...` error? That should tell us whether Paramiko is passing a bracketed host and port or a resolved name into the matcher.

martijnvanveelen:
```inventory.py
hosts = [
  (
      '192.168.1.32',
      {
          'ssh_user': 'martijn',
          'ssh_port': 2231,
          'ssh_key': '~/.ssh/user_key',  # Path to your private key
          'ssh_cert': '~/.ssh/user_key-cert.pub',  # Optional: Path to the SSH certificate
          'ssh_strict_host_key_checking': 'yes',
      }
  ),
]
```

```~/.ssh/known_hosts
@cert-authority 192.168.1.* ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDM1gp9lP5m0773pu2VUEWgpYX/uIdGZQSbVcpeKpj7l SSH CA for pyinfra and host
```

0xTaoZ:
@martijnvanveelen the custom port explains it: Paramiko passes `[192.168.1.32]:2231`, while the CA check was comparing that value directly with the host pattern and certificate principal. I reproduced it and opened #1971 with coverage for your wildcard form as well as bracketed port-specific and negated patterns.

PR #1949, opened 2026-09-06
fix(api): strip su password prompt from fact output

### LINKED ISSUE #1947: Facts not working properly with su and FreeBSD target
## Describe the bug
When Pyinfra uses `su` on a FreeBSD target, it injects garbage into facts. On Linux it seems that `su` outputs the `Password:` prompt to stderr but on FreeBSD this gets sent to stdout instead. This means that on FreeBSD it will assume the password prompt as part of the command output.

## To Reproduce
`inventory.py`:
```python
servers = [
    "my_freebsd_server",
]
```

`group_data/servers.py`:
```python
import getpass

_su_user = "root"
# Needed because unlike sudo, it won't ask for a password interactively and will just fail
_su_password = getpass.getpass()
```

`userfact.py`:
```python
from pyinfra import host
from pyinfra.facts.server import Users

users = host.get_fact(Users)
print(f"{users=}")
```

Then run: `pyinfra inventory.py userfact.py`

Some facts might just include the password prompt in their output, the `Users` fact actually raises an error because of the unexpected command output.

Target system: FreeBSD 15.1-RELEASE-p3

## Expected behavior
I'd expect it to not crash/include a password prompt in the output 😉 

## Potential fix
What works for me is to trim the first few bytes of the output. I added this:
```python
    if executor_kwargs["_su_user"]:
        # Remove "Password:" from stdout
        stdout_lines[0]=stdout_lines[0][9:]
```

Below this part:
```python
    try:
        status, output = host.run_shell_command(
            command,
            print_output=state.print_fact_output,
            print_input=state.print_fact_input,
            **executor_kwargs,
        )
    except (TimeoutError, OSError, SSHException) as e:
        log_host_command_error(
            host,
            e,
            timeout=global_kwargs.get("_timeout"),
        )

    stdout_lines, stderr_lines = output.stdout_lines, output.stderr_lines
```

Of course this is not a proper fix as it should only trim the password prompt when using `su` and FreeBSD.

## Meta
```

    System: Linux
      Platform: Linux-6.18.37_1-x86_64-with-glibc2.41
      Release: 6.18.37_1
      Machine: x86_64
    pyinfra: v3.10.0
      click: v8.5.0
      distro: v1.9.0
      gevent: v26.8.0
      jinja2: v3.1.6
      packaging: v26.3
      paramiko: v4.0.0
      pydantic: v2.13.5
      python-dateutil: v2.9.0.post0
      typeguard: v4.6.0
      typing-extensions: v4.16.0
    Executable: /home/heavyjoost/.local/bin/pyinfra
    Python: 3.14.6 (CPython, GCC 14.2.1 20250405)
```

`uv tool install pyinfra`

0xTaoZ:
## Summary
- strip a leading `Password:` prompt from fact stdout when facts run through `su`
- keep normal fact stdout unchanged unless `_su_user` is active
- add regression coverage for the FreeBSD `su` prompt leaking into fact output

Fixes #1947

## Validation
- `uv run pytest tests/test_api/test_api_facts.py tests/test_connectors/test_util.py -q`
- `uv run ruff check src/pyinfra/api/facts.py tests/test_api/test_api_facts.py`
- `uv run ruff format --check src/pyinfra/api/facts.py tests/test_api/test_api_facts.py`
- `uv run mypy src/pyinfra/api/facts.py`
- `git diff --check`

## Risk
Low. The change only adjusts the first stdout line for fact commands executed with `_su_user`, and only when that line starts with the standard `Password:` prompt.

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -43,6 +43,7 @@ if TYPE_CHECKING:
 
 # Sentinel output line emitted when skip_unless_command binary is absent on the remote host.
 _MISSING_COMMAND_MARKER = "##PYINFRA_NOCMD##"
+_SU_PASSWORD_PROMPT = "Password:"
 
 SUDO_REGEX = r"^sudo: unknown user"
 SU_REGEXES = (
@@ -333,6 +334,15 @@ def _get_fact(
         )
 
     stdout_lines, stderr_lines = output.stdout_lines, output.stderr_lines
+    if (
+        executor_kwargs["_su_user"]
+        and stdout_lines
+        and stdout_lines[0].startswith(_SU_PASSWORD_PROMPT)
+    ):
+        stdout_lines = stdout_lines.copy()
+        stdout_lines[0] = stdout_lines[0][len(_SU_PASSWORD_PROMPT) :]
+        if stdout_lines[0] == "":
+            stdout_lines.pop(0)
 
     # Detect the "binary absent" sentinel from the if/then/else shell guard.
     if status and stdout_lines == [_MISSING_COMMAND_MARKER]:
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -22,7 +22,7 @@ from typing_extensions import override
 
 from pyinfra import logger
 from pyinfra.api.output import format_text
-from pyinfra.api import StringCommand
+from pyinfra.api import QuoteString, StringCommand
 from pyinfra.api.arguments import all_global_arguments, pop_global_arguments
 from pyinfra.api.exceptions import FactPreconditionError, FactProcessError, MissingCommandError
 from pyinfra.api.util import (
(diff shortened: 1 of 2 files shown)


src/pyinfra/api/facts.py
```diff
@@ -333,6 +334,15 @@ def _get_fact(
...
     stdout_lines, stderr_lines = output.stdout_lines, output.stderr_lines
+    if (
+        executor_kwargs["_su_user"]
+        and stdout_lines
+        and stdout_lines[0].startswith(_SU_PASSWORD_PROMPT)
+    ):
+        stdout_lines = stdout_lines.copy()
+        stdout_lines[0] = stdout_lines[0][len(_SU_PASSWORD_PROMPT) :]
+        if stdout_lines[0] == "":
+            stdout_lines.pop(0)
```
heavyjoost, line 345: Thanks for creating a potential fix so quickly!

Please correct me if I'm wrong but I think there might be two potential issues here:
1. On Linux `su` outputs the prompt to stderr. If for some reason my facts command output actually starts with `Password:`, it'll get stripped even though it's part of the real output and not the prompt.
2. Does it need to pop the first line if it's empty? What if the output of my command starts with a newline for some reason?

I'd say both of aren't very likely to happen though.

0xTaoZ: @heavyjoost You were right about both edge cases. I replaced prefix matching with an internal marker emitted immediately before the fact command, so only prompt text before that marker is discarded and the command's first line is left untouched. I also added regression coverage for legitimate `Password:` output and a leading newline in 305cc3fc.

heavyjoost (approved):
Haven't tested it but this should do the trick

Issue #1474, opened 2025-10-03
Async & immediate-mode API


JakkuSakura:
## Is your feature request related to a problem? Please describe
Currently pyinfra-cli is good on its own, but sometimes, we better integrate pyinfra into our internal tools. I tried to copy & adapt pyinfra-cli, but only managed to make it non-async and planned mode, just like manually calling pyinfra-cli
I also made an attempt with help from AI. This version is async and can work on its own. 
```python
"""
Async wrapper for PyInfra operations.

Simple host-aware async wrapper that accepts operations as parameters
rather than providing custom wrappers for each operation type.
"""
import asyncio
import logging
from contextlib import ExitStack
from threading import Lock
from typing import Any, Callable, Optional
from concurrent.futures import ThreadPoolExecutor

from pyinfra.api.host import Host
from pyinfra.api.operation import OperationMeta, execute_immediately
from pyinfra.api.state import State
from pyinfra.context import ctx_config, ctx_host, ctx_inventory, ctx_state

logger = logging.getLogger(__name__)


class AsyncPyInfra:
    """
    Simple async wrapper for PyInfra operations.
    
    Host-aware and accepts operations as parameters for maximum flexibility.
    """

    def __init__(
            self,
            host: Optional[Host] = None,
            state: Optional[State] = None,
            *,
            max_workers: int = 1,
    ):
        """
        Initialize AsyncPyInfra wrapper.
        
        Args:
            host: Specific host to run operations on
            state: PyInfra state object
            max_workers: Size of the thread pool used when offloading operations
        """
        self.host = host
        self.state = state
        self._executor: Optional[ThreadPoolExecutor] = None
        self._executor_lock = Lock()
        self._max_workers = max_workers
        self._config = state.config
        self._inventory = state.inventory

    @property
    def executor(self):
        """Get or create thread pool executor."""
        if self._executor is None:
            with self._executor_lock:
                if self._executor is None:
                    self._executor = ThreadPoolExecutor(
                        max_workers=self._max_workers,
                        thread_name_prefix="async-pyinfra",
                    )
        return self._executor

    async def run_operation(self, operation: Callable, *args, **kwargs) -> Any:
        """
        Run any pyinfra operation asynchronously.
        
        Args:
            operation: The pyinfra operation function (e.g., files.directory, systemd.service)
            *args: Arguments to pass to the operation
            **kwargs: Keyword arguments to pass to the operation
            
        Returns:
            Result of the operation
        """
[...]

PR #1951, opened 2026-09-07
feat(api): add AsyncPyinfra asyncio embedding API

wowi42:
## Summary

Add `pyinfra.async_api.AsyncPyinfra`, an asyncio-compatible embedding API that runs pyinfra's synchronous (gevent-based) core on a dedicated worker thread, so deploys can be driven from async applications (web frameworks, GUIs, etc.) without an event loop conflict.

```python
async with AsyncPyinfra(state, host=host) as async_pyinfra:
    await async_pyinfra.connect()
    op_meta = await async_pyinfra.run_operation(server.shell, commands=["echo hi"])
    os_name = await async_pyinfra.get_fact(Os)
    await async_pyinfra.disconnect()
```

- All calls are serialised onto a single worker thread (`State` and the gevent hub are thread-bound), so concurrent `asyncio.gather` is safe.
- The process is not gevent monkey-patched (patching would break the host asyncio loop).
- `host` can be given per instance or per call; without a host, `connect()` / `disconnect()` use `connect_all` / `disconnect_all`.
- `run_operation` executes immediately and returns a completed `OperationMeta`.

## Supporting change: `connectors/util.py`

gevent's subprocess child watchers only exist on the default event loop (main thread). `run_local_process` and `read_output_buffers` now detect a non-default loop (worker thread) and fall back to stdlib `subprocess.Popen` and plain reader threads, bypassing any monkey-patching via `gevent.monkey.get_original`. Behaviour on the main thread (CLI, sync embedding) is unchanged. The discriminator is `gevent.get_hub().loop.default`, not thread identity, because gevent's threading patch makes main-thread greenlets look like non-main threads.

## Tests

The pytest process itself is gevent monkey-patched (`pyinfra_testing` imports `pyinfra_cli`), which is exactly the environment asyncio conflicts with, so the new tests run `AsyncPyinfra` scenarios in a clean subprocess via `sys.executable`.

- `uv run pytest tests/test_async_api.py -q`
- `uv run pytest --disable-warnings -m 'not end_to_end'` (2000 passed)
- `uv run ruff check` / `uv run ruff format --check` / `uv run mypy` on changed files

Refs #1474

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -105,3 +105,46 @@ run_ops(state)
 ```
 
 The available hooks are defined on `BaseStateCallback` — subclass it and override only the ones you care about. The current set covers host connect/disconnect, operation start/end, and per-host operation start/success/error/retry. See [`pyinfra.api.state.BaseStateCallback`](reference.md) for the full signature list.
+
+
+## Embedding in asyncio applications
+
+pyinfra's core is synchronous (gevent-based), which conflicts with an asyncio event loop running in the same thread. `pyinfra.async_api.AsyncPyinfra` runs all pyinfra work on a dedicated worker thread, so deploys can be driven from async applications (web frameworks, GUIs, and similar):
+
+```python
+import asyncio
+
+from pyinfra.api import Config, Inventory, State
+from pyinfra.async_api import AsyncPyinfra
+from pyinfra.facts.server import Os
+from pyinfra.operations import server
+
+inventory = Inventory((["@local"], {}))
+state = State(inventory=inventory, config=Config())
+host = inventory.get_host("@local")
+
+
+async def main():
+    async with AsyncPyinfra(state, host=host) as async_pyinfra:
+        await async_pyinfra.connect()
+
+        op_meta = await async_pyinfra.run_operation(
+            server.shell,
+            commands=["echo hello"],
+        )
+        print(op_meta.did_change())
+
+        print(await async_pyinfra.get_fact(Os))
+
+        await async_pyinfra.disconnect()
+
+
+asyncio.run(main())
+```
+
+A few things worth knowing:
+
+- **All calls are serialised onto one worker thread.** `State` and the gevent hub are bound to the thread that uses them, so `AsyncPyinfra` funnels every call through a single-thread executor. Concurrent `asyncio.gather` calls are safe, they just execute one after another.
+- **The process is not gevent monkey-patched.** Patching would break the host asyncio loop. pyinfra's internal cross-host concurrency (`run_ops` and friends) is unaffected when used inside the worker.
+- **The `host` is optional per call.** Pass a default host to `AsyncPyinfra(state, host=host)` or per call with `host=...` on `run_operation` / `get_fact`. Without a host, `connect()` / `disconnect()` apply to the whole inventory via `connect_all` / `disconnect_all`.
+- **`run_operation` executes immediately.** The operation runs on the target host as soon as the call is awaited, and the returned `OperationMeta` is complete (`did_change()`, `did_succeed()`, output, and so on are available right away).
--- a/src/pyinfra/async_api.py
+++ b/src/pyinfra/async_api.py
@@ -1,0 +1,150 @@
+"""
+Asyncio embedding API.
+
+pyinfra's core is synchronous (gevent-based), which conflicts with an asyncio
+event loop running in the same thread. This module runs all pyinfra work on a
+dedicated worker thread so deploys can be driven from asyncio applications.
+
+Note this does not gevent monkey-patch the process - doing so would break the
+host asyncio loop - so pyinfra's own internal concurrency still applies within
+the worker thread only.
+"""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Callable
+from concurrent.futures import ThreadPoolExecutor
+from contextlib import ExitStack
+from functools import partial
+from typing import Any
+
+from typing_extensions import Self
+
+from pyinfra.api.host import Host
+from pyinfra.api.operation import OperationMeta, execute_immediately
+from pyinfra.api.state import State
+from pyinfra.context import ctx_config, ctx_host, ctx_inventory, ctx_state
+
+
+class AsyncPyinfra:
+    """
+    Run pyinfra operations and facts from asyncio code.
+
+    All pyinfra calls are serialised onto a single worker thread because
+    ``State`` and the gevent hub are bound to the thread that uses them.
+
+    + state: the ``pyinfra.api.State`` to run against
+    + host: default ``Host`` to target, may be overridden per call
+
+    **Example:**
+
+    .. code:: python
+
+        async with AsyncPyinfra(state, host=host) as async_pyinfra:
+            await async_pyinfra.connect()
+            meta = await async_pyinfra.run_operation(server.shell, commands=["echo hi"])
+            uptime = await async_pyinfra.get_fact(Uptime)
+    """
+
+    def __init__(self, state: State, host: Host | None = None) -> None:
+        self.state = state
+        self.host = host
+        self._executor: ThreadPoolExecutor | None = None
+
+    async def __aenter__(self) -> Self:
+        return self
+
+    async def __aexit__(self, *exc_info: Any) -> None:
+        await self.close()
+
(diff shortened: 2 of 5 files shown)


src/pyinfra/async_api.py
```diff
@@ -0,0 +1,150 @@
...
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(self._get_executor(), partial(func, *args, **kwargs))
+
+    async def connect(self) -> None:
+        def _connect() -> None:
+            with ExitStack() as stack:
+                if self.host is not None:
+                    self._enter_context(stack, self.host)
+                    self.host.connect()
+                    self.state.activate_host(self.host)
```
StasShk, line 94: Maybe I'm missing something here, but could this activate a host even if the connection failed?

Host.connect() seems to catch ConnectError and leave host.connected false, while this calls activate_host() unconditionally. connect_all() checks host.connected before activating it.

wowi42: You're right, good catch. The single-host path now calls `host.connect(raise_exceptions=True)` so a failed connect raises `ConnectError` to the caller instead of activating the host. Added a test covering this.

src/pyinfra/connectors/util.py
```diff
@@ -162,6 +187,40 @@ def read_output_buffers(
...
+        for reader in readers:
+            reader.start()
+
+        deadline = None if timeout is None else time.monotonic() + timeout
+        for reader in readers:
+            remaining = None if deadline is None else max(0, deadline - time.monotonic())
+            reader.join(remaining)
+
+        if any(reader.is_alive() for reader in readers):
+            raise TimeoutError()
```
StasShk, line 261: I think there might be a cleanup issue here too. If a reader is still blocked on the pipe, this raise TimeoutError while the daemon thread is still alive. Then run_local_process() also skips process.wait() and closing the pipes.

The timeout test seems to hide this because the helper process exits right after. Could repeated timeouts leave reader threads/processes around in a long-running asyncio app?

wowi42: Good catch, fixed in 4513a917. `run_local_process` now catches the `TimeoutError`, kills the child, reaps it with `wait()` and closes both pipes before re-raisin, killing the child closes its pipe write ends, so the blocked daemon reader threads see EOF and exit. This also covers the pre-existing gevent path, which leaked the child process the same way.

Added a regression test (`TestRunLocalProcessTimeout`) that times out a genuinely long-lived command (`sleep 60`, timeout 1) on both path, worker thread (threaded readers) and main thread (gevent, and asserts no child processes remain and no reader threads linger. It fails without the fix with the leaked child you predicted.

One known limit: `process.kill()` kills the direct child only, so for compound shell commands a grandchild could still hold the pipe open. Handling that would need process-group kills (`start_new_session` + `killpg`), which felt like a bigger behavioral change than this fixx, happy to do it as a follow-up if you think it is worth it.

wowi42: Follow-up: the grandchild caveat turned out to be the default case, not an edge case. CI failed on this very test, on Debian/Ubuntu `/bin/sh` (dash) forks even a simple `sleep 60` instead of execing it, so `kill()` reached only the shell, the orphaned child kept the pipes open, and the subsequent `close()` deadlocked behind the blocked reader thread. (macOS `/bin/sh` does exec, which is why it passed locally.)

Fixed properly in 3cc8752d: timed commands now spawn in their own session (`start_new_session=True`, so PGID == child PID) and the timeout path calls `killpg()` on the whole group. Scoped to timed commands only, since a new session detaches the child from the controlling terminal and Ctrl-C delivery, which untimed paths (`local.shell`, rsync transfers) rely on. Verified in an ubuntu/python:3.11 container: both the threaded and gevent paths clean up fully, including a compound `sleep 60; echo done` command where the grandchild outlives the shell.

docs/api/index.md
```diff
@@ -105,3 +105,46 @@ run_ops(state)
...
+        await async_pyinfra.disconnect()
+
+
+asyncio.run(main())
+```
+
+A few things worth knowing:
+
+- **All calls are serialised onto one worker thread.** `State` and the gevent hub are bound to the thread that uses them, so `AsyncPyinfra` funnels every call through a single-thread executor. Concurrent `asyncio.gather` calls are safe, they just execute one after another.
+- **The process is not gevent monkey-patched.** Patching would break the host asyncio loop. pyinfra's internal cross-host concurrency (`run_ops` and friends) is unaffected when used inside the worker.
```
StasShk, line 148: I tested the cross-host concurrency locally, and I think there is a real concurrency issue in this path.

Repro: https://gist.github.com/StasShk/eb4bdb1694c71e2c00dae8e148b1a82c
With two hosts and PARALLEL=2, each calling run_local_process("sleep 5"), the second host doesn't start until the first one finishes:

max_active=1, total runtime ~10s instead of ~5s.

It looks like the stdlib/threaded run_local_process path blocks the single worker thread while waiting, so the other gevent greenlet can't run. In practice, the host operations end up running sequentially even though PARALLEL=2.

Is this expected, or does it mean cross-host concurrency isn't preserved in AsyncPyinfra mode?

wowi42: Not expected at all, it was a real bug, thanks for the repro. Fixed in b27a263e.

The threaded reader path waited on its reader threads with a blocking `Thread.join()`, and `run_local_process` then blocked in `Popen.wait()`. Both park the single worker thread, which freezes the gevent hub for the whole duration of the command - so the greenlet for host 2 never got scheduled until host 1 finished. Exactly what your repro showed.

The fix keeps the blocking pipe reads on the real OS reader threads (that part is required, since stdlib file objects can't be read cooperatively off the default loop) but waits for them cooperatively: poll `is_alive()` / `process.poll()` with `gevent.sleep(0.05)` so the hub keeps scheduling other hosts' greenlets.

Verified with your exact repro in an ubuntu/python:3.11 container: `max_active: 2`, elapsed ~5.03s, both hosts start immediately. Also added `test_cross_host_concurrency` to `tests/test_async_api.py` (two hosts, PARALLEL=2, `run_local_process("sleep 2")` per connect, asserts `max_active == 2`) - it fails without the fix.

src/pyinfra/connectors/util.py
```diff
@@ -63,16 +104,40 @@ def run_local_process(
...
 
     logger.debug("--> Waiting for exit status...")
-    process.wait()
+    if on_default_loop:
+        process.wait()
+    else:
+        # Poll cooperatively rather than blocking in wait(): this runs on the
+        # single worker thread of an embedded deploy, where a blocking wait
+        # would freeze the gevent hub and serialise greenlets for other hosts.
+        while process.poll() is None:
```
StasShk: I found one more timeout edge case while testing this.

`run_local_process("exec >/dev/null 2>&1; sleep 5", timeout=1)` returns normally after ~5 seconds instead of timing out after ~1 second:

returned after: 5.024s

It looks like once stdout/stderr are closed, read_output_buffers() returns and the timeout is no longer checked while waiting for the process itself to exit.

Should the timeout cover the whole command lifetime here?

wowi42: Yes, it should - `_timeout` is documented as "timeout for *each* command executed", so it should bound the whole command lifetime. Fixed in a3b59b99.

The read phase returned as soon as the pipes hit EOF (your `exec >/dev/null 2>&1` case) and the final `process.wait()` had no deadline. This affected the gevent path too, pre-existing.

The timeout deadline is now computed once and enforced in the final wait as well: `gevent.Timeout` around `process.wait()` on the default loop, and the deadline checked inside the cooperative poll loop off it. Both share the existing cleanup (kill process group, reap, close pipes, re-raise `TimeoutError`).

Verified in an ubuntu/python:3.11 container with your exact command on both paths: `timed out after 1.02s` (was ~5s), no leaked children. Added `test_timeout_covers_process_lifetime_after_pipes_close` - it runs the full 30s sleep without the fix.

Fizzadar (commented):
This works but I think we can just use #1957 which we should be able to land pretty quickly, and avoid a compatibility layer.

wowi42:
> This works but I think we can just use #1957 which we should be able to land pretty quickly, and avoid a compatibility layer.

correct, closing it

PR #1954, opened 2026-09-08
feat(facts): support requires_root attribute

SecT0uch:
Adds a `requires_root` attribute to `FactBase` to allow facts to opt out of inheriting privilege escalation arguments from the enclosing operation, as discussed in #1939 

`requires_root` defaults to `True`, preserving the existing behaviour.

When set to `False`, privilege escalation arguments such as `_sudo`, `_su_user`, `_doas`, and `_dzdo` are not inherited by the fact. Explicit privilege escalation arguments passed directly to the fact still take precedence.

### DIFF
--- a/src/pyinfra/api/facts.py
+++ b/src/pyinfra/api/facts.py
@@ -43,6 +43,22 @@ if TYPE_CHECKING:
 
 # Sentinel output line emitted when skip_unless_command binary is absent on the remote host.
 _MISSING_COMMAND_MARKER = "##PYINFRA_NOCMD##"
+_FACT_PRIVILEGE_ARGUMENTS = (
+    "_sudo",
+    "_sudo_user",
+    "_use_sudo_login",
+    "_sudo_password",
+    "_preserve_sudo_env",
+    "_su_user",
+    "_use_su_login",
+    "_preserve_su_env",
+    "_su_shell",
+    "_su_password",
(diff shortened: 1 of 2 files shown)


src/pyinfra/api/facts.py
Fizzadar: I think the sudo removal needs to be down here? otherwise any host, global or data level sudo argument will still be applied.

Fizzadar (changes requested):
Idea is good just think it needs a little tweak to prevent more sudo bools leaking in!

src/pyinfra/api/facts.py
SecT0uch: How about the new version ?
Diff: https://github.com/pyinfra-dev/pyinfra/compare/8e6926df762905a27e7eaf4c9f967099759dd14b..6fb62ef807c17b863393a86c7e926e402f0261ce

Fizzadar (changes requested):
Hi @SecT0uch - I’ve had a chance to think about this one a bit more and I don’t think we can land it as-is. My concern is introducing a hidden behavior that breaks all the data inheritance expectations we set [in the docs](https://docs.pyinfra.com/en/3.x/inventory-data.html#data-hierarchy).

Aside from the unnecessary sudo call is there any reason suppressing privilege escalation is useful? Might be other options..

